PR #760, opened 2014-12-16
Restore headlines plugin

juxtin:
This reverts commit 29d80ce2361651b2533da6b69b49b6e000f8f9e0.

Headlines may not be used internally, but they're still a great feature that some people (including myself and some of my co-workers) find indispensable. I'd love to see the functionality restored if that's all right. Thanks!

### DIFF
--- a/helm-plugin.el
+++ b/helm-plugin.el
@@ -142,6 +142,93 @@
   Will list all lines in .emacs.el.")
 
 
+;;; Plug-in: `headline'
+;;
+;;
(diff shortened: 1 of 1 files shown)


holguinj:
@krisajenkins as @thierryvolpiatto pointed out (and graciously demonstrated), it's still possible to write headline functions. See the above, or my slightly updated version [here](https://github.com/holguinj/emacs.d/blob/master/config/clj.el#L19)

Issue #466, opened 2014-03-29
overriding-local-map override other minor-mode keymap


myuhe:
See https://github.com/emacs-helm/helm/blob/master/helm.el#L2193

Other minor-mode keymap is broken by overriding-local-map.

In detail,

``` lisp
(Info-goto-node "(elisp)Controlling Active Maps")
```

PR #785, opened 2014-12-30
Fix inheritance of input method for prompt

Fuco1:
Minibuffer inherits the input method from the current-buffer.  At the time `read-from-minibuffer` is called, the active buffer is the helm buffer displaynig candidates---this has always the default active method, thus the inheritance doesn't work.

We need to inherit the method from the `helm-current-buffer`, which is where the helm was called from.  We can set the method manually in the `minibuffer-setup-hook`

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -2298,6 +2298,7 @@ For ANY-PRESELECT ANY-RESUME ANY-KEYMAP ANY-DEFAULT ANY-HISTORY, See `helm'."
                (unwind-protect
                     (minibuffer-with-setup-hook
                         #'(lambda ()
+                            (set-input-method (with-current-buffer helm-current-buffer current-input-method))
                             (setq timer (run-with-idle-timer
                                          (max helm-input-idle-delay 0.001) 'repeat
                                          #'(lambda ()


tuhdo:
@thierryvolpiatto If I understand correctly, here is what he wants:

- Suppose you have French documents, with French text inside.
- You set your input method to `french`.
- You run `helm-occur` or `helm-do-grep` to search text and you expect your input method to remain in the minibuffer, so you can enter French text go grep/occur.
- Currently, you have to set input method again.

@Fuco1 Your patch introduces a bug, as @thierryvolpiatto said. Here is a demo:

fuco-patch

Does Japanese use Helm? [Yes](https://github.com/emacs-helm/helm/issues/466). If your patch is merged, someone will open an issue soon. So, until you fix such bug, the patch can be merged.

tuhdo:
Earlier I said that I did not encounter this problem in Windows. Well, I forgot the apply the change in this PR. My bad.

thierryvolpiatto:
Tu Do <notifications@github.com> writes:

> @thierryvolpiatto If I understand correctly, here is what he wants:
>
>   * Suppose you have French documents, with French text inside.
>   * You set your input method to french.
>   * You run helm-occur or helm-do-grep to search text and you expect your input method to remain in the minibuffer, so you can enter French text go grep/occur.
>   * Currently, you have to set input method again.
>
> @Fuco1 Your patch introduces a bug, as @thierryvolpiatto said. Here is a demo:
>
> fuco-pat
>
> Does Japanese use Helm? Yes. If your patch is merged, someone will open an issue soon. So, until you fix such bug, the patch can be merged.

For memory for self if I have to come back on this, the fix should happen in
helm-create-helm-buffer and just setq current-input-method.
We'll still have a problem with multibyte characters, don't know yet
what to to with this.

-- 
Thierry
Get my Gnupg key:

PR #801, opened 2015-01-05
Make helm-autoresize-mode use percentage

tuhdo:
Percentage is more flexible and consistent across screen size than fixed
number of lines. i.e. 30% is always 30%, while 30 lines can be too big
in a monitor but can be small in another monitor.


thierryvolpiatto:
Thanks.

PR #813, opened 2015-01-10
Make helm mode-line ending match regular emacs mode line

PythonNut:
Sorry for the trivial patch.

This makes the helm mode-line behave more like the regular emacs one. By default, the dashes at the end are shown only in a TTY. If the user customizes `mode-line-end-spaces`, the changes are reflected in both mode lines.

This came about as part of my war on visual clutter. I have `mode-line-end-spaces` set to `nil` at all times.


thierryvolpiatto:
Thanks, I wanted to change this old emacs style since long time and always forget.

PR #814, opened 2015-01-10
Make helm-score-candidate-for-pattern a defun, not a defsubst

PythonNut:
Advantages:
* `helm-score-candidate-for-pattern` can be customized by the user for great good.
* Smaller byte-compiled result, as `helm-score-candidate-for-pattern` is a decently large function.

Disadvantages:
* `helm-fuzzy-matching-default-sort-fn` pays at best twice `O(n log n)` function calls. However, this is negligible compared to the number of calls already done in the sort function. I have not noticed a performance difference, even on large sets.

Advantage 1 is huge, as it allows the user to tweak and modify helm's scoring function, possibly to substantially improve the intelligence of helm's fuzzy matching.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -2816,7 +2816,7 @@ This function is used with sources build with `helm-source-sync'."
            when (cdr str)
            collect (list (car str) (cadr str))))
 
-(defsubst helm-score-candidate-for-pattern (candidate pattern)
+(defun helm-score-candidate-for-pattern (candidate pattern)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Agree, about making the score candidate smarter, I didn't make it smarter than this to have a good balance between performances and being enough smart.

thierryvolpiatto:
Thanks, merged.

PR #816, opened 2015-01-11
Explicitly enable semantic-mode

tuhdo:
Currently, when entering a file that is supported by Semantic,
helm-semantic returns an empty candidate buffer because semantic-mode is
disabled in current buffer even if previously semantic-mode was
activated. This happens in emacs-lisp-mode. semantic-mode works properly
for C/C++ only.

To avoid empty buffer when using helm-semantic, explicitly enable
semantic-mode for fetching its tags.

### DIFF
--- a/helm-semantic.el
+++ b/helm-semantic.el
@@ -109,6 +109,7 @@
 
 (defclass helm-semantic-source (helm-source-in-buffer)
   ((init :initform (lambda ()
+                     (semantic-mode 1)
                      (helm-semantic--maybe-set-needs-update)
                      (setq helm-semantic--tags-cache (semantic-fetch-tags))
                      (with-current-buffer (helm-candidate-buffer 'global)


thierryvolpiatto:
Tu Do <notifications@github.com> writes:

> Currently, when entering a file that is supported by Semantic,
> helm-semantic returns an empty candidate buffer because semantic-mode is
> disabled in current buffer even if previously semantic-mode was
> activated. This happens in emacs-lisp-mode. semantic-mode works properly
> for C/C++ only.
>
> To avoid empty buffer when using helm-semantic, explicitly enable
> semantic-mode for fetching its tags.

No, I don't want to enable semantic-mode in this command, it will not
work as expected in many previous emacs versions, it will not work
properly for elisp files without extra config etc... It is why we have
helm-semantic-or-imenu and it is the default command (C-x c i).

-- 
Thierry
Get my Gnupg key:

tuhdo:
Semantic was bundled with Emacs since 23.3. By simply enable semantic-mode, it's supposed to work without extra config for Emacs Lisp. Howerver, semantic-mode bundled with 24.4, for some reason even when semantic-mode is enabled, open an Elisp buffer it is still disabled. That's why we ecplicitly enable it in this command just to fetch tags. Yes, just enable semantic-mode is enough, no extra config is needed. Maybe you are afraid that it affects Emacs locally, then we disable it again after fetching tags. But I think, for users who use this command , they would be Semantic users too.

tuhdo:
Okay I think that it's better to not cause global side effect with Semantic everytime helm-semantic is used. The latest CEDET fixed this issue.

PR #820, opened 2015-01-13
new macro `helm-with-gensyms'; fix local macro variables

michael-heerdegen:
Please review. I treated only those macros that can "paste" in arbitrary code (à la &rest body), the other macro definitions are not susceptible for variable capturing AFAIK.

And I excluded helm-aif of course, because there we want a user visible local variable.


thierryvolpiatto:

> Please review. I treated only those macros that can "paste" in arbitrary code (à la &rest body), the other macro definitions are not susceptible for variable
> capturing AFAIK.
>
> And I excluded helm-aif of course, because there we want a user visible local variable.

Thanks, merged.

-- 
Thierry
Get my Gnupg key:

PR #825, opened 2015-01-17
Add a "Find File recursive" action to helm-find-files

PythonNut:
This adds a action that triggers `helm-find` from `helm-find-files`. I use this to search large unstructured directory trees. I use it habitually when I know there's a deep path ahead and I know `find` can traverse down faster than I can.

I can't guarantee that it's perfect, but it _does_ work as far as I can tell.


thierryvolpiatto:
PythonNut <notifications@github.com> writes:

> This adds a action that triggers helm-find from helm-find-files.

This already exists.  See also C-x C-d and C-x C-f both with and without
prefixes.


-- 
Thierry
Get my Gnupg key:

PythonNut:
Ah. I see. Why is it under a different name?

PR #833, opened 2015-01-20
Pr/get display minimize cons

lewang:
This is a slight optimization, plus refactoring the insert code to reuse the canonical display part extraction function.


thierryvolpiatto:
Looks correct, merged, thanks.

PR #839, opened 2015-01-24
helm-find-shell-command-fn: fix using case fold

michael-heerdegen:
Limit determination of whether to use case fold to globbing patterns


thierryvolpiatto:
Thanks Michael.

PR #843, opened 2015-01-27
helm-imenu-candidates: memorize buffer tick at end (fix #841)

### LINKED ISSUE #841: Keyboard input while helm-imenu is being autoloaded breaks the result
1. Install Helm as a package, don't do any additonal setup.

2. Open some Emacs Lisp file with several function definitions.

3. Type `M-x helm-imenu`, don't wait for the interface to show up, type the first letter of the input.

4. See the empty list.

Repeated `M-x helm-imenu` invocations show the empty list as well. Only if after I make a buffer modification, `M-x helm-imenu` will show the right results.

michael-heerdegen:


thierryvolpiatto:
Great, nothing to do :-)
Thanks.

PR #847, opened 2015-01-28
rewrite with-helm-restore-variables to use let

michael-heerdegen:
I noticed that with-helm-restore-variables reinvents the wheel.  I think it is saner and safer to just use let.
This proposed change should not change anything in behavior.

If you think helm-restored-variables should not become a defconst, we could still use cl-progv instead of let.


thierryvolpiatto:
Agreed, merged, thanks.

PR #856, opened 2015-02-02
fix #854

### LINKED ISSUE #854: helm-apropos is slow
`helm-apropos`  is slow, it takes more than 2 seconds to show commands-functions-variables-and-faces list window, but `C-h v` and `C-h f` are instantly. 

Is is possible to make it faster?

michael-heerdegen:
optimize helm-apropos-init by accumulating candidates in a string

### DIFF
--- a/helm-elisp.el
+++ b/helm-elisp.el
@@ -416,6 +416,23 @@ Filename completion happen if string start after or between a double quote."
         (all-completions "" obarray test)
         "\n")))))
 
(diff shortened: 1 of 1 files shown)


michael-heerdegen:
After the first optimization, facep tests were responsible for 80% of the remaining delay. So I propose to use face-list to initialize the face source.

thierryvolpiatto:
Thanks.

PR #857, opened 2015-02-05
First step of fuzzy re-order

lewang:
Hi Thierry,

There is a lot to talk about here, so please bear with me:

- Call the highlighter from within the helm-fuzzy-matching-default-sort-fn
- Remove customizability of highlighter.
- Only highlight candidates that will be printed.
- Remove "use-real" hack of sort function.
    - Instead rely on text-properties to figure out portion of display that should be matched.

# BREAKING CHANGE

I propose normalizing the collection of candidates to an alist as a first step to processing filtered-candidate-transformer.

This makes the shape of candidates structure more predictable.  I want to do this as a step towards the end goal of making fuzzy a more integral part of the helm experience.  I believe doing so will solve a lot of problems with the sorting.

This means that any `filtered-candidate-transformer` used needs to be modified to process a-lists.  I've made the corresponding changes to the sources in the main repository (work in progress, as I don't use most of them).

----------

Please let me know what you think of the idea.  I look forward to discussing any concerns you have.

### DIFF
--- a/helm-bbdb.el
+++ b/helm-bbdb.el
@@ -133,7 +133,7 @@ All other actions are removed."
     (filtered-candidate-transformer . (lambda (candidates _source)
                                         (setq helm-bbdb-name helm-pattern)
                                         (if (not candidates)
-                                            (list "*Add to contacts*")
+                                            (list (helm-normalize-candidate "*Add to contacts*"))
                                           candidates)))
     (action-transformer . (lambda (actions candidate)
                             (helm-bbdb-create-contact actions candidate))))
--- a/helm-bookmark.el
+++ b/helm-bookmark.el
@@ -100,18 +100,19 @@
   "See (info \"(emacs)Bookmarks\").")
 
 (defun helm-bookmark-transformer (candidates _source)
-  (cl-loop for i in candidates
-        for loc = (bookmark-location i)
-        for len =  (string-width i)
+  (cl-loop for cand in candidates
+           for real = (cdr cand)
+        for loc = (bookmark-location real)
+        for len =  (string-width real)
         for trunc = (if (> len bookmark-bmenu-file-column)
-                        (helm-substring i bookmark-bmenu-file-column)
-                      i)
+                        (helm-substring real bookmark-bmenu-file-column)
+                      real)
         for sep = (make-string (- (+ bookmark-bmenu-file-column 2)
                                   (length trunc))
                                ? )
         if helm-bookmark-show-location
-        collect (cons (concat trunc sep (if (listp loc) (car loc) loc)) i)
-        else collect i))
+        collect (helm-normalize-candidate (cons (concat trunc sep (if (listp loc) (car loc) loc)) real))
+        else collect cand))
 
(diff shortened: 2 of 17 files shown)


thierryvolpiatto:
Le Wang <notifications@github.com> writes:

> Hi Thierry,
>
> There is a lot to talk about here, so please bear with me:
>
>   * Call the highlighter from within the helm-fuzzy-matching-default-sort-fn
>   * Remove customizability of highlighter.
>   * Only highlight candidates that will be printed.
>   * Remove "use-real" hack of sort function.
>       + Instead rely on text-properties to figure out portion of display that should be matched.
>
> BREAKING CHANGE
>
> I propose normalizing the collection of candidates to an alist as a first step to processing filtered-candidate-transformer.
>
> This makes the shape of candidates structure more predictable. I want to do this as a step towards the end goal of making fuzzy a more integral part of the helm
> experience. I believe doing so will solve a lot of problems with the sorting.
>
> This means that any filtered-candidate-transformer used needs to be modified to process a-lists. I've made the corresponding changes to the sources in the main
> repository (work in progress, as I don't use most of them).

Please don't change so many sources, just make changes on one or two
sources, say one listing symbols and one listing files and see what
happen (e.g lisp-completion and recentf).
For the moment your branch is not compiling at all, so I can't check.

-- 
Thierry
Get my Gnupg key:

lewang:
On Wed, Feb 4, 2015 at 11:54 PM, Thierry Volpiatto <notifications@github.com
> wrote:

> Please don't change so many sources, just make changes on one or two
> sources, say one listing symbols and one listing files and see what
> happen (e.g lisp-completion and recentf).
> For the moment your branch is not compiling at all, so I can't check.


I'm sorry if you got the impression that this is ready.  I just wanted on
your feedback on the idea of "normalizing" candidates to an alist.  I will
continue to work on this branch to fix the problems.

-- 
> Thierry
> Get my Gnupg key:
> gpg --keyserver pgp.mit.edu --recv-keys 59F29997
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/emacs-helm/helm/pull/857#issuecomment-72994238>.
>



-- 
Le

lewang:
@thierryvolpiatto This branch compiles fine now.  I'm sure there is more work to fix specific filtered-candidate-transformers with some sources, because I don't use them all.

How have you handled breaking API changes before?

thierryvolpiatto:
Le Wang <notifications@github.com> writes:

> @thierryvolpiatto This branch compiles fine now. I'm sure there is
> more work to fix specific filtered-candidate-transformers with some
> sources, because I don't use them all.
>
> How have you handled breaking API changes before?

Your changes have to be backward compatible, there is a lot of helm
packages that are working fine actually that will break.
IOW your changes should not affect existing sources.

-- 
Thierry
Get my Gnupg key:

lewang:
> Your changes have to be backward compatible

I'm closing this as if this is a set-in-stone rule, then there is no point.  API have to be possible.  That is the whole point of semantic versioning.

PR #877, opened 2015-02-18
allow debugging candidate functions

michael-heerdegen:
Currently, calling candidate functions is wrapped inside condition-case.  This makes it very difficult to debug.

If we would use c-c-unless-debug instead, like we already do in other places, this would make our life with helm much nicer.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -2602,7 +2602,7 @@ Helm plug-ins are realized by this function."
                        (error
                         "`%s' must either be a function, a variable or a list"
                         (or candidate-fn candidate-proc))))
-         (candidates (condition-case err
+         (candidates (condition-case-no-debug err
                          ;; Process candidates-(process) function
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
> Very good.

Will merge as soon you change this.

michael-heerdegen:

>     Very good.
>
> Will merge as soon you change this.

Change what?  Adding the noticeable error message?

thierryvolpiatto:

>
>> Very good.
>>
>> Will merge as soon you change this.
>
> Change what?

condition-case-no-debug => condition-case-unless-debug

-- 
Thierry
Get my Gnupg key:

michael-heerdegen:
Oops, I'm sorry.

thierryvolpiatto:
Thanks, merged.

PR #886, opened 2015-02-26
save-excursion when retrieving heading name for linking in

### LINKED ISSUE #884: helm-org inserts link to wrong location
When calling `helm-org-in-buffer-headings`, if I try to insert a heading, it ends up inserted immediately after the heading, and not at point. This also happens with `helm-org-agenda-files-headings`, if the header selected is the same as the current buffer. Example:

`* Book [[*Book]]`

The problem is in `helm-org-insert-link-to-heading-at-marker`. It assumes the header buffer is different from the current buffer, and `goto-char` ends up moving the point to the header, but `with-helm-current-buffer` doesn't return the point back to the current location. 

I fixed it like this, but there may be a better solution:

```elisp
(defun helm-org-insert-link-to-heading-at-marker (marker)
  (let ((old-pnt (point-marker)))
    (with-current-buffer (marker-buffer marker)
      (goto-char (marker-position marker))
      (let ((heading-name (nth 4 (org-heading-components)))
            (file-name buffer-file-name))
        (message heading-name)
        (message file-name)
        (with-helm-current-buffer
          (goto-char old-pnt)
          (org-insert-link
           file-name (concat "file:" file-name "::*" heading-name)))))))
```

IvanMalison:
helm-org-insert-link-to-heading-at-marker. Fixes #884

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -62,11 +62,9 @@ NOTE: This will be slow on large org buffers."
 
 (defun helm-org-insert-link-to-heading-at-marker (marker)
   (with-current-buffer (marker-buffer marker)
-    (goto-char (marker-position marker))
-    (let ((heading-name (nth 4 (org-heading-components)))
-          (file-name buffer-file-name))
-      (message heading-name)
-      (message file-name)
+    (let ((heading-name (save-excursion (goto-char (marker-position marker))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Great, thanks.

Issue #889, opened 2015-02-27
Show hex code in helm-ucs.


tmalsburg:
helm-ucs shows the name of a character and the character itself.  In many situations it is also useful to know the hex code of the character.  Please consider adding that.

thierryvolpiatto:
I have no idea how to do this, maybe you can enlight me, or even better provide a PR.

tmalsburg:
Hi Thierry, please have a look at #892.

michael-heerdegen:
Are we speaking here about the hex code of a character (in the sense that every character in Gnu Emacs is a number), or about the Unicode hex code?

Seems these are not always identical, since there are different read syntaxes for these two things.

tmalsburg:
In this context, we'd like to see the unicode codepoint.  Are you suggesting that the results may depend on the locale under which Emacs is running? Looking at the source of `ucs-names`, which supplies the the data for `helm-ucs`, it seems to me that the results should be invariant to the locale.  I may be wrong, though.

michael-heerdegen:
Titus von der Malsburg <notifications@github.com> writes:

> In this context, we'd like to see the unicode codepoint. Are you
> suggesting that the results may depend on the locale under which Emacs
> is running? Looking at the source of ucs-names, which supplies the the
> data for helm-ucs, it seems to me that the results should be invariant
> to the locale. I may be wrong, though.

I vaguely remember to have heard they can differ a long time ago.
Sadly, I do not know anything about this stuff.  We should be sure to
DTRT.

I don't think it is locate dependent however.  This is something
"outside of Emacs".

tmalsburg:
I think it should be ok. `ucs-names` gives us the code points and all my code is doing is to convert them to hex.

michael-heerdegen:
I see, mule does the critical work.

I could not find a case here were unicode hexcode and character number in hex would differ btw:

```
(loop for i from 1 to 1000000
      when (not (= (eval (read (format "?\\U00%06x" i)))
                   (eval (read (format "?\\x%06x" i)))))
      collect i)
==> nil
```

tmalsburg:
Thanks for checking that, Michael!

tmalsburg:
This issue was addressed in db55080767030cfdf78f1ae06171274b7f67c3f7 and 92b4506b16695d1e33dcb67c7070117692875afb.

PR #892, opened 2015-02-27
Show codepoint in hex with description and char.

tmalsburg:
Addresses #889. I'm not sure if `helm-ucs-max-len` is the max length of the description or the max length of the line.  If it's the length of the description, the PR should be ready to be merged.  If not, I have to amend it.  Adding a docstring to `helm-ucs-max-len` is a good idea either way.

### DIFF
--- a/helm-font.el
+++ b/helm-font.el
@@ -111,12 +111,19 @@ Only math* symbols are collected."
   (with-helm-current-buffer
     (delete-char -1)))
 
-(defun helm-ucs-insert-char (candidate)
+(defun helm-ucs-insert (candidate n)
   (with-helm-current-buffer
-    (insert
-     (replace-regexp-in-string
-      " " ""
-      (cadr (split-string candidate ":"))))))
+    (string-match "^\\([^:]+\\): +\\(.\\) #x\\([a-f0-9]+\\)$" candidate)
+    (insert (match-string n candidate))))
+
+(defun helm-ucs-insert-name (candidate)
+  (helm-ucs-insert candidate 1))
+
+(defun helm-ucs-insert-char (candidate)
+  (helm-ucs-insert candidate 2))
+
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Hi Titus,

Titus von der Malsburg <notifications@github.com> writes:

> I'm not sure if helm-ucs-max-len is the max length of the description
> or the max length of the line. If it's the length of the description,
> the PR should be ready to be merged. If not, I have to amend
> it. Adding a docstring to helm-ucs-max-len is a good idea either way.
Yes it is the length of the car of each elements, it is used to align all.

Ok, looks good, but insertion is wrong, you have to modify the action:

(defun helm-ucs-insert-char (candidate)
  (with-helm-current-buffer
    (insert
     (replace-regexp-in-string
      " " ""
      (car (split-string
            (cadr (split-string candidate ":"))))))))

And probably you want an action for inserting the hex code:

(defun helm-ucs-insert-hex (candidate)
  (with-helm-current-buffer
    (insert
     (replace-regexp-in-string
      " " ""
      (cadr (split-string
            (cadr (split-string candidate ":"))))))))

And finally add it to the source (If you have time you can rewrite it
with helm-build-in-buffer-source).

Many thanks ;-)


-- 
Thierry
Get my Gnupg key:

tmalsburg:
@thierryvolpiatto, I rewrote the code for inserting characters.  The previous version failed when then inserted character was the space character.  I also added insert functions for the name and the code point of a character.

Why is `with-helm-current-buffer` needed here? Inserting seems to work just as well without it.

thierryvolpiatto:
Thanks.

> Why is with-helm-current-buffer needed here? Inserting seems to work just as well without it.
It may work most of the time but may fail in some corner cases, using `with-helm-current-buffer` ensure insertion in the right place.

PR #893, opened 2015-03-01
helm-browse-menubar: use multiple sources instead of prefix arg (fix #885)

### LINKED ISSUE #885: helm-browse-menubar: prefix arg vs. multiple sources
Hi,

would you accept a PR regarding the following:

You decided for "helm-browse-menubar" to control which parts from the menu-bar are candidates (global, minor-mode, major-mode menus) via prefix arg.  But I think in this case, it would be simpler and justifiable to use a combined session for all of them, since the decomposition of the menu-bar in parts is an implementation detail - there is only one menu bar.  This is what I use and would suggest more or less:

```
  (defun my-helm-browse-menubar ()
    "My helm interface to the menubar using lacarte.el."
    (interactive)
    (require 'lacarte)
    (helm :sources (mapcar 
                    (lambda (spec) (helm-make-source (car spec) 'helm-lacarte
                                :candidates (lambda () (helm-lacarte-get-candidates (cdr spec)))))
                    '(("Major Mode"  . (local))
                      ("Minor Modes" . (minor))
                      ("Global Map"  . (global))))
          :buffer "*helm lacarte*"))
```

Thanks.

michael-heerdegen:


thierryvolpiatto:
Thanks, Michael.

PR #919, opened 2015-03-15
Full outline path to a heading included in org helm candidates

amarek:
Hello,
Currently, the only information about org heading's context within the helm candidates list
(after invoking helm-org-in-buffer-headings/helm-org-agenda-files-headings) is the number of asterisks preceding the title that reflect the heading's level and its position within the list with respect to its parent and peers.  Unfortunately, after narrowing down the list via helm facilities, the latter information is lost (as the peers and parent are likely filtered out) and identifying the precise heading we are looking for gets tricky if there are multiple headings with similar names but different position within the org hierarchy.

To remedy that, this change introduces the full outline paths to org headings in the candidates list. The paths are formatted via Org's facilities that make sure that important information (such as the title itself) is not obscured if parts of the path need to be truncated.

In addition to removing any ambiguity in terms of the location of particular heading within Org's hierarchy, it allows to filter out the headings not only by headings' titles but also by their outline path (much in the vein of narrowing down the list of files based on their file-system paths).

Cheers,
Adam

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -101,9 +101,10 @@ NOTE: This will be slow on large org buffers."
         (cl-loop while (re-search-forward org-complex-heading-regexp nil t)
               if (let ((num-stars (length (match-string-no-properties 1))))
                    (and (>= num-stars min-depth) (<= num-stars max-depth)))
-              collect `(,(if nofname (funcall match-fn 0)
-                             (concat (helm-basename filename)
-                                 ": " (funcall match-fn 0)))
+              collect `(,(let ((heading (funcall match-fn 4))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Otherwise looks nice, thanks.

amarek:
Switched to unless.

thierryvolpiatto:
Thanks, merged!

amarek:
Yeah ... come to think of it, the "fast mode" looks like the way to go (at least based on the docstring) - we are always scanning full files after all. Will take a look at that. If it proves too complex, the additional defcustom might indded be necessary (thought of it initially, but it didn't pass the "Ockhahm's razor" test).

PR #931, opened 2015-03-19
[Fix #929] Improve helm-semantic interface

### LINKED ISSUE #929: How can i config the default input of helm-semantic-or-imenu
Because I just want to see functions in most case.
By the way , I think in C/C++ mode ,displaying function param in helm-semantic-or-imenu as one line is a little bit wasteful .
image

tuhdo:
- Add a custom variable helm-semantic-display-style that allows user to
  choose between two display style: the default (the current style) and
  a more concise style, semantic-format-tag-prototype that omits all the
  information about a tag (i.e. no "Function: " or "Variable: " is
  appended in each entry).

- Remove "Class" in format string to reduce the verbosity. Instead, only
  show the parent name next to current tag.

- Don't create a candidate for function parameters. User would be faster
  to jump to a function and navigate to the function parameter.

### DIFF
--- a/helm-semantic.el
+++ b/helm-semantic.el
@@ -50,6 +50,13 @@
 ;; Internals vars
 (defvar helm-semantic--tags-cache nil)
 
+(defcustom helm-semantic-display-style 'semantic-format-tag-summarize
+  "Function to present a semantic tag."
+  :group 'helm-semantic
+  :type '(radio
+          (const :tag "Default" semantic-format-tag-summarize)
+          (const :tag "Prototype" semantic-format-tag-prototype)))
+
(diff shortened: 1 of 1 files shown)


tuhdo:
Here is a screenshot of using `semantic-format-tag-prototype` in C++:

helm-semantic-prototype

As you see, function parameters are no longer appended under each function. This is the same for default format function `semantic-format-tag-summarize`.

tuhdo:
Here is a screenshot of the default: 

helm-semantic-default

tuhdo:
Here is a screenshot of our current `helm-semantic`, so you can compare:

helm-semantic-current

PR #934, opened 2015-03-20
org-get-outline-path used in fast mode to accelerate large org file scans

amarek:
Listing org headlines with full outline path included is noticibly faster now on my (moderately sized) org files.


thierryvolpiatto:
Looks good, thanks.

PR #950, opened 2015-03-27
Inhibit load message

thomasf:
Inhibit load message


thierryvolpiatto:
If you want :-)

Thanks, merged.

PR #957, opened 2015-03-30
Option to search in all buffers for dabbrev-expend

amarek:
helm-dabbrev restricts searching for expansions to buffers with identical major mode or buffers matching one of the mappings in helm-dabbrev-major-mode-assoc. This is different from the dabbrev-expand behavior which has an option (enabled by default) to remove any restrictions by mean of the dabbrev-check-all-buffers setting.

This change introduces additional semantics for helm-dabbrev-major-mode-assoc to allow non-restricted search if set to nil  (the default is not changed though).


thierryvolpiatto:
Thanks, but I think this function should do it's job, no more.
I have introduced a new variable that when nil allow searching all buffers inconditionally.

PR #965, opened 2015-04-02
helm-etags-select selected region if possible

redguardtoo:
In modern html5 projects, the tag/keyword to
look up could be complicated

For example, in AngularJS, a controller is a string
which could contain any characters.


thierryvolpiatto:
Thanks.

PR #972, opened 2015-04-11
Allow helm-semantic to open in other window

tuhdo:
It creates a cloned buffer and open a tag there. Useful when we want to
keep the current window still and navigate with other window.


tuhdo:
Nevermind. I would love to have this feature but it needs to be carefully with the cloned buffer. I will revisit when I have time.

PR #1007, opened 2015-04-28
Don't re-define fboundp

lunaryorn:
If `fboundp` is already defined, don't re-define it.  Makes `find-function` jump to its real definition instead this one…


thierryvolpiatto:
Thanks to remind me about this.
We can just remove this alias safely now, it was useful only at the time helm had its autoload definitions inside helm-config, now the autoloads are builded either by make or package.el, we are sure the `function-put` is used only if exists.

PR #1010, opened 2015-05-01
helm-files pattern auto-expansion for user defined path aliases

amarek:
I find the root and home directory auto-expansion in search pattern extremely useful for rapid navigation around the file system. So useful in fact, that I have created a simple extension of the concept that allows for rapid and friction-less navigation to any oft visited directory based on a configured list of aliases. 

It has proven pretty convenient for me so far, so I am sharing it here. Still not sure about possible overlap with the 'bookmark' facility, but as far as I managed to glean from the sources the latter requires some additional interactions and therefore does not provide the same sort of "immediacy" . 

Regards,
Adam

### DIFF
--- a/helm-dabbrev.el
+++ b/helm-dabbrev.el
@@ -46,11 +46,13 @@ Have no effect when `helm-dabbrev-always-search-all' is non--nil."
 (defcustom helm-dabbrev-major-mode-assoc
   '((emacs-lisp-mode . lisp-interaction-mode))
   "Major mode association alist.
-This allow helm-dabbrev searching in buffers with the associated `major-mode'.
+If non--nil this allows helm-dabbrev searching in buffers with the associated `major-mode'.
 e.g \(emacs-lisp-mode . lisp-interaction-mode\)
 will allow searching in the lisp-interaction-mode buffer when `current-buffer'
 is an `emacs-lisp-mode' buffer and vice versa i.e
-no need to provide \(lisp-interaction-mode . emacs-lisp-mode\) association."
+no need to provide \(lisp-interaction-mode . emacs-lisp-mode\) association.
+If nil the `major-mode' of buffers is disregarded and the search happens in all buffers
+(similar behaviour to `dabbrev-expand' with `dabbrev-check-all-buffers' set to non--nil)"
   :type '(alist :key-type symbol :value-type symbol)
   :group 'helm-dabbrev)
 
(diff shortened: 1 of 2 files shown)


thierryvolpiatto:
amarek <notifications@github.com> writes:

> I find the root and home directory auto-expansion in search pattern
> extremely useful for rapid navigation around the file system. So
> useful in fact, that I have created a simple extension of the concept
> that allows for rapid and friction-less navigation to any oft visited
> directory based on a configured list of aliases.
>
> It has proven pretty convenient for me so far, so I am sharing it
> here. Still not sure about possible overlap with the 'bookmark'
> facility, but as far as I managed to glean from the sources the latter
> requires some additional interactions and therefore does not provide
> the same sort of "immediacy" .

Thanks for sharing your code.
However I have recently introduced (as you noticed apparently) the
ability to bookmark helm sessions.
There is no configuration to make, just hit C-x r m when you are in
helm-find-files.
The advantage is you have nothing to set manually.
To jump back to your directory use M-x helm-filtered-bookmarks.
Isn't it enough ?
Have you tried it ?

-- 
Thierry
Get my Gnupg key:

amarek:
Yes, I have tried it a couple weeks back and I really liked the zero config
concept of marking the locations, but I couldn't find a way to integrate
the bookmark selection into the helm files UI.

I guess, the thing I am trying to avoid is having to make a conscious
decision on whether to reach a path via a bookmark or directly, ahead of
invoking the find-file UI. I view a bookmark as a shortcut - a way to
minimize the number of key strokes needed to reach a file, but I will only
use it once I realize that I'm substantially far away from the desired path.

So is there  a way to integrate the helm-filtered-bookmarks as a source
into the helm-files selection UI (similarly to how the previous locations
are accessible  there) ? That would be perfect.

Thanks, Adam

thierryvolpiatto:
amarek <notifications@github.com> writes:

> Yes, I have tried it a couple weeks back and I really liked the zero config
> concept of marking the locations, but I couldn't find a way to integrate
> the bookmark selection into the helm files UI.
>
> I guess, the thing I am trying to avoid is having to make a conscious
> decision on whether to reach a path via a bookmark or directly, ahead of
> invoking the find-file UI. I view a bookmark as a shortcut - a way to
> minimize the number of key strokes needed to reach a file, but I will only
> use it once I realize that I'm substantially far away from the desired path.

Yes I understand this, what you wrote is nice, but I don't think people
will want to go to customize to write aliases.

> So is there a way to integrate the helm-filtered-bookmarks as a source
> into the helm-files selection UI (similarly to how the previous locations
> are accessible there) ? That would be perfect.

Hmm, I don't like the idea of integrating as a source, but a key to
switch to the bookmark source dedicated to helm-find-files, yes, I guess
it is possible, would it fit your need ?

-- 
Thierry
Get my Gnupg key:

amarek:
Yes, that actually sounds good.

And I'm with you on the slim chances that the config based approach would
see a wider adoption - I was actually thinking of capturing the aliases in
a manner similar to the way you capture bookmarks, but I definitely don't
like the redundancy it would create.

thierryvolpiatto:
I have introduced a new command that allow you to switch to bookmark and
to come back to helm-find-files, it is bound to C-x r b in
helm-find-files.
So first hit switch to bookmark and another hit come back to hff.
If you press RET on a bookmark entry and you change your mind and want
to come back to hff just enter "./".
It is in exp branch, have a try! ;-).

-- 
Thierry
Get my Gnupg key:

amarek:
Thanks! Will try that definitely.

thierryvolpiatto:
Great! Closing.

Thanks.

PR #1035, opened 2015-05-27
Make helm-man compatible with man v1.6

itiut:
man v1.6 does not support `-l` option and it can open man-pages only by passing absolute file paths as arguments.
man v2 (man-db) also works without `-l` option when its arguments are absolute paths.

I tested this with man v1.6c on Mac OS X and man v2.7.1 on Arch Linux, and both seem to work well.

### DIFF
--- a/helm-man.el
+++ b/helm-man.el
@@ -57,7 +57,7 @@ source.")
             (let ((file (helm-comp-read
                          "ManFile: " wfiles :must-match t)))
               (if (eq helm-man-or-woman-function 'Man-getpage-in-background)
-                  (manual-entry (format "-l %s" file))
+                  (manual-entry file)
                 (woman-find-file file)))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Yuichi TANIKAWA <notifications@github.com> writes:

> man v1.6 does not support -l option and it can open man-pages only by
> passing absolute file paths as arguments.  man v2 (man-db) also works
> without -l option when its arguments are absolute paths.
>
> I tested this with man v1.6c on Mac OS X and man v2.7.1 on Arch Linux,
> and both seem to work well.

I have commited something different, with a user var that allow using or
not using "-l":

(setq helm-man-format-switches "%s")

Default is still using "-l".

Thanks anyway for the patch and for showing this.

-- 
Thierry
Get my Gnupg key:

itiut:
That's fine. Thank you!

PR #1061, opened 2015-06-13
Change keyword :default to :input in function helm-do-grep-1

syl20bnr:
Hi Thierry,

This PR fixes the keyword `:default` passed to `helm` function in `helm-do-grep-1`, it should be `:input`.

Cheers,
syl20bnr

### DIFF
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -918,7 +918,7 @@ in recurse, search being made on `helm-zgrep-file-extension-regexp'."
     (helm
      :sources 'helm-source-grep
      :buffer (format "*helm %s*" (if zgrep "zgrep" (helm-grep-command recurse)))
-     :default default-input
+     :input default-input
      :keymap helm-grep-map
      :history 'helm-grep-history
      :truncate-lines t)))


thierryvolpiatto:
Sylvain Benner <notifications@github.com> writes:

> Hi Thierry,
>
> This PR fixes the keyword :default passed to helm function in
> helm-do-grep-1, it should be :input.

Thanks for the PR, but the intention is really to use :default here.
This is used only by user building their own grep functions.

You can insert this at prompt with M-n.

-- 
Thierry
Get my Gnupg key:

syl20bnr:
Thank you for the answer.

We are trying to unify all the helm search backends using helm-grep for grep and helm-ag for ag, pt and ack. The idea is to have 2 variants for each search tool and avoid `M-n` which is limited to only the symbol under point:
- one variant with empty pattern (stock one in most cases)
- another one defined at the spacemacs level with as few hack as possible to feed the initial pattern with the current active region or the symbol under point

Not being able to provide our own `:input` to `helm-grep` forces us to redefine `helm-do-grep-1` which is pretty awkward :-)
Being able to provide `:input` allows us to simplify like this (function `spacemacs/helm-do-grep-symbol` vs. `spacemacs/helm-do-grep-region-or-symbol`): https://github.com/syl20bnr/spacemacs/commit/609ce0c

thierryvolpiatto:
Sylvain Benner <notifications@github.com> writes:

> Thank you for the answer.
>
> We are trying to unify all the helm search backends using helm-grep
> for grep and helm-ag for ag, pt and ack.

helm-grep is supporting ack (it is the backend I use).

> The idea is to have 2 variants for each search tool and avoid M-n
> which is limited to only the symbol under point:

It is not. M-n is using symbol under point by default if no :default is
provided, but if you specify :default M-n will use this.(See docstring
of `helm')

So you can use as default for example the region or the symbol at point
if region is not active.

Using this in :input would be annoying if for example a function of 100
lines is marked and you call grep will insert this function in the grep
prompt.

So if you want to use region or symbol at point, call helm-do-grep-1
with the default-input argument using region or symbol and as soon you
press M-n you will have the region (if region-active-p) or the symbol at
pt.

-- 
Thierry
Get my Gnupg key:

thierryvolpiatto:
What I can do is providing a variable that allow using default-as-input.
It was not enabled because helm was not supporting this before 1.7.2 in
async sources.

To see the effect, try helm-occur/moccur or helm-gid.

-- 
Thierry
Get my Gnupg key:

thierryvolpiatto:

> What I can do is providing a variable that allow using default-as-input.
> It was not enabled because helm was not supporting this before 1.7.2 in
> async sources.
>
> To see the effect, try helm-occur/moccur or helm-gid.

I just realize helm-grep is defining helm-source-grep, so what you have
to do to enable this is just:

(add-to-list 'helm-sources-using-default-as-input 'helm-source-grep)

-- 
Thierry
Get my Gnupg key:

thierryvolpiatto:
Here an example of grep with default-as-input enabled:

The point in current-buffer is at helm-sources-using-default-as-input.

screenshot - 13062015 - 11 48 10

Now I launch grep (C-s):

screenshot - 13062015 - 11 48 50

As you can see grep (ack-grep) searched helm-sources-using-default-as-input in all files of helm directory but nothing is in prompt which allow you to insert immediately something new in prompt.

thierryvolpiatto:
So if you really want to use `:input`, I have added the option in addition of `:default`.
I am closing this now, feel free to open a new issue if needed.

Thanks.

syl20bnr:
> So if you really want to use :input, I have added the option in addition of :default.

Excellent, thank you :+1:

PR #1065, opened 2015-06-15
* helm-files.el: Allow browsing svn project when helm-ls-svn is installed

xuchunyang:
helm-ls-svn (http://melpa.org/#/helm-ls-svn) is like helm-ls-git and helm-ls-hg, but works for svn, I write this package and I hope it can be integrated into `helm-browse-project`.


thierryvolpiatto:
Great! Thanks.

PR #1066, opened 2015-06-15
Change link from git to https

joelparkerhenderson:
This change is GitHub's standard practice for public repos. (This change also fixes a glitch for users like me that have multiple private keys)


thierryvolpiatto:
Indeed, thanks!

PR #1079, opened 2015-06-25
Add helm-mode-fuzzy-match.

bling:
this adds a new variable to allow fuzzy matching to be enabled in helm-mode.

the variable name sounds more "global" than it is, but it follows existing naming conventions of other modes, but i'm happy to rename it to something else.  thanks.

### DIFF
--- a/helm-mode.el
+++ b/helm-mode.el
@@ -100,6 +100,11 @@ Affect among others `completion-at-point', `completing-read-multiple'."
   :group 'helm-mode
   :type 'boolean)
 
+(defcustom helm-mode-fuzzy-match nil
+  "Enable fuzzy matching in `helm-mode'."
+  :group 'helm-mode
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Bailey Ling <notifications@github.com> writes:

> this adds a new variable to allow fuzzy matching to be enabled in
> helm-mode.
>
> the variable name sounds more "global" than it is, but it follows
> existing naming conventions of other modes, but i'm happy to rename it
> to something else. thanks.

Ok, but this will not enable fuzzy in C-h f/d you need also a :fuzzy
entry in `helm-completing-read-symbols'.

-- 
Thierry
Get my Gnupg key:

bling:
oops, thanks for spotting the error.

thierryvolpiatto:
Ok thanks.

Issue #1064, opened 2015-06-14
helm completions close to where I type


vspinu:
I would be happy to use helm everywhere instead of completing read and ido if only it could show completions close to where I type the candidates.  I currently have `helm-split-window-in-side-p` set to `t`  and `helm-split-window-default-side` set to  `other` and I  am reasonably happy with it, except of the case when split happens on the right. Then the distance between minibuffer and helm completion is too big.

Is there anything in helm that comes close to either of the following:

1)  helm completion buffer to  contain its own pseudo minibuffer line. Preferably on the top of the completion buffer.  Then distance between the candidates and line where I type is zero - an ideal case.
2) `helm-split-window-default-side` to accept `below-left` value.  In other words, I want helm to always split left or bottom window and display candidates close to minibuffer. 
3) display candidates in the minibuffer (like ivy does)

Thanks.

thierryvolpiatto:
Vitalie Spinu <notifications@github.com> writes:

> I would be happy to use helm everywhere instead of completing read and
> ido if only it could show completions close to where I type the
> candidates. I currently have helm-split-window-in-side-p set to t and
> helm-split-window-default-side set to other and I am reasonably happy
> with it, except of the case when split happens on the right. Then the
> distance between minibuffer and helm completion is too big.

Did you try helm-always-two-windows ?
It is what I use.

Also for special cases if you use helm-reuse-last-window-split-state,
you can move your helm window with C-t and helm will reuse this split
until you change (always with two windows when using
helm-always-two-windows).


-- 
Thierry
Get my Gnupg key:
gpg --keyserver pgp.mit.edu --recv-keys 59F29997

vspinu:
> Did you try helm-always-two-windows ?

Yes I tried but I cannot get used to it because it changes the location of the destination buffer.  It also doesn't seem to play well with other tools that keep track of window configurations. Particularly with sunrise commander.   So I try to achieve the opposite - to keep the window configuration unchanged as much as I can.

michael-heerdegen:
I agree that working into that direction would be good.

Something like a popup as in autocomplete or company also comes to my mind for things like completion at point.

But all that would probably mean a bigger amount of work to do.

thierryvolpiatto:
Michael Heerdegen <notifications@github.com> writes:

> I agree that working into that direction would be good.

You have to understand that a minibuffer can be displayed only at bottom
of frame, unless you use a minibuffer with its own frame, which would
[...]

PR #1080, opened 2015-06-27
Keep using a prompt even with truncated line in header-line (#1064).

thierryvolpiatto:
* helm.el (helm--set-header-line): Do it.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -3822,21 +3822,22 @@ Possible value of DIRECTION are 'next or 'previous."
                           (and (listp source)
                                (assoc-default 'header-line source))
                           source))
-                  (hlend (make-string (max 0 (- (window-width) (length hlstr))) ? )))
+                  (hlend (make-string (max 0 (- (window-width)
+                                                (length hlstr))) ? )))
              (setq header-line-format
                    (propertize (concat " " hlstr hlend) 'face 'helm-header))))))
   (when force (force-mode-line-update)))
 
 (defun helm--set-header-line (&optional update)
   (with-selected-window (minibuffer-window)
-    (let* ((beg (save-excursion (vertical-motion 0) (point))) 
-           (end (save-excursion (end-of-visual-line) (point)))
+    (let* ((beg  (save-excursion (vertical-motion 0) (point))) 
+           (end  (save-excursion (end-of-visual-line) (point)))
            (vprt (concat (propertize helm--prompt 'face 'minibuffer-prompt)
                          "[...]"))
            ;; The visual line where the cursor is.
            (cont (buffer-substring beg end))
            (pref (propertize " " 'display '(space :width left-fringe)))
-           (pos (- (point) beg)))
+           (pos  (- (point) beg)))
       (with-helm-buffer
         (unless (string-match helm--prompt cont)
           (setq pref (propertize " " 'display (concat pref vprt))
@@ -3846,7 +3847,7 @@ Possible value of DIRECTION are 'next or 'previous."
                                      (substring cont (length vprt))))))
         (setq header-line-format (concat pref cont " "))
         (put-text-property
-         ; Increment pos to handle the space before prompt (i.e `pref').
+         ;; Increment pos to handle the space before prompt (i.e `pref').
          (1+ pos) (+ 2 pos)
          'face 'cursor header-line-format)
         (when update (force-mode-line-update))))))
--- a/helm.el
+++ b/helm.el
@@ -666,6 +666,11 @@ input method with `toggle-input-method'."
   "Face used to highlight matches."
   :group 'helm)
 
+(defface helm-header-line-left-margin
+  '((t (:foreground "black" :background "yellow")))
+  "Face used to highlight helm-header sign in left-margin."
+  :group 'helm)
+
 
 ;;; Variables.
 ;;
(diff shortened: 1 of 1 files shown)


michael-heerdegen:
Tried it only shortly, will use it the whole day for testing.  Didn't look at the code yet.

My first impression is good.

But there are big problems.  The cursor vanishes for long inputs (when the [...] appeared) when hitting C-b.  When I hit C-f, I get an "end of buffer" error and the input is in nirvana.

Sometimes, the whole thing doesn't work.  If I hit C-f and keep pressing the "a" key, the minibuffer expands to multiple lines, and the header line doesn't show the [...] as expected.

Sometimes, I just get an error, and my input is gone.

michael-heerdegen:
Sorry, that the [...] doesn't appear and as well as the expanding minibuffer was tested with the old code, Doesn't happen with yours.

michael-heerdegen:
Another, minor, problem: if I just for testing, hit C-x C-f and keep hitting the "a" key, I see that after the ellipsis appeared, the second line will only hold a twenty chars or so until it is truncated again.

thierryvolpiatto:

> Tried it only shortly, will use it the whole day for testing. Didn't look at the code yet.
>
> My first impression is good.
>
> But there are big problems. The cursor vanishes for long inputs (when the [...] appeared) when hitting C-b. When I hit C-f, I get an "end of buffer" error and the
> input is in nirvana.
>
> Sometimes, the whole thing doesn't work. If I hit C-f and keep pressing the "a" key, the minibuffer expands to multiple lines, and the header line doesn't show the
> [...] as expected.
>
> Sometimes, I just get an error, and my input is gone.

Thanks for trying it.
Yes I have noticed all these problems and I have commited something
different, perhaps I will retain this new version, but let's try it
first.

-- 
Thierry
Get my Gnupg key:

michael-heerdegen:

> Yes I have noticed all these problems and I have commited something
> different, perhaps I will retain this new version, but let's try it
> first.

Ok.  Yes, seems to work better indeed.

One thing: if I try "helm-eval-expression", you get no prompt at all
with that version, the header-line doesn't mirror the minibuffer, but
the minibuffer is hidden as well.  With other words, you don't see what
you type, nowhere.

thierryvolpiatto:

>
>> Yes I have noticed all these problems and I have commited something
>> different, perhaps I will retain this new version, but let's try it
>> first.
>
> Ok. Yes, seems to work better indeed.

Good thanks.

> One thing: if I try "helm-eval-expression", you get no prompt at all
> with that version, the header-line doesn't mirror the minibuffer, but
> the minibuffer is hidden as well. With other words, you don't see what
> you type, nowhere.

It is fixed, I guess you have just evaled helm--set-header-line, load
the whole file instead or eval helm-read-pattern-maybe (or reinstall).

-- 
Thierry
Get my Gnupg key:

michael-heerdegen:
Michael Heerdegen <michael_heerdegen@web.de> writes:

> One thing: if I try "helm-eval-expression", you get no prompt at all
> with that version, the header-line doesn't mirror the minibuffer, but
> the minibuffer is hidden as well.  With other words, you don't see what
> you type, nowhere.

Oops - that's my own fault, of course.  I was testing with an outdated
version of my `helm-hide-minibuffer-maybe'.  Forget this.

thierryvolpiatto:

> Michael Heerdegen <michael_heerdegen@web.de> writes:
>
>> One thing: if I try "helm-eval-expression", you get no prompt at all
>> with that version, the header-line doesn't mirror the minibuffer, but
>> the minibuffer is hidden as well. With other words, you don't see what
>> you type, nowhere.
>
> Oops - that's my own fault, of course. I was testing with an outdated
> version of my `helm-hide-minibuffer-maybe'. Forget this.

Ok, do you think this version is good enough to merge ?
You will be able to change details later if you want (face, sign ...).

-- 
Thierry
Get my Gnupg key:

thierryvolpiatto:
BTW You have now write access to helm repo, you can create branches,
make changes etc ... (be careful ;-).
Of course you can still make PR from your fork, but you have better time
to make a new branch and make the PR from this branch.

-- 
Thierry
Get my Gnupg key:

michael-heerdegen:

> BTW You have now write access to helm repo, you can create branches,
> make changes etc ... (be careful ;-).  Of course you can still make PR
> from your fork, but you have better time to make a new branch and make
> the PR from this branch.

Ok, thanks.  I'll first have a look at the links github sent me as a
side affect.

michael-heerdegen:

> Ok, do you think this version is good enough to merge ?

I think so, the defun is more or less finished.

Question: Would you prefer adding "helm-hide-minibuffer-maybe" to the
hook by default?  We could make "helm-echo-input-in-header-line" accept
three values (one more meaning "show in header line and hide
minibuffer"), or add a new var controlling this.  Then users not
mastering Elisp would be able to use this feature.

thierryvolpiatto:

>
>> Ok, do you think this version is good enough to merge ?
>
> I think so, the defun is more or less finished.

Ok, you can perhaps merge yourself the PR (click on the button) to see
if all works well.

> Question: Would you prefer adding "helm-hide-minibuffer-maybe" to the
> hook by default? We could make "helm-echo-input-in-header-line" accept
> three values (one more meaning "show in header line and hide
> minibuffer"), or add a new var controlling this. Then users not
> mastering Elisp would be able to use this feature.

Ok for this, go ahead.

-- 
Thierry
Get my Gnupg key:

michael-heerdegen:

> Ok for this, go ahead.

Sorry, I was short on time the last days, and will be on a tour soon.
I'll do this ASAP.  I should be online on the trip, unless I have bad
luck.

FWIW, this is the current version of the defun that should go to
"helm-minibuffer-set-up-hook":

  (defun helm-hide-minibuffer-maybe ()
    "Hide minibuffer contents in a Helm session.
This function should normally go to `helm-minibuffer-set-up-hook'.
It has no effect if `helm-echo-input-in-header-line' is nil."
    (when (buffer-local-value 'helm-echo-input-in-header-line
                              (get-buffer (helm-buffer-get)))
      (let ((ov (make-overlay (point-min) (point-max) nil nil t)))
        (overlay-put ov 'window (selected-window))
        (overlay-put ov 'face (let ((bg-color (face-background 'default nil)))
                                `(:background ,bg-color :foreground ,bg-color)))
        (setq-local cursor-type nil))))

thierryvolpiatto:

>
>> Ok for this, go ahead.
>
> Sorry, I was short on time the last days, and will be on a tour soon.
> I'll do this ASAP. I should be online on the trip, unless I have bad
> luck.
>
> FWIW, this is the current version of the defun that should go to
> "helm-minibuffer-set-up-hook":

Thanks, added with small modifications.

-- 
Thierry
Get my Gnupg key:

michael-heerdegen:
Thanks.

PR #1081, opened 2015-06-27
require helm-utils for helm-org.el

dfeich:
helm-org-agenda-files-headings needs helm-basename defined

Discovered with joy that there exists `helm-org-agenda-files-headings`. Works like a charm, except in my freshly opened session I get

helm-get-org-candidates-in-file: Symbol's function definition is void: helm-basename

Thanks again for helm!
Derek


thierryvolpiatto:
Thanks.

PR #1085, opened 2015-06-30
Escape regex characters in `helm--prompt`

randymorris:
I ran into this issue after the "->" was added because helm-projectile uses "[project] pattern:" for its prompt.  When `string-match` does this comparison, the "[project]" part is treated as a set of characters rather than a raw string, thus always failing the condition in the `if`.

I've changed the test to treat `helm--prompt` as a raw string as well as changed it to use `string-match-p` since you aren't using the returned index.


thierryvolpiatto:
Ok thanks.

PR #1091, opened 2015-07-09
* helm-locate.el: require helm-types

bsuh:
`helm-locate-source` parent class is `helm-type-file` defined in `helm-types`


thierryvolpiatto:
Indeed thanks!

PR #1117, opened 2015-07-28
Fix helm-next-source when have empty dummy source

xuchunyang:
The problem, which I'm trying to solve, is described at http://emacs.stackexchange.com/questions/14286/how-to-cycle-through-helm-sources


xuchunyang:
oop, this also breaks the old good function.

xuchunyang:
It works, thanks!

PR #1125, opened 2015-08-09
Fix crash in llvm-mode

Wilfred:
Currently, helm crashes in llvm-mode due to assuming that the mode map is never nil. This PR fixes that.

I've also renamed a variable which confused me slightly, let me know what you think.

### DIFF
--- a/helm-command.el
+++ b/helm-command.el
@@ -67,10 +67,11 @@ Show all candidates on startup when 0 (default)."
 
 (cl-defun helm-M-x-get-major-mode-command-alist (mode-map)
   "Return alist of MODE-MAP."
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Wilfred Hughes <notifications@github.com> writes:

> Currently, helm crashes in llvm-mode due to assuming that the mode map is never nil. This PR fixes that.

Looks good.

> I've also renamed a variable which confused me slightly, let me know what you think.

Looks good, perhaps you have added the message for debugging purpose ?
Can you remove it.

Thanks.

-- 
Thierry
Get my Gnupg key:

Wilfred:
Woops, `message` removed.

thierryvolpiatto:
Thanks, merged.

Issue #1137, opened 2015-08-24
helm-ucs is slow 


robbyoconnor:
`helm-ucs` is slow when compared to [helm-unicode][] -- this is not a good thing. I am using **24.4.1**.  It's not at all performant -- I noticed this when testing for https://github.com/syl20bnr/spacemacs/pull/2728 -- and it was really, really slow, whereas [helm-unicode][] was near instantaneous. 

[helm-unicode]: https://github.com/shosti/helm-unicode

thierryvolpiatto:
Should be fixed now.

robbyoconnor:
:+1: thank you!

OrionRandD:
I am getting this error 
collecting ucs names...99% 
sort: Symbol’s function definition is void: helm-generic-sort-fn [2 times]
Quit
sort: Symbol’s function definition is void: helm-generic-sort-fn

when calling 
helm-ucs

thierryvolpiatto:
OrionRandD ***@***.***> writes:

> 1.  ( ) text/plain          (*) text/html           
>
> I am getting this error
> collecting ucs names...99%
> sort: Symbol’s function definition is void: helm-generic-sort-fn [2 times]
> Quit
> sort: Symbol’s function definition is void: helm-generic-sort-fn

This is not reproductible, double check your installation following the
instructions and if it persists open a new bug report starting from
emacs-helm.sh.

Thanks.

-- 
Thierry

PR #1138, opened 2015-08-24
remove candidate-number-limit from helm-source-ucs

cmccloud:
The load time for helm-source-ucs is unreasonably high, particularly on
OSX. Tested with a global candidate-number-limit of 100, load time is
greatly improved.

See #1137 and shosti/helm-unicode#2

### DIFF
--- a/helm-font.el
+++ b/helm-font.el
@@ -152,7 +152,6 @@ Only math* symbols are collected."
 (defvar helm-source-ucs
   (helm-build-in-buffer-source "Ucs names"
     :init #'helm-ucs-init
-    :candidate-number-limit 9999
     :help-message 'helm-ucs-help-message
     :filtered-candidate-transformer
     (lambda (candidates _source) (sort candidates #'helm-generic-sort-fn))


CestDiego:
Is it faste than helm-unicode though ?

robbyoconnor:
If it's not faster than `helm-unicode` -- it's still insufficient for spacemacs but otherwise good for others :)

cmccloud:
@thierryvolpiatto Thanks so kindly for the improvements to helm-ucs - I was about to make a few more changes but it looks like you got them. Two things that I was considering adding, that I would be curious to hear your thoughts on:

1) Sorting the ucs-names data on cache creation. I thought about sorting within helm-ucs-init, and exposing the sorting function as a variable to make it easily customizable. 
2) Adding a tiny delay to the source

Thanks again, the changes look great.

thierryvolpiatto:
Christopher McCloud <notifications@github.com> writes:

> 1) Sorting the ucs-names data on cache creation. I thought about
> sorting within helm-ucs-init, and exposing the sorting function as a
> variable to make it easily customizable.

Is is really necessary to make an initial sort ?
Sorting will happen as soon as you start typing something.
BTW the sorting function actually used is not very good for ucs, it is a
generic function initially used to sort symbols, probably a better one
would be used.

> 2) Adding a tiny delay to the source

Why ?


Thanks.

-- 
Thierry
Get my Gnupg key:

cmccloud:
Re: Sorting - I don't think it's necessary. One thing I appreciated about helm-ucs with a candidate limit of 9999 was that it made it much easier to explore the range of available unicode symbols. I was thinking that an inital sort might let you order the ucs data by 'category'  to aid discoverability. But again, I don't think it's necessary at all. 

Re: Delay - Good question. I had thought it might reduce the stutter that sometimes occurs (at least for me) when updating the search, but now I'm thinking that it's caused by emacs rendering rather than helm.

cmccloud:
@thierryvolpiatto Thank you again for being so quick to respond with fixes. Unless you think there is anything else we can close this issue.

thierryvolpiatto:
Thanks for your help, closing now.

Issue #1141, opened 2015-08-26
helm-documentation cleanup and key binding


hmelman:
I've been using helm for a few weeks now and only just discovered helm-documentation. It seems quite useful and more helpful for trying to learn all of helm than just using C-h m.

1. I think it should have its own key binding, perhaps `helm-command-prefix-key ?'
2. The helm buffer of the outline is a bit wide. In the org buffer I shortened the two star headings for tips and specific commands to just "Tips" and "Commands" and I feel it helped a lot. The outline structure provided context so it wasn't confusing between sections.
3. There were a few places where the org buffer had double blank lines. Collapsing these to single blank lines made the helm-doc.org buffer much cleaner IMHO when showing just the outline with org-shifttab.

thierryvolpiatto:
Howard Melman <notifications@github.com> writes:

> I've been using helm for a few weeks now and only just discovered
> helm-documentation. It seems quite useful and more helpful for trying
> to learn all of helm than just using C-h m.


>  1. I think it should have its own key binding, perhaps
>  `helm-command-prefix-key ?'

Ok, needs also an entry in menu.

>  2. The helm buffer of the outline is a bit wide. In the org buffer I
>     shortened the two star headings for tips and specific commands to
>     just "Tips" and "Commands" and I feel it helped a lot. The outline
>     structure provided context so it wasn't confusing between
>     sections.

Agreed.

>  3. There were a few places where the org buffer had double blank
>     lines. Collapsing these to single blank lines made the
>     helm-doc.org buffer much cleaner IMHO when showing just the
>     outline with org-shifttab.

Instead of modifying in helm-doc.org generated by helm-documentation,
could you remove these newlines in the helm helm strings located in
helm-help.el ?
To see the result of your changes once evaled the strings (C-M x) rerun
helm-documentation with a prefix arg to regenerate helm-doc.org.
Once you are happy with your changes, perhaps you can send a PR, would
be great (same for 1 and 2), don't hesitate to ask for help on this.

Thanks for your feedback.

-- 
Thierry
Get my Gnupg key:
gpg --keyserver pgp.mit.edu --recv-keys 59F29997

hmelman:
Thanks. I'll do the cleanup of the doc strings in helm-help and submit a PR. I'm learning git and github so this seems like an easy first experiment. :)

thierryvolpiatto:
Howard Melman <notifications@github.com> writes:

[...]

PR #1152, opened 2015-09-06
Help cleanup

hmelman:
This is for issue #1141 

I changed the ** headings for tips and specific commands to just be Tips and Commands. I removed a trailing colon from them and from some other headings. 

All the help messages begin with a single \n but I changed the others to be actual newlines so it's easier to see in the source. Headings now have a blank line before and after, though I left some **** headings with no following blank line and they probably should have one. 

Help messages end without a newline. I changed helm-documentation to insert a \n after each one when it's inserted. It seems embedding a keymap with \<> inserts blank lines before and after, so those messages that end with one (esh-on-file, semantic and kmacro) all have an extra trailing blank line.

There should probably be a general policy about how headings end (I'd recommend no punctuation).
helm-help-message should be cleaned up too
and helm-help should probably insert a \n or two before inserting helm-help-message.
Finally I think there should be some standard to how example blocks are formatted. I think they should be indented 4 spaces.

### DIFF
--- a/helm-help.el
+++ b/helm-help.el
@@ -76,7 +76,8 @@ Find here the documentation of all sources actually documented."
       (cl-loop for elm in helm-help--string-list
                for str = (symbol-value elm)
                do (insert (substitute-command-keys
-                           (if (functionp str) (funcall str) str))))))
+                           (if (functionp str) (funcall str) str))
+                          "\n"))))
   (let ((helm-org-headings--nofilename t))
     (helm :sources (helm-source-org-headings-for-files
                     (list helm-documentation-file))
--- a/helm-help.el
+++ b/helm-help.el
@@ -89,13 +89,13 @@ Find here the documentation of all sources actually documented."
 ;;
 ;;
 (defvar helm-buffer-help-message
-  "\n* Helm Buffer\n
+  "\n* Helm Buffer
 
-** Helm buffers tips:
+** Tips
 
-*** Completion:
+*** Completion
 
-**** Major-mode:
+**** Major-mode
 
 You can enter a partial name of major-mode (e.g lisp, sh) to narrow down buffers.
 To specify the major-mode, prefix it with \"*\" e.g \"*lisp\".
@@ -107,14 +107,14 @@ fundamental-mode.
 
 Enter then a space and a pattern to narrow down to buffers matching this pattern.
 
-**** Search inside buffers:
+**** Search inside buffers
 
 If you enter a space and a pattern prefixed by \"@\" helm will search for text matching
 this pattern INSIDE the buffer (i.e not in the name of buffer).
 NOTE that if you enter your pattern prefixed with \"@\" but escaped, helm will search a buffer
 matching \"@pattern\" but will not search inside.
 
-**** Search by directory name:
+**** Search by directory name
 
 If you prefix the beginning of pattern with \"/\" the match will occur on directory name
 of buffer, it is interesting to narrow down to one directory for example, subsequent string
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Many thanks for your work.

hmelman:
Let me know your thoughts on the last paragraph of my note and I'll tackle that next.

thierryvolpiatto:
Howard Melman <notifications@github.com> writes:

> There should probably be a general policy about how headings end (I'd
> recommend no punctuation).

Agreed.
Probably these rules should be added to helm wiki (it is public writable
now).

> helm-help-message should be cleaned up too and helm-help should
> probably insert a \n or two before inserting helm-help-message.

Agreed.

> Finally I think there should be some standard to how example blocks
> are formatted. I think they should be indented 4 spaces.

Agreed.


Thanks for your help on this.

-- 
Thierry

Get my Gnupg key:

hmelman:
I have another. In lists of commands typically the code has "\t\t->". I'd like to replace this which just one embedded TAB. Sometimes the code has just one tab so unifying that is cleaner. To my eye the -> just adds clutter. With just one TAB it looks more like what emacs generates for keymap lists.

hmelman:
Would you consider changing the mode of *helm-help* buffer to org-mode? 

Being able to use org tables would make things much nicer (particularly the Basic Operations list in helm-help-message). org-mode has been included with Emacs since 22.1 which shipped in June 2007 (which is what is currently shipped on OS X for what's that worth).

Something I'm running into is that I use a proportional font for text modes, so that includes org and outline, but org-table-mode knows to use a fixed-width font. I doubt I'm the only one with this problem. Emacs' help-mode derives from special so that stays a fixed-width font for me, so it's only with the helm help that I run into this issue.

thierryvolpiatto:
Howard Melman <notifications@github.com> writes:

> Would you consider changing the mode of helm-help buffer to org-mode?

Done.

-- 
Thierry

Get my Gnupg key:

thierryvolpiatto:
Howard Melman <notifications@github.com> writes:

> I have another. In lists of commands typically the code has
> "\t\t->". I'd like to replace this which just one embedded
> TAB. Sometimes the code has just one tab so unifying that is
> cleaner. To my eye the -> just adds clutter. With just one TAB it
> looks more like what emacs generates for keymap lists.

I have no objections if you change the \t\t-> for something else.

-- 
Thierry

Get my Gnupg key:

hmelman:
Is there a reason helm-help-message is defined using (lambda () (concat "...")) instead of just being a string? I doesn't seem so to me, but I might be misunderstanding helm-interpret-value.

thierryvolpiatto:
Howard Melman <notifications@github.com> writes:

> Is there a reason helm-help-message is defined using (lambda ()
> (concat "...")) instead of just being a string?

Because it was using in the past concat, so it needed to be evaled,
it is no more needed now I have removed it.

I also removed one tab in commands to align properly.

Is it really needed to add the tabs in help-messages literally instead
of using "\t\t" ?
As all the source code files in helm use as a general rule 
indent-tabs-mode => nil, probably it would be cleaner to use "\t\t".

-- 
Thierry

Get my Gnupg key:

hmelman:
Really needed? No. In the same way the tabs make the rendered help more readable, I think they make the source more readable. It's easier to see the two columns. I can change them back to \t\t if you want.

thierryvolpiatto:
Howard Melman <notifications@github.com> writes:

> Really needed? No. In the same way the tabs make the rendered help
> more readable, I think they make the source more readable. It's easier
> to see the two columns.

Hmm, no, at least for me unless I launch whitespace-mode.
The "\t\t" make the intention clear to the reader of the source code
file.

> I can change them back to \t\t if you want.

I would prefer.

Otherwise now the documentation looks really nice with your changes,
thanks.

-- 
Thierry
Get my Gnupg key:

PR #1154, opened 2015-09-08
Helm help

hmelman:
More help cleanup.

I fixed up helm-help-message.
removed remaining \n's in docstrings and now include two when concatenating help messages.
\t\t-> is now two embedded tabs
in helm-help-message I use an org table for the basic commands. it seemed the cleanest way to multiple keybindings for some commands. I added a note that these are the default bindings since they won't be updated when generating help.
I changed "e.g " to "e.g. " and cleaned up the english in a couple of places.

### DIFF
--- a/helm-help.el
+++ b/helm-help.el
@@ -77,7 +77,7 @@ Find here the documentation of all sources actually documented."
                for str = (symbol-value elm)
                do (insert (substitute-command-keys
                            (if (functionp str) (funcall str) str))
-                          "\n"))))
+                          "\n\n"))))
   (let ((helm-org-headings--nofilename t))
     (helm :sources (helm-source-org-headings-for-files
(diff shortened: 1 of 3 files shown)


thierryvolpiatto:
Howard Melman <notifications@github.com> writes:

> More help cleanup.
>
> I fixed up helm-help-message.  removed remaining \n's in docstrings
> and now include two when concatenating help messages.  \t\t-> is now
> two embedded tabs in helm-help-message I use an org table for the
> basic commands. it seemed the cleanest way to multiple keybindings for
> some commands. I added a note that these are the default bindings
> since they won't be updated when generating help.  I changed "e.g " to
> "e.g. " and cleaned up the english in a couple of places.

Great, looks very nice now.

Thanks.

-- 
Thierry

Get my Gnupg key:

PR #1155, opened 2015-09-08
Don't add font-lock keywords to lisp-interaction-mode

xuchunyang:
lisp-interaction-mode is derived from emacs-lisp-mode, so adding
these font-lock keywords to emacs-lisp-mode applies in both modes.

### DIFF
--- a/helm-config.el
+++ b/helm-config.el
@@ -135,23 +135,22 @@
 
 
 ;;; Fontlock
-(cl-dolist (mode '(emacs-lisp-mode lisp-interaction-mode))
-  (font-lock-add-keywords
-   mode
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:

> lisp-interaction-mode is derived from emacs-lisp-mode, so adding
> these font-lock keywords to emacs-lisp-mode applies in both modes.

It doesn't for some reasons (emacs bug ?).
It is why I have added to both.

-- 
Thierry
Get my Gnupg key:

xuchunyang:
> It doesn't for some reasons (emacs bug ?).
> It is why I have added to both.

I did't know it, maybe in some condition, I didn't test it carefully.

PR #1172, opened 2015-09-16
Fix helm-fuzzy-matching-highlight-fn not called.

bsuh:
`helm-fuzzy-matching-highlight-fn` is exposed as a customizable variable, but the default value is always called.


thierryvolpiatto:
Good catch, thanks.

PR #1181, opened 2015-09-19
helm-select-xfont: Preview font without timer

alphapapa:
* helm-font.el:
  + (helm-source-xfonts):
    - Restore the font when the candidates buffer is closed rather than with
      a 2-second timer.
    - Minor adjustments to help messages for clarity.
  + (helm-persistent-xfont-action): Remove function.

I haven't contributed to Helm before, so I tried to format the PR like recent commits.  Thanks for your work on Helm.

### DIFF
--- a/helm-font.el
+++ b/helm-font.el
@@ -35,30 +35,30 @@
 ;;; Xfont selection
 ;;
 ;;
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Looks good, thanks.

Issue #1175, opened 2015-09-17
Add migemo support


thierryvolpiatto:
I have added a PR at #1174 please migemo users test as much as possible.
You can test with helm-occur among others which use by default the migemo support.
I tested also with helm-kill-ring which works also.
The only dependencies needed is migemo.el and of course a migemo backend.
The only one I found was cmigemo and for me migemo.el was broken with this backend (see 
https://github.com/emacs-jp/migemo/issues/21).
I also lack of examples to fully test this, if one can provide examples that would be great.

Thanks.

thierryvolpiatto:
Now merged.

thierryvolpiatto:
No body interested, closing.

momomo5717:
Thank you for the migemo support! It is a big help for migemo users.


I'd like to use `helm-mini` with migemo.
Could a function like the following be added to use predefined sources with migemo?

```emacs-lisp
(defun helm-migemo-migemize-sorce (source)
  "Migemize SOURCE."
  (let* ((source (cond ((listp source) source)
                       ((symbolp source) (symbol-value source))
                       (t nil)))
         (match (assq 'match source))
         (migemo (assq 'migemo source)))
    (when match
      (unless (or (member 'helm-mm-3-migemo-match match)
                  (eq (cdr match) 'helm-mm-3-migemo-match))
        (setcdr match
                (if (functionp (cdr match))
                    (list (cdr match) 'helm-mm-3-migemo-match)
                  (append (cdr match) (list 'helm-mm-3-migemo-match)))))
      (unless migemo
        (nconc source (list '(migemo))))))
  source)
```

The function is used like the following.

``` emacs-lisp
;; Setting for helm-mini with migemo
(require 'helm-files)
(require 'helm-buffers)
(unless helm-source-buffers-list
  (setq helm-source-buffers-list
        (helm-make-source "Buffers" 'helm-source-buffers)))
(dolist (s helm-mini-default-sources)
  (helm-migemo-migemize-sorce s))
```

Thank you.

P.S. Sorry for the delayed response. It took me some time to dig it.

thierryvolpiatto:
momomo5717 <notifications@github.com> writes:

> Thank you for the migemo support! It is a big help for migemo users.
>
> I'd like to use helm-mini with migemo.
> Could a function like the following be added to use predefined sources with migemo?

No this will not work.
It is working because you still have a helm-migemo.el somewhere, now
things are different, it's much simpler to enable:

;; for buffers
[...]

PR #1184, opened 2015-09-23
Highlight migemo matches (#1175).

thierryvolpiatto:
* helm-grep.el (helm-grep-highlight-match): Do it.
* helm-multi-match.el (helm-mm-migemo-string-match):
The cache is now a full alist containing all patterns.
Allow references for highlighting and also don't compute twice
pattern with migemo.
* helm.el (helm-fuzzy-default-highlight-match): Use it.

### DIFF
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -1014,14 +1014,18 @@ in recurse, and ignoring EXTS, search being made on
         (with-temp-buffer
           (insert (propertize str 'read-only nil))
           (goto-char (point-min))
-          (cl-loop for reg in (if multi-match
-                                  ;; (m)occur.
-                                  (cl-loop for r in (helm-mm-split-pattern
-                                                     helm-pattern)
-                                           unless (string-match "\\`!" r)
-                                           collect r)
-                                  ;; async sources (grep, gid etc...)
-                                  (list helm-input))
+          (cl-loop for reg in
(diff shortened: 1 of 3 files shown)


momomo5717:
Thank you for the implementation!

When pattern has more than one word, highlight doesn't work.

When I tried to use `search-forward-regexp` instead of `search-forward` in `helm-fuzzy-default-highlight-match`, it works.

And I suppose `(concat  (assoc-default p helm-mm--previous-migemo-info) "\\|" p)` would be needed in helm-fuzzy-default-highlight-match.

Thank you.

momomo5717:
Sorry I made typos. "I tried using `re-search-forward`" is right.

Thank you.

thierryvolpiatto:
momomo5717 <notifications@github.com> writes:

> When I tried to use search-forward-regexp instead of search-forward in
> helm-fuzzy-default-highlight-match, it works.

Of course, I forget to replace search-forward by re-search-forward.

> And I suppose (concat (assoc-default p helm-mm--previous-migemo-info)
> "\\|" p) would be needed in helm-fuzzy-default-highlight-match.

Yes, you are right, done too.

Thanks for your help on this.

-- 
Thierry
Get my Gnupg key:

momomo5717:
Thank you again for the migemo support!

PR #1185, opened 2015-09-25
migemo support for helm-moccur-action

tam17aki:

PR #1186, opened 2015-09-25
 Allow migemo+regex in helm-buffers--match-from-inside (#1175).

momomo5717:
Hi,

I'd like to use `@` form in `helm-mini`.
Could `helm-mm-migemo-forward` be used in `helm-buffers--match-from-inside`?

And it seems `helm-mm-migemo-forward` could be better to allow regex in `helm-moccur-action` (#1185).

BTW, should I separate the PR?

Thank you.

P.S. It seems the following declarations could be needed to suppress warnings in helm-regexp and helm-multi-match.

For helm-regexp
```emacs-lisp
(declare-function helm-grep-split-line "helm-grep")
(declare-function helm-grep-highlight-match "helm-grep")
(declare-function helm-buffer-list "helm-buffers")
(declare-function helm-comp-read "helm-mode")
```

For helm-multi-match
```emacs-lisp
(declare-function helm-buffer-get "helm")
```

### DIFF
--- a/helm-buffers.el
+++ b/helm-buffers.el
@@ -523,7 +523,9 @@ i.e same color."
         (with-current-buffer buf
           (save-excursion
             (goto-char (point-min))
-            (re-search-forward regexp nil t)))
+            (if helm-migemo-mode
+                (helm-mm-migemo-forward regexp nil t)
+             (re-search-forward regexp nil t))))
         t)))
 
 (defun helm-buffers--match-from-directory (candidate)
--- a/helm-multi-match.el
+++ b/helm-multi-match.el
@@ -290,9 +290,10 @@ i.e the sources which have the slot :migemo with non--nil value."
   (let ((regex (if (delq 'ascii (find-charset-string word))
                    word
                    (concat (migemo-search-pattern-get word) "\\|" word))))
-    (with-helm-buffer
-      (setq helm-mm--previous-migemo-info
-            (push (cons word regex) helm-mm--previous-migemo-info)))
+    (unless (assoc word helm-mm--previous-migemo-info)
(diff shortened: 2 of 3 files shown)


thierryvolpiatto:
momomo5717 <notifications@github.com> writes:

> P.S. It seems the following declarations could be needed to suppress warnings in helm-regexp and helm-multi-match.
>
> For helm-regexp
>
> (declare-function helm-grep-split-line "helm-grep")
> (declare-function helm-grep-highlight-match "helm-grep")
> (declare-function helm-buffer-list "helm-buffers")
> (declare-function helm-comp-read "helm-mode")
>
> For helm-multi-match
>
> (declare-function helm-buffer-get "helm")

I can't reproduce these warnings, do they appear when you compile with
make ?

-- 
Thierry

momomo5717:
>I can't reproduce these warnings, do they appear when you compile with make ?

When I use `auto-async-byte-compile` on the buffer, the warnings occurs.

Thank you.

thierryvolpiatto:
momomo5717 <notifications@github.com> writes:

>     I can't reproduce these warnings, do they appear when you compile with make ?
>    
> When I use auto-async-byte-compile on the buffer, the warnings occurs.

Yes, that's normal, if you use make or if you async-compile the whole
helm directory it should not happen.
(Should not happen when compiled from package.el).

Thanks.

-- 
Thierry
Get my Gnupg key:

momomo5717:
Thank you for the reply.

>Yes, that's normal, if you use make or if you async-compile the whole
>helm directory it should not happen.
>(Should not happen when compiled from package.el).

Sorry for my lack of knowledge about that.

Thank you for explaining it.

Issue #1193, opened 2015-09-30
helm-recentf's persistent-help is wrong: C-j: helm-type-file-actions (keeping session)


xuchunyang:
Persistent help message show action list symbol-name.

thierryvolpiatto:
Should work now.

Thanks.

xuchunyang:
Thanks.  I've tried with your latest commit and it works fine.

PR #1194, opened 2015-09-30
Add helm-type-file to helm-recentf-source's superclass list (#1193).

xuchunyang:


thierryvolpiatto:
Thanks, you forgot to remove actions from class, I am doing it.

PR #1195, opened 2015-10-01
Add helm-gid-program customization option

xuchunyang:
On Mac OS X, "gid" might be the name of id(1) from GNU coreutils,
so add option to allow user to customize the name of gid(1).

As a side note: when I tried `helm-gid` at the beginning ("gid" is not the one from id-utils), `helm-gid` just didn't show any candidates regardless of any input (I thought it's a bug of helm), so we should remind OSX user to customize the option when someone documents `helm-id-utils` in helm wiki.

### DIFF
--- a/helm-id-utils.el
+++ b/helm-id-utils.el
@@ -24,6 +24,15 @@
   "ID-Utils related Applications and libraries for Helm."
   :group 'helm)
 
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Great, thanks.

PR #1199, opened 2015-10-02
Avoid fake cursor in header-line hiding characters

### LINKED ISSUE #1198: Improve fake cursor in header-line?
When `(setq helm-echo-input-in-header-line t)`, for the current cursor's face, it's better to make it transparent when cursor is on a char. 

And someone also wants to customize that face (see http://emacs.stackexchange.com/questions/17058/change-cursor-type-in-helm-header-line), for example, use the face of cursor with `(setq cursor-type 'bar)`.

michael-heerdegen:
This fixes #1198.  The fake cursor hided the character underneath when
the cursor color and the default foreground color were similar.

I can't imagine a much more elegant solution.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -3802,7 +3802,11 @@ Possible value of DIRECTION are 'next or 'previous."
         (put-text-property
          ;; Increment pos to handle the space before prompt (i.e `pref').
          (1+ pos) (+ 2 pos)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks!

PR #1201, opened 2015-10-06
Use helm-completing-read-symbols for describe-symbol

xuchunyang:
The `describe-symbol' command is available from 25.1.


thierryvolpiatto:
Thanks, didn't know `describe-symbol`,  I use rarely 25.1 actually.

Issue #1212, opened 2015-10-11
Make helm-find-files's action find-file-other-window (C-c o) handle multi selection?


xuchunyang:
Someone wants it to work according to [this question](http://emacs.stackexchange.com/q/14076/3889)'s second comment.

thierryvolpiatto:
Chunyang Xu <notifications@github.com> writes:

> Someone wants it to work according to this question's second comment.

He have to use C-u RET after marking files.
The split is done with `dired-simultaneous-find-file'.

-- 
Thierry
Get my Gnupg key:
gpg --keyserver pgp.mit.edu --recv-keys 59F29997

xuchunyang:
Now I am not sure what he wants, but I though `helm-find-files`'s `C-c o` should have similar function to `helm-buffers-list`'s `C-c o`, i.e., keep `helm-current-buffer`'s window and open marked files with other new windows.

blaenk:
Hey @xuchunyang thanks for creating this! In fact, I had already created an issue (#1100) about this, and you actually [helped me with it](https://github.com/emacs-helm/helm/issues/1100#issuecomment-128939418)! It works great! The only annoyance is that I have to add this for every helm command that I want it in.

Is the solution you gave me there different from what I was asking here? I honestly don't remember.

xuchunyang:
> Is the solution you gave me there different from what I was asking here? I honestly don't remember.

@blaenk yes, there are a little different, that solution was for opening *buffers*, while you were asking for opening *files*, though it should be very easy to change that solution to fit your needs.

Anyway, now `C-c o` now can handle multi selection since #1213 has been merged.

PR #1213, opened 2015-10-12
Handle marked candidates in Find file other window

xuchunyang:
Fixes (#1212)

I wrote this base on helm buffer's `C-c o` (that is, `helm-switch-to-buffers`).

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -465,7 +465,7 @@ Don't set it directly, use instead `helm-ff-auto-update-initial-value'.")
    "Symlink files(s) `M-S, C-u to follow'" 'helm-find-files-symlink
    "Relsymlink file(s) `C-u to follow'" 'helm-find-files-relsymlink
    "Hardlink file(s) `M-H, C-u to follow'" 'helm-find-files-hardlink
-   "Find file other window `C-c o'" 'find-file-other-window
+   "Find file other window `C-c o'" 'helm-find-files-other-window
    "Switch to history `M-p'" 'helm-find-files-switch-to-hist
    "Find file other frame `C-c C-o'" 'find-file-other-frame
    "Print File `C-c p, C-u to refresh'" 'helm-ff-print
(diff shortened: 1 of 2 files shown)


thierryvolpiatto:
Looks good except the error handling for number max of windows to split.

xuchunyang:
Yes, I just realized that, actually, I only tested 1/2/3 candidates before. helm buffer's `C-c o` has the same limitation.

thierryvolpiatto:

> Yes, I just realized that, actually, I only tested 1/2/3 candidates
> before. helm buffer's C-c o has the same limitation.

See also `dired-simultaneous-find-file' which we are using in
`helm-find-file-or-marked'.

The point is if one marks a lot of files and then decides to do C-c o, at
best it will hang emacs, otherwise emacs will crash.

-- 
Thierry

xuchunyang:
> The point is if one marks a lot of files and then decides to do C-c o, at
> best it will hang emacs, otherwise emacs will crash.

I don't understand your meaning, how that can happen?

I don't notice any serious problem, I've just tried to open all 49 files in helm git repo with `C-c o`, emacs does open all files but just show me 4 windows of them (all files are opened successfully), it interrupts with the error from `split-window` call

```
split-window: Window #<window 157 on helm-apt.el> too small for splitting (2)
```

I think the error is clear to user and we don't have to handle this, just like typing too many `C-x 3` at a time. And `dired-simultaneous-find-file` works in a different way: it can usually show much more windows (10+ in my case) and once it finds there are too many files to open, it will just give it up without open any file so that `split-window` will not fail.

thierryvolpiatto:

>     The point is if one marks a lot of files and then decides to do C-c o, at
>     best it will hang emacs, otherwise emacs will crash.
>    
> I don't understand your meaning, how that can happen?

I guess I had problems in the past when I wrote helm-switch-to-buffers.

> I don't notice any serious problem, I've just tried to open all 49
> files in helm git repo with C-c o, emacs does open all files but just
> show me 4 windows of them (all files are opened successfully), it
> interrupts with the error from split-window call
>
> split-window: Window #<window 157 on helm-apt.el> too small for splitting (2)

Ok so let use that, we will see later if we need to handle errors
ourself.

-- 
Thierry

thierryvolpiatto:
Thanks, merged.

PR #1216, opened 2015-10-13
Two very trivial fixes

xuchunyang:
Feel free to correct me if I am wrong.

### DIFF
--- a/helm-command.el
+++ b/helm-command.el
@@ -62,6 +62,9 @@ Show all candidates on startup when 0 (default)."
   :group 'helm-command-faces)
 
 
+(defvar helm-M-x-input-history nil)
+
+
 (cl-defun helm-M-x-get-major-mode-command-alist (mode-map)
   "Return alist of MODE-MAP."
   (when mode-map
@@ -217,6 +220,7 @@ than the default which is OBARRAY."
                 :persistent-help "Describe this command"
                 :history (or history extended-command-history)
                 :reverse-history helm-M-x-reverse-history
+                :input-history 'helm-M-x-input-history
                 :del-input nil
                 :help-message 'helm-M-x-help-message
                 :must-match t
--- a/helm-command.el
+++ b/helm-command.el
@@ -62,9 +62,6 @@ Show all candidates on startup when 0 (default)."
   :group 'helm-command-faces)
 
 
-(defvar helm-M-x-input-history nil)
-
-
 (cl-defun helm-M-x-get-major-mode-command-alist (mode-map)
   "Return alist of MODE-MAP."
   (when mode-map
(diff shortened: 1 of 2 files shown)


michael-heerdegen:
FWIW, I would prefer to switch to `helm-completing-read-symbols' internally whenever the second argument to `completing-read' is `eq' to (the value of) `obarray'.  Thierry didn't like the idea, though.  It's not perfect, but I  think it would still be better than a hardcoded whitelist of command names.  We could also combine the two methods.

thierryvolpiatto:

> Feel free to correct me if I am wrong.

>   * Remove unused variable helm-M-x-input-history

It is unused indeed, but if it is here, it is I used it in the past and
just forgot to reenable it, can't remember.
No body complained about this, but perhaps it is a good thing to
reenable ?
It is used when you use M-p, actually we have minibuffer-history in M-x
which send unrelated stuff for M-x.
If you want to reenable it in this PR, add it to :input-history.

>   * Use helm-completing-read-symbols for disassemble

Ok for this one, but perhaps we should follow Michael advice for this
and use it inconditionally when we find an obarray.
But lets go like this for now ;-).

Thanks.

-- 
Thierry

thierryvolpiatto:

> FWIW, I would prefer to switch to helm-completing-read-symbols'
> internally whenever the second argument tocompleting-read' is eq' to
> (the value of)obarray'. Thierry didn't like the idea, though. It's not
> perfect, but I think it would still be better than a hardcoded
> whitelist of command names. We could also combine the two methods.

I am not against, I am just reluctant to such changes when not really
necessary, but yes perhaps we should do this at some point.

-- 
Thierry

thierryvolpiatto:

> Chunyang Xu <notifications@github.com> writes:
>
>> Feel free to correct me if I am wrong.
>
>>   * Remove unused variable helm-M-x-input-history
>
> It is unused indeed, but if it is here, it is I used it in the past and
> just forgot to reenable it, can't remember.

I looked in git history and couldn't find it.
I think I added the variable with the intention of using it in the
helm-comp-read of helm-M-x and always forgot to bind it argh!

-- 
Thierry

michael-heerdegen:
I don't like my approach much, too, but I think it's the best we can currently do.  Of course could we handle the case more generally, e.g. via a `obarrayp' test, and handle arbitrary obarrays.

xuchunyang:
> It is unused indeed, but if it is here, it is I used it in the past and
> just forgot to reenable it, can't remember.
> No body complained about this, but perhaps it is a good thing to
> reenable ?
> It is used when you use M-p, actually we have minibuffer-history in M-x
> which send unrelated stuff for M-x.
> If you want to reenable it in this PR, add it to :input-history.

Done in a7eee27

thierryvolpiatto:
Thanks, merged.

PR #1218, opened 2015-10-13
Make file path style and truncate lines in helm grep configurable.

etanzapinsky:
Add two configuration options:
1. Do or do not truncate the grep lines (and have them wrap or not wrap
according to your config).
2. Configure filename in grep results style: just filename, absolute
path, relative path.

All these configurations default to the current behavior.


thierryvolpiatto:
Merged, with some minor changes, thanks.

etanzapinsky:
My pleasure. And always good to learn better Elisp :smile:

PR #1220, opened 2015-10-16
Fix two docstrings

xuchunyang:
The user option `helm-yank-symbol-first' has been removed. The [related part](https://github.com/emacs-helm/helm/wiki#yanking-text) of helm wiki also needs to be updated.


thierryvolpiatto:
Thanks, merged.

Issue #1222, opened 2015-10-17
helm-execute-persistent-action in describe-function, describe-variable, and helm-apropos


issue-dispenser:
I really like using helm-execute-persistent-action (C-j by default) in helm-M-x where it opens a temporary window with the whole function documentation. Then a simple C-j closes that window.

Would this feature be possible with helm-mode-describe-function and helm-mode-describe-variable? Currently doing helm-execute-persistent-action in either of these opens up a brief docstring in the mode-line, which is not nearly as useful as the above helm-M-x functionality.

Additionally, in helm-apropos, doing helm-execute-persistent-action after another helm-execute-persistent-action doesn't close the temporary window like helm-M-x does. I believe this is unintended.

xuchunyang:
> Would this feature be possible with helm-mode-describe-function and helm-mode-describe-variable?

Yes, you can customize the user option `helm-elisp-help-function` (default value is `'helm-elisp-show-doc-modeline`):

    (setq helm-elisp-help-function #'helm-elisp-show-help)

> Additionally, in helm-apropos, doing helm-execute-persistent-action after another helm-execute-persistent-action doesn't close the temporary window like helm-M-x does. I believe this is unintended.

I also found this.

issue-dispenser:
>Yes, you can customize the user option helm-elisp-help-function (default value is 'helm-elisp-show-doc->modeline):

Ah, that's nice, thanks. However, just changing the user option results in the same issue that helm-apropos has, where doing another helm-execute-persistent-action does not close the temporary window.

Also, changing the value does not change the helm header's string (that is, it still shows `C-j: Show brief doc in mode-line (keeping session)` when it should change to `Describe this function...` or `Describe this variable...` similar to what helm-M-x shows.

xuchunyang:
> Ah, that's nice, thanks. However, just changing the user option results in the same issue that helm-apropos has, where doing another helm-execute-persistent-action does not close the temporary window.

Yes, you are right. I have a quick and dirty fix for `helm-mode-describe-*`.

```elisp
(setq helm-elisp-help-function #'my-helm-elisp-show-help)

(setq in-help nil
      help-cand nil)

(defun my-helm-elisp-show-help (candidate)
  "Show full help for the function."
  (let ((hbuf (get-buffer (help-buffer))))
    (if (and in-help (string= candidate help-cand))
        (progn
          ;; When M-x is started from a help buffer,
          ;; Don't kill it as it is helm-current-buffer.
[...]

PR #1224, opened 2015-10-17
Use same persistent fn for M-x and apropos (#1222).

thierryvolpiatto:
* helm-elisp.el (helm-elisp--persistent-help): New.
(helm-def-source--emacs-variables):  Use it.
(helm-def-source--emacs-commands):  Use it.
(helm-def-source--emacs-functions):  Use it.
(helm-def-source--eieio-classes):  Use it.
(helm-def-source--eieio-generic):  Use it.
* helm-command.el (helm-M-x-read-extended-command):  Use it.

### DIFF
--- a/helm-command.el
+++ b/helm-command.el
@@ -169,29 +169,11 @@ than the default which is OBARRAY."
                  (read-extended-command))
             (helm-mode 1))
           (read-extended-command))
-      (let* (in-help
-             help-cand
-             (orig-fuzzy-sort-fn helm-fuzzy-sort-fn)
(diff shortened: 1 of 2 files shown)


xuchunyang:
I've just tried and it works nicely. While current persistent-help displays "Anonymous", so users can't learn what persistent-action is from there.

thierryvolpiatto:

> I've just tried and it works nicely.

Thanks to check.

> While current persistent-help displays "Anonymous", so users can't
> learn what persistent-action is from there.

Yes, I see, I didn't finish this.

-- 
Thierry

PR #1227, opened 2015-10-17
Customize updatedb function

igorepst:
As GNU `updatedb` doesn't have an option of `prunenames`, users should use `prunepaths` only to exclude some paths from being scanned into the DB. The proposed PR brings `helm-locate-create-db-function` and adds an option to redefine it, by default using `format` like it was until now. Hence the user has an option to write something like the following:

```
(defun ig-helm-locate-create-db-function (db-name directory)
  (format helm-locate-create-db-command db-name directory
	  (expand-file-name ".git" directory)))

(setq helm-locate-create-db-command
		"updatedb --output=%s --localpaths='%s' −−prunepaths='%s'")
(setq helm-locate-create-db-function
	      #'ig-helm-locate-create-db-function)
```

### DIFF
--- a/helm-locate.el
+++ b/helm-locate.el
@@ -155,6 +155,19 @@ fall back to `default-directory' if FROM-FF is nil."
                  helm-ff-locate-db-filename
                  default-directory))))))
 
+
+(defun helm-locate-create-db-default-function (db-name directory)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks, merged.

PR #1229, opened 2015-10-18
Clear echo area after calling describe-* (#1222)

xuchunyang:


thierryvolpiatto:
Sorry , I have commited a similar fix, probably at the same time.
Perhaps you can close this.
Thanks.

Issue #1233, opened 2015-10-20
Improve persistent-help when a source's action slot is variable name


xuchunyang:
For example, `helm-locate-library`'s persistent-help is "C-j: helm-type-file-actions (keeping session)", and I can't learn what persistent action is from this message. Some other commands such as `helm-filtered-bookmarks` has the similar problem.

---
code snippet to explain the title:
```elisp
(defvar foo-actions '(("hello" . message)))
(helm-build-sync-source "foo"
  :candidates '("one")
  :action 'foo-actions)

=> (header-line . "C-j: foo-actions (keeping session)")

(helm-build-sync-source "foo"
  :candidates '("one")
  :action '(("hello" . message)))

=> (header-line . "C-j: hello (keeping session)")
```

thierryvolpiatto:
Chunyang Xu <notifications@github.com> writes:

> For example, helm-locate-library's persistent-help is "C-j:
> helm-type-file-actions (keeping session)", and I can't learn what
> persistent action is from this message.  Some other commands such as
> helm-filtered-bookmarks has the similar problem.

Well, it is not that the information is incorrect, it is just sending
the symbol-name of action when nothing better is available, and if
nothing else is available, it is hard to deduce what the action does
better.

To have a useful info, one should add either a :persistent-help in its
source or a (display . action) entry (like you describe in your
examples).
I will add :persistent-help in sources where it is missing as soon as
possible (unless you beat me with a PR ;-))

-- 
Thierry
https://emacs-helm.github.io/helm/

PR #1234, opened 2015-10-20
Set persistent-help to caar of action alist if applicable

xuchunyang:
Re #1233


xuchunyang:
Furthermore, can we set persistent-help to "Do nothing" if action is `identity`?

thierryvolpiatto:
> Furthermore, can we set persistent-help to "Do nothing" if action is identity?

Yes.

thierryvolpiatto:
Thanks.

Issue #1237, opened 2015-10-21
Providing better candidates to Score functions in fuzzy matching stage


ghost:
## Problem

With fuzzy-matching turned on (helm-flx or not), querying for `elm` in `helm-M-x` does not bring up `emacs-lisp-mode` anywhere in the first 100 (actually, 800) results.

## Background
@thierryvolpiatto , you can skip this :)

I and it seems quite a few other people consider https://github.com/emacs-helm/helm/issues/145#issuecomment-148177275 and [helm-flx](https://github.com/PythonNut/helm-flx) to be a big deal.

`helm-flx` uses helm's built in fuzzy matching to filter candidates into a list of matches,
which it passes on to lewang's `flx` package for scoring. This works well.

One limitation of this is that flx can't boost search results it doesn't get to see, so if `helm-candidate-number-limit` is set to a "normal" value (default 100), helm's fuzzy-match algo quickly finds the first 100 candidates that match and stops looking at the remaining 10000, where the best candidate is waiting. The uninformed blame this on "poor fuzzy matching" but it's really just an optimization/safety that backfires in this case.

Right now, the only way around this is to crank up `helm-candidate-number-limit` to unreasonable values which can result in  nasty side-effects like really bad GC pauses.

Note: Why is matching and scoring kept seperate? Presuambly scoring potentially all candidates 
using an elisp function would be too expensive.

## Proposed Solution

To ensure that the "best" candidates are included in the matches we run a cheap (regex-based),  specialized function over **all** candidates to get any exceptionally good candidates in the matches list. Following that, we make another pass using the existing fuzzy-match regex using the fuzzy matching code to fill the quota with the more "generic" fuzzy matching. Together they provide a much better result overall.

For now, I've defined "exceptionally good" as a candidate whose initials cover the entire query:
`elm` matches `emacs-lisp-mode`
`fbpf` matches `foo/bmkp-previous-non-file-bookmark`

Helm has a few optimizations that I had to work around to implement this, but the results are very good and if @thierryvolpiatto agrees this is a good thing to have, helm can be changed to accomodate.

Here's a proof-of-concept:
```elisp
(require 'cl-lib) ; for lexical-let

(defun helm--mapconcat-initials-pattern (pattern seperators)
  "Transform string PATTERN into a regexp for fuzzy matching as initials.
e.g abcd$
    => \"\\(^\\|[^a]*[-/]\\)a[^b]*[-/]b[^c]*[-/]c$\".

[...]

PR #1238, opened 2015-10-21
WIP: better fuzzy matches

ghost:
Rough Demo for better results as described https://github.com/emacs-helm/helm/issues/1237

To see the difference, try `helm-M-x` with the default `helm-candidate-number-limit` of 100
and see what "elm" brings up compared to unpatched version.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -869,6 +869,11 @@ It also accepts function or variable symbol.")
   "Variable `buffer-file-name' when `helm' is invoked.")
 (defvar helm-candidate-cache (make-hash-table :test 'equal)
   "Holds the available candidate within a single helm invocation.")
+(defvar helm-preferred-matches-cache (make-hash-table :test 'equal :size 8192)
+  "Caches the preferred matches within a single helm invocation.")
+(defvar helm-preferred-candidates-cache (make-hash-table :test 'equal :size 1024)
+  "Caches the complete candidates list for locating preferred matches.")
+(defvar helm-preferred-matches-candidate-copy '())
 (defvar helm-input ""
   "The input typed in the candidates panel.")
 (defvar helm-input-local nil
@@ -3114,12 +3119,35 @@ and `helm-pattern'."
              (if (string-match "[[:upper:]]" pattern) nil t)))
     (t helm-case-fold-search)))
 
-(defun helm-match-from-candidates (cands matchfns match-part-fn limit source)
-  (let (matches)
+
+(defun helm--mapconcat-initials-pattern (pattern seperators)
+  "Transform string PATTERN into a regexp for fuzzy matching as initials.
+e.g (helm--mapconcat-initials-pattern \"abcd\" \"-/\")
+    => \"^a\\(.+?[- /]b\\)\\(.+?[- /]c\\)\\(.+?[- /]d\\)\"
+
+SEPERATORS is a string contains one or more delimiters denotice word boundaries.
+For 'foo/my-function' SEPERATORS would be \"/-\""
+  (let ((ls (split-string-and-unquote pattern "")))
+    (concat
+     (format "^%s" (car ls) seperators (car ls))
+     (mapconcat (lambda (c)
+                  (if (and (string= c "$")
+                           (string-match "$\\'" pattern))
+                      c (format "\\(.+?[%s]%s\\)" seperators c)))
+                (cdr ls) ""))))
+
+(defun helm--make-initials-matcher (pattern &optional seperators)
+  (let* ((initials-pat (helm--mapconcat-initials-pattern pattern (or seperators "- /")))
+         (matcher (lambda (candidate)
+                    ;; (message "initials matcher Ran")
+                    (string-match initials-pat candidate))))
+    matcher))
+
+(defun helm-match-from-candidates-1 (cands matchfns match-part-fn limit source)
+  (let* (matches)
     (condition-case-unless-debug err
         (let ((item-count 0)
               (case-fold-search (helm-set-case-fold-search)))
-          (clrhash helm-match-hash)
           (cl-dolist (match matchfns)
             (when (< item-count limit)
               (let (newmatches)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks for the PR, it is easy to see what have been done, comment etc...

IIUC the point of this PR is to take in account the separators in matching regexp.
This is interesting, we miss this indeed, but I guess it could be implemented differently and more efficiently (will see what I can do as soon as I have a journey to work on this, I have too few time actually for complex stuff).
Also in your PR in-buffer method (i.e `helm-source-in-buffer`) is not implemented.

AlKapote:
Thanks for reviewing, clearly I should have given you a branch to begin with so review is easier.. I'll remember that.

As for "more efficiently" see my response above, I think you missed something crucial about
how this works. But, one thing I've tried which works very well is to force the regex to have
the first letter of the query match the first letter of the candidate. In this way the regex FSM
can exit very quickly when examining candidates and this becomes quite quick to do. I think
the performance is reasonable even without it, but for longer queries (4,5) the extra regex
pass does become a little laggy without that change.


I'll look into `helm-source-in-buffer` because I have no idea what that is.

AlKapote:
> IIUC the point of this PR is to take in account the separators in matching regexp.
That's only half of it. The other half as that you *must* look through all the candidates 
to locate some promising candidtes first. It's not enough to find a 100 "matches" and stop,
becuase those matches are very likely to score very low.

As an example, the top resul helm currently returns for `elm` (from #1237) is:
`helm-insert-latex-math`. 

That's a *terrible* fuzzy match.

AlKapote:
@thierryvolpiatto  Here's the better support for the `source-in-buffer` support you repeatedly
mentioned. I'll note again that this also worked fine before this change, but gave even better
results if `candidates-in-buffer` was disabled. Now, there's not need and I've reset `candidates-in-buffer` to `t` again.

To enable us to find really "good" candidates before the general pass, we pin a new slot on the source called `all-candidates` which holds the entire candidate list and use that for the first pass only. This should be faster and should apply to all types of sources, although it may mean
making small changes in several places to support it. Right now, anything that uses `helm-comp-read` should work.

This is still not meant to be merged, it's just an example to see what you think about doing
things this way.

For anyone following along, this is currently only enabled for the "Emacs Commands" source.

AlKapote:
Even better matches now with a tweaked regex, works in `describe{-variables,commands}`
too now. Still just experimenting.

AlKapote:
Better fuzzy matches implemented for all describe-* functions (in-source-buffer),
`hcnl` now brings up `helm-candidate-number-limit`, in case we get to tests later.

thierryvolpiatto:
Alfred Kapote <notifications@github.com> writes:

> @thierryvolpiatto Here's the better support for the source-in-buffer support you repeatedly
> mentioned. I'll note again that this also worked fine before this change, but gave even better
> results if candidates-in-buffer was disabled. Now, there's not need and I've reset candidates-in-buffer to t again.
>
> To enable us to find really "good" candidates before the general pass, we pin a new slot on the source called all-candidates which holds the entire candidate list and
> use that for the first pass only. This should be faster and should apply to all types of sources, although it may mean
> making small changes in several places to support it. Right now, anything that uses helm-comp-read should work.
>
> This is still not meant to be merged, it's just an example to see what you think about doing
> things this way.

I don't think doing two passes is a good thing, it will alter
significantly performances, thus in your second pass the regexp you use
is not cached, isn't it?
When the regexp is not cached things become really slow.

Keep in mind that for fuzzy matching, the difficult part is finding the
good balance between "being fast/being smart".

-- 
Thierry

thierryvolpiatto:
Alfred Kapote <notifications@github.com> writes:

>     IIUC the point of this PR is to take in account the separators in matching regexp.
>     That's only half of it. The other half as that you must look through all the candidates
>     to locate some promising candidtes first. It's not enough to find a 100 "matches" and stop,
>     becuase those matches are very likely to score very low.
>    
> As an example, the top resul helm currently returns for elm (from #1237) is:
> helm-insert-latex-math.
>
> That's a terrible fuzzy match.

Depend, maybe yes, maybe no, if you focus on elm=emacs-lisp-mode, yes it
is a bad match, buf people may try to find something else too.
If you write something more precise you match immdiately, so what we
have currently is not so bad as you would say.

e.g "emlsp" => emacs-lisp-mode.

-- 
Thierry

AlKapote:
> I don't think doing two passes is a good thing, it will alter significantly performances,

Yes, but slower isn't itself a problem only "too slow" and it should be an opt-in thing like fuzzy matchng anyway. My cpu isn't very fast and for me this is acceptable even in apropos with 10k+ entries.

> thus in your second pass the regexp you use is not cached, isn't it?

I don't see why. Constructing the regex happens once after a query is updated, when the closure is constructed. Do you mean something else?

> Keep in mind that for fuzzy matching, the difficult part is finding the good balance between "being 
> fast/being smart".

Absolutely, otherwise you could simply use the score function over all candidates. But, I would add
that the idea of "fuzzy" matching is to dwim, behave as the user expects. querying by initials
is a very common expectation, I think.

>>  `elm` matches `helm-insert-latex-math` That's a terrible fuzzy match.

> maybe yes, maybe no, people may try to find something else too.
> If you write something more precise you match immdiately "emlsp" => emacs-lisp-mode.

Sure It works for some queries but also often enough it doesn't do what I expect. The problem is
specifically that for initials queries the results almost never bring up the best candidates even though
they express a very specific class of matches. Part of the idea is to get there with a minimum of keystrokes and a minimum of suprise requiring trial and error to find a working query. 
That's why initials queries deserve to be better supported I think. I agree that for longer queries the current code works very well.

I'd like helm to guess what I'm thinking more of the time and I think this demo shows this could
be done with acceptable (I think) performance. For me, querying by initials is a major use case.

Also, I would love to do the same kind of "preferred-matching" for "EMacs-Lisp-Mode" but I think the regex for all combinations will be too slow.

AlKapote:
@thierryvolpiatto, not sure if this is what you meant with "regex caching" but if it was, it's a good idea.

Once a scan of all candidates for `QUERY` returns less then LIMIT matches we can be certain that any match for `QUERY.` we might find in a full scan is already present in the candidates we found for `QUERY`. By caching results we can avoid doing an extra full scan quite often.

This doesn't make a huge difference (it seems fast enough already) but it may in some situations.

**Please consider** cherry-picking:
- e3d54f1 unrelated optimization, hash table resizing is usually an expensive operation, a better initial guess on size can improve performance.
- 5099dc4 whitespace fixes, would save me from having to omit these each time I hit save and
stage for a commit.

AlKapote:
Cleaned up much smaller patch now and there's no need to touch any of the existing sources
now, if `:fuzzy-match` is set for the source it just works.

helm.el
```diff
@@ -3114,12 +3119,35 @@ and `helm-pattern'."
              (if (string-match "[[:upper:]]" pattern) nil t)))
     (t helm-case-fold-search)))
 
-(defun helm-match-from-candidates (cands matchfns match-part-fn limit source)
-  (let (matches)
+
+(defun helm--mapconcat-initials-pattern (pattern seperators)
```
lewang: Can you explain why this pattern is so restrictive?

If this is a preliminary filtering step, then inserting ".*" in between characters should be enough.  Is there benchmarking to prove this is necessary?

The more candidates flx sees the better your results will be.

AlKapote:
Added a new knob to trade off speed for accuracy.

`helm-preferred-max-group-length` now lets you control how hard to try variations.
with a value of 1 will match `a-b-c-d`
with a value of 2 will also match `ab-c-d` `a-bc-d`, etc'
etc'

 I don't notice a slowdown even with this on and it seems to work *fantastically* well.
Even with a very low `(setq helm-candidate-number-limit 10)` all the following queries bring up `helm-candidate-number-limit` as the first match in `describe-variable`:
```
hcnl
hec
hcn
henl
...
```

helm.el
AlKapote: What you're suggesting I've already covered in the original issue #1237, please read it.
Also, Inserting greedy matching operators can and will drop the best match from the list.

> The more candidates flx sees the better your results will be.

Right, but that I've had to set the limit at 2500 at one point to get the match I wanted,
it was acceptable but not very quick, It caused long GCs  (yes, I know about the
GC tuning you recommend) and most of that work is completely unnecessary. 
What you're suggesting

lewang: So this is designed to not need flx?  That's really interesting.  I will follow along and see how it goes.

AlKapote: No, it works best when used together with `helm-flx`. This keeps the number of matches sent to `flx` small-ish, but tries harder to have the best candidates available among those sent to `flx`.

You could use the latest iteration of the regex to replace helm's stock one for fuzzy matching,
and even without flx it should work quite well, but I haven't tried it.

AlKapote:
Now that the full candidate list is cached on first invocation instead of living in a slot on the source,
there's a mysterious bug when a query is typed quickly instead of one character at a time. I'll fix this later.

MikeSanMichel2:
@thierryvolpiatto, I've cleaned this up, added support for `source-in-buffer` and much caching as I could. Performance is good even in apropos, the fuzzy matches are *vastly* better IMO and afaik nothing is broken by it (I've tested find-files, describe-,M-x, apropos).

What else should I test?

thierryvolpiatto:
MikeSanMichel2 <notifications@github.com> writes:

> @thierryvolpiatto, I've cleaned this up, added support for
> source-in-buffer and much caching as I could. Performance is good even
> in apropos, the fuzzy matches are vastly better IMO and afaik nothing
> is broken by it (I've tested find-files, describe-,M-x, apropos).
>
> What else should I test?

I haven't had the time to look at all your code carefully, but from the
few I see, it still have errors and it changes way too much things that
I guess will create tons of bugreport, I am not sure also the
performance will not be altered.

As I said I think there is a better approach to fix this issue -- better
fuzzy matching and scoring on prefixes after a separator --.

So thanks for your work, we can keep it around and work on it
eventually, but it is not ready to merge in master, it would break helm
badly IMO.
Thus your account have disapeared now and your branch is gone...

-- 
Thierry

thierryvolpiatto:
I am closing this because I see you have now a ghost account, feel free to submit again.

PR #1255, opened 2015-10-31
Make helm-kill-marked-buffers count killed buffers

issue-dispenser:
Before the function didn't print the correct number of buffers actually killed (if you changed your mind at the kill prompt when the buffers were modified). Now it prints out how many buffers were actually killed.


thierryvolpiatto:
Thanks.

Issue #1253, opened 2015-10-31
helm-execute-persistent-action in various helm commands (find/locate/mini/find-files)


issue-dispenser:
* In `helm-find` and `helm-locate`, I believe doing two consecutive `helm-execute-persistent-action`'s should kill the buffer that was being previewed (similar to what happens when one gives a prefix argument to `helm-execute-persistent-action` in `helm-find-files`). It currently does nothing.

* It would be very nice if doing two consecutive `helm-execute-persistent-action`'s in `helm-mini` would close (but not kill) the buffer that is being previewed. Currently it does nothing.

* Giving a prefix argument to `helm-execute-persistent-action` in `helm-mini` is not only undocumented in `helm-mini`'s interface, but also it saves the buffer being killed without asking. I believe that it should display an additional prompt asking if the buffer should be saved or not (or just act as if you killed it with `M-D` and don't write at all, which seems more consistent).

* Giving a prefix argument to `helm-execute-persistent-action` in `helm-find-files` does not kill/close the previewed buffer if said buffer was followed from a symlink. I believe it should, if possible.

thierryvolpiatto:
I think all issues have been fixed here, except the last (symlink) but I wont fix it, it is ok for me as it is actually.

issue-dispenser:
helm-buffers-list/mini still doesn't mention the prefix argument but other than that and the symlink deal it's been fixed.

thierryvolpiatto:
issue-dispenser <notifications@github.com> writes:

> helm-buffers-list/mini still doesn't mention the prefix argument

You mean the prefix arg to kill buffer ? (i.e C-u C-j)

Perhaps it's time to drop it (no more prefix arg for this) now we have
C-c d.

-- 
Thierry
https://emacs-helm.github.io/helm/

PR #1256, opened 2015-11-01
Change helm-buffers-list PA order (#1253 & 81742ca)

issue-dispenser:
Makes it so that it goes to the helm-current-buffer only when the candidate is the same as the current PA display window.
This is more consistent with the PA in some other helm commands.
Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,7 @@ If you don't want the Helm window to be resized, but a smaller Helm window, you 
 **Warning** You will find many extensions outside of [emacs-helm](https://github.com/emacs-helm)
 that provide nothing more than what is provided natively by [helm](https://github.com/emacs-helm/helm) and may not be in sync will `helm` core.
 So generally prefer what is provided natively in `helm` instead of its counterpart provided externally.
+Actually more than 20 unuseful or deprecated packages you can find on MELPA, be aware.
 
 # Known issues
 


thierryvolpiatto:
This will not work, I have fixed this behavior differently, please update.

issue-dispenser:
How would this not work?

It seems the difference between our versions is whether or not performing another action affects the function call (for example, doing a PA to preview, then C-n, then C-p, then doing another PA on the same buffer would not switch the buffer in your version).

The PA in some other commands (find-files, describe-*) do not care about another action, while your fix for buffers-list does. I don't believe it should, but I don't feel too strongly about it either way. Either way, the functionality is better than before.

thierryvolpiatto:
issue-dispenser <notifications@github.com> writes:

> How would this not work?

Well, look carefully at your code again, it can't work like this, I have comitted
another fix though.

Thanks for helping at this.

-- 
Thierry

issue-dispenser:
Hmm, I still don't see the error...
 It doesn't matter now, but I'm curious since you mentioned it.

The display window (call it PA) can be one of 3 things, right? Either helm-current-buffer, the candidate buffer, or some other buffer. It could also be both helm-current-buffer and the candidate (in which case, it doesn't matter which buffer it switches to since they're the same).

So, I believe your version does this:
```
If PA == helm-current-buffer      => switch to candidate
If PA == candidate                => switch to helm-current-buffer
If PA == other                    => switch to candidate
```
Mine did:
```
If PA == candidate                => switch to helm-current-buffer
else                              => switch to candidate
```
Though the new fix does work exactly as I expected. Thank you for that.

PR #1258, opened 2015-11-06
Make sure the org refile command works even in the agenda buffer

Konubinix:


thierryvolpiatto:
Thanks, merged.

alphapapa:
This is great, thank you!  I have noticed this behavior for a while now but wasn't sure what to do about it.  I didn't even realize it was an issue with Helm.  Great work.  :)

PR #1261, opened 2015-11-07
Update helm-ff-kill-buffer-fname

issue-dispenser:
* Do not kill helm-current-buffer
* Report the buffer-name when killing buffer


thierryvolpiatto:
Thanks, I will probably modify to have a message like in buffers (someting like "can't kill helm-current-buffer").

PR #1263, opened 2015-11-08
Make "file name at point" the default input for helm-locate

YorkZ:
Hi Thierry,

I've made "file name at point" the default input for helm-locate so that one can easily press "M-n" to get `locate` to find the file name at point. In my opinion, this is more convenient. What do you think?
Thanks,

York

### DIFF
--- a/helm-locate.el
+++ b/helm-locate.el
@@ -389,7 +389,7 @@ Where db_path is a filename matched by
 `helm-locate-db-file-regexp'."
   (interactive "P")
   (setq helm-ff-default-directory default-directory)
-  (helm-locate-1 arg))
+  (helm-locate-1 arg nil nil (thing-at-point 'filename)))
 
 (provide 'helm-locate)
 


thierryvolpiatto:
York Zhao <notifications@github.com> writes:

> Hi Thierry,
>
> I've made "file name at point" the default input for helm-locate so
> that one can easily press "M-n" to get locate to find the file name at
> point. In my opinion, this is more convenient. What do you think?

I don't think it is really useful, it does more or less what we actually
have.
The problem is that if the file displayed at point is absolute and not
located exactly at this place you will give this full path to locate and
it will not find it (same if there is a tilde in the path, etc...).

What I propose is the same as you propose but using instead the basename
of the file at point, if you find this useful, correct your PR to do
this and I will merge (i.e (helm-basename (thing-at-point 'filename))).

Thanks.

-- 
Thierry

YorkZ:
> I don't think it is really useful

I guess the reason why you don't think it is useful is because we are using different option for `locate`. I use "-A" to match the full path but helm doesn't seem to use "-A by default.

I've actually been in a situation of spending a lot of time analyzing log files where being able to "M-n" the file name at point has been extremely useful. For example, when the log file says "Loaded init/script.xml", I could simply type "M-n" and have locate finding precisely the file for me. If in this situation I was locating only the base name "script", I could end up having to pick the one in "init/" directory a lot of files with the name "script" in many different directories.

> The problem is that if the file displayed at point is absolute and not located
> exactly at this place you will give this full path to locate and it will not
> find it (same if there is a tilde in the path, etc...).

Again, this maybe because you don't use the "-A" option. If this is the case, I can see your point. Therefore, do you think it might be better if we make this customizable so that people who want to pop out file name at point when pressing "M-n" could customize the behaviour in their .emacs file?

Thanks,

York

thierryvolpiatto:
This will not change much actual behavior (thing-at-point 'filename is not much better than what we use by default) but if you insist I merge this.

Thanks.

YorkZ:
Thanks for merging my changes even though we are not completely agree on it's usefulness. I really appreciate it!

I think popping out the file name at point when pressing "M-n" is still easier. In my example, previously I would have to first move the point to the beginning of the file name, i.e., move to the beginning of "init/script.xml", then type "M-n" followed by three to fours times of "C-w". It will definitely make a difference when you have to this over and over.

thierryvolpiatto:
York Zhao <notifications@github.com> writes:

> Thanks for merging my changes even though we are not completely agree on it's usefulness. I really appreciate it!
>
> I think popping out the file name at point when pressing "M-n" is
> still easier. In my example, previously I would have to first move the
> point to the beginning of the file name, i.e., move to the beginning
> of "init/script.xml", then type "M-n" followed by three to fours times
> of "C-w". It will definitely make a difference when you have to this
> over and over.

You can avoid hitting M-n by adding helm-source-locate to
helm-sources-using-default-as-input.

-- 
Thierry

YorkZ:
> You can avoid hitting M-n by adding helm-source-locate to
> helm-sources-using-default-as-input

This is fantastic, great to know. Thanks a lot!

PR #1268, opened 2015-11-19
helm-org.el: Add helm-org-parent-headings

alphapapa:
I added a new Helm function that lists Org headings that are parents of the current heading.  This is a nice alternative to doing `C-c C-u` repeatedly to go up the Org tree.

I'm very much a novice when it comes to modifying Helm, so I may not have done this in the way you would prefer.  At first, I added a new source and associated functions that basically duplicated existing code.  Then I decided it would be better to, instead, modify the existing source and `helm-org-get-candidates` to take a `parents` argument; this way, if `helm-org-get-candidates` or `helm-source-org-headings-for-files` change in the future, the `helm-org-parent-headings` function will benefit from the improvements as well. And, of course, there's no duplicated code this way.

Thanks for your work on Helm!  Couldn't imagine using Emacs without it!

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -58,9 +58,10 @@ NOTE: This will be slow on large org buffers."
   (org-show-entry))
 
 (cl-defun helm-source-org-headings-for-files (filenames
-                                              &optional (min-depth 1) (max-depth 8))
+                                              &optional (min-depth 1) (max-depth 8) (parents nil))
   (helm-build-sync-source "Org Headings"
-    :candidates (helm-org-get-candidates filenames min-depth max-depth)
+    :candidates (helm-org-get-candidates filenames min-depth max-depth parents)
+    :candidate-transformer (and parents 'reverse)
     :action '(("Go to line" . helm-org-goto-marker)
               ("Refile to this heading" . helm-org-heading-refile)
               ("Insert link to this heading"
@@ -87,26 +88,36 @@ NOTE: This will be slow on large org buffers."
       (org-end-of-subtree t t)
       (org-paste-subtree (+ target-level 1)))))
 
-(defun helm-org-get-candidates (filenames min-depth max-depth)
+(cl-defun helm-org-get-candidates (filenames min-depth max-depth &optional (parents nil))
   (apply #'append
-   (mapcar (lambda (filename)
-             (helm-get-org-candidates-in-file
-              filename min-depth max-depth
-              helm-org-headings-fontify
-              helm-org-headings--nofilename))
-           filenames)))
+         (mapcar (lambda (filename)
+                   (helm-get-org-candidates-in-file
+                    filename min-depth max-depth
+                    helm-org-headings-fontify
+                    (if parents t helm-org-headings--nofilename)
+                    parents))
+                 filenames)))
 
 (defun helm-get-org-candidates-in-file (filename min-depth max-depth
-                                        &optional fontify nofname)
+                                                 &optional fontify nofname parents)
   (with-current-buffer (pcase filename
                          ((pred bufferp) filename)
                          ((pred stringp) (find-file-noselect filename)))
     (and fontify (jit-lock-fontify-now))
-    (let ((match-fn (if fontify 'match-string 'match-string-no-properties)))
+    (let ((match-fn (if fontify 'match-string 'match-string-no-properties))
+          (goto-char-fn (if parents
+                            '(lambda () t)
+                          '(lambda () (goto-char (point-min)))))
+          (search-fn (if parents
+                         '(lambda () (and (org-up-heading-safe) (re-search-forward org-complex-heading-regexp nil t)))
+                       '(lambda () (re-search-forward org-complex-heading-regexp nil t))))
+          (get-outline-path-fn '(lambda (level heading parents) (if parents
+                                                                    (org-get-outline-path)
+                                                                  (org-get-outline-path t level heading)))))
       (save-excursion
-        (goto-char (point-min))
+        (funcall goto-char-fn)
         (cl-loop with width = (window-width)
-                 while (re-search-forward org-complex-heading-regexp nil t)
+                 while (funcall search-fn)
                  if (let ((num-stars (length (match-string-no-properties 1))))
                       (and (>= num-stars min-depth) (<= num-stars max-depth)))
                  collect `(,(let ((heading (funcall match-fn 4))
@@ -114,7 +125,7 @@ NOTE: This will be slow on large org buffers."
                                           (concat (helm-basename filename) ":")))
                                   (level (length (match-string-no-properties 1))))
                               (org-format-outline-path
-                               (append (org-get-outline-path t level heading)
+                               (append (funcall get-outline-path-fn level heading parents)
                                        (list heading)) width file))
                            . ,(point-marker)))))))
 
@@ -135,6 +146,18 @@ NOTE: This will be slow on large org buffers."
                     (list (current-buffer)))
           :candidate-number-limit 99999
           :buffer "*helm org inbuffer*")))
+
+;;;###autoload
+(defun helm-org-parent-headings ()
+  "Preconfigured helm for org headings that are parents of the
+current heading."
+  (interactive)
+  (helm :sources (helm-source-org-headings-for-files
+                  (list (current-buffer))
+                  1 50 t)  ; Use a large max-depth to ensure all parents are displayed
+        :preselect (org-format-outline-path (org-get-outline-path))
+        :candidate-number-limit 99999
+        :buffer "*helm org parent headings*"))
 
 ;;;###autoload
 (defun helm-org-capture-templates ()


helm-org.el
```diff
@@ -137,6 +147,16 @@ NOTE: This will be slow on large org buffers."
           :buffer "*helm org inbuffer*")))
 
 ;;;###autoload
+(defun helm-org-parent-headings ()
+  "Preconfigured helm for org headings that are parents of the current heading."
+  (interactive)
+  (helm :sources (helm-source-org-headings-for-files
+                  (list (current-buffer))
+                  1 8 t)
```
alphapapa: I hesitated to hard-code `1 8` here, but I don't know how else to do it, unless the `helm-source-org-headings-for-files` function were to have its arguments rearranged, which seems like a bad idea.  Of course, I am an elisp novice, so maybe there is a better way to do this.  :)

thierryvolpiatto: alphapapa <notifications@github.com> writes:

> In helm-org.el:
>
>> @@ -137,6 +147,16 @@ NOTE: This will be slow on large org buffers."
>>            :buffer "*helm org inbuffer*")))
>>  
>>  ;;;###autoload
>> +(defun helm-org-parent-headings ()
>> +  "Preconfigured helm for org headings that are parents of the current heading."
>> +  (interactive)
>> +  (helm :sources (helm-source-org-headings-for-files
>> +                  (list (current-buffer))
>> +                  1 8 t)
>
> I hesitated to hard-code 1 8 here, but I don't know how else to do it,
> unless the helm-source-org-headings-for-files function were to have
> its arguments rearranged, which seems like a bad idea. Of course, I am
> an elisp novice, so maybe there is a better way to do this. :)

Perhaps use (parent t) in helm-source-org-headings-for-files and then
you don't have to specify 1, 8 and parent in helm-org-parent-headings:

    (helm :sources (helm-source-org-headings-for-files
                    (list (current-buffer)))

-- 
Thierry

alphapapa: Sorry, I'm confused: if `(parent t)` were the default, then `helm-source-org-headings-for-files` would always show only parent headings, so all the other places that source is used would have to specify `(parents nil)` to avoid that.

However, I think it doesn't matter, because I was thinking about it last night, and I think the best thing to do is to specify a large `max-depth` anyway.  That way, if a user has more than 8 levels of headings (and I sometimes do), all of the parent headings will still be displayed.

I'll push another commit with the changes soon.  Thanks.

helm-org.el
```diff
@@ -58,9 +58,11 @@ NOTE: This will be slow on large org buffers."
   (org-show-entry))
 
 (cl-defun helm-source-org-headings-for-files (filenames
-                                              &optional (min-depth 1) (max-depth 8))
+                                              &optional (min-depth 1) (max-depth 8) (parents nil))
   (helm-build-sync-source "Org Headings"
-    :candidates (helm-org-get-candidates filenames min-depth max-depth)
+    :candidates (helm-org-get-candidates filenames min-depth max-depth parents)
+    :candidate-transformer (if parents 'reverse nil)
```
alphapapa: This seems like it should go in `helm-org-parent-headings`, because it would make the `if` unnecessary, but when I specified it there as a keyword attribute, it didn't have any effect.  Is there a better way to do this?

helm-org.el
```diff
@@ -58,9 +58,11 @@ NOTE: This will be slow on large org buffers."
...
+    :after-init-hook 'helm-follow-mode
```
alphapapa: I've tried several ways, but I can't find the right way to enable `helm-follow-mode`.  It needs to be enabled *after* the Helm buffer is created, because if it's already enabled when the buffer is created, it "follows" to the first candidate immediately, before the `:preselect` has a chance to select the current heading, which causes the point to jump to the top of the buffer and back down again.  It's not big deal, but it's a bit distracting, and there must be a way around it.  :)

alphapapa:
Here is a new commit with some improvements.  There are a couple of small issues that I'm not sure how to fix correctly.  Thanks for your help.

helm-org.el
thierryvolpiatto: Use the `:follow` slot with a 1 value.

helm-org.el
thierryvolpiatto: Try to avoid `(if foo 1 nil)` use instead `(and foo 1)`.

thierryvolpiatto: BTW I am not sure to understand what you want to do with your candidate-transformer.

thierryvolpiatto:
Thanks for your work, let me know when it will be ready to merge, my org usage and knowledge is quite limited.

helm-org.el
alphapapa: The candidate-transformer is to reverse the list of headings so that they appear in the proper order.  The headline-getting code works by walking up the tree from the current heading, which creates a list with the lowest heading at the top of the list, so it needs to be reversed before being displayed.

helm-org.el
alphapapa: That works, but it still has the problem I mentioned: the `follow` happens before the `preselect`, which causes the point to briefly jump to the top of the buffer before returning to the current heading, which is a bit distracting and...unpolished.    :)

The docstring for `after-init-hook` says that it runs after the Helm buffer is created, which sounds like it should allow activating of `helm-follow-mode` after the `preselect` runs, but I haven't been able to find a way to activate `helm-follow-mode` from the `after-init-hook`.  Is there a way to do this?

thierryvolpiatto: alphapapa <notifications@github.com> writes:

> In helm-org.el:
>
>>    (helm-build-sync-source "Org Headings"
>> -    :candidates (helm-org-get-candidates filenames min-depth max-depth)
>> +    :candidates (helm-org-get-candidates filenames min-depth max-depth parents)
>> +    :candidate-transformer (if parents 'reverse nil)
>> +    :after-init-hook 'helm-follow-mode
>
> That works, but it still has the problem I mentioned: the follow
> happens before the preselect,

Did you try using :follow 1 instead of :after-init-hook ?

But anyway, I am not sure it is a good idea to enable follow-mode by
default (not what I want for myself anyway).

-- 
Thierry

thierryvolpiatto: alphapapa <notifications@github.com> writes:

> In helm-org.el:
>
>>    (helm-build-sync-source "Org Headings"
>> -    :candidates (helm-org-get-candidates filenames min-depth max-depth)
>> +    :candidates (helm-org-get-candidates filenames min-depth max-depth parents)
>> +    :candidate-transformer (if parents 'reverse nil)
>> +    :after-init-hook 'helm-follow-mode
>
> That works, but it still has the problem I mentioned: the follow
> happens before the preselect, which causes the point to briefly jump
> to the top of the buffer before returning to the current heading,
> which is a bit distracting and...unpolished. :)
>
> The docstring for after-init-hook says that it runs after the Helm
> buffer is created, which sounds like it should allow activating of
> helm-follow-mode after the preselect runs, but I haven't been able to
> find a way to activate helm-follow-mode from the after-init-hook. Is
> there a way to do this?

Also I remember now I have created helm-after-preselection-hook some
days ago for this same reason, but still using :follow is better IMO.

-- 
Thierry

alphapapa: > Did you try using :follow 1 instead of :after-init-hook ?

Yes, it does activate follow-mode, but it happens before the preselection, causing the point-jumping behavior.

>  But anyway, I am not sure it is a good idea to enable follow-mode by default (not what I want for myself anyway).

I imagine that it's not for everyone, no.  But for my use case, it definitely helps.  For example, when I would use this command, I'm buried somewhere deep in a 12-level outline, so having the buffer follow to each heading as I select it lets me see the context around it, and then jump back to where I was if I `C-g` out of it.  It's a no-harm-done operation that can be very helpful, so that's why I would like to activate it for this command.  But if you think it shouldn't be, it could be left off.  I would like to find a way to offer support for it, though.  :)

> Also I remember now I have created helm-after-preselection-hook some days ago for this same reason, but still using :follow is better IMO.

I noticed that hook, but from my limited understanding, it appears that it's a global hook that would apply to all Helm sources--and definitely not all Helm sources should have follow-mode activated.  Is there a way to use that hook for individual sources?

Thanks.

thierryvolpiatto: alphapapa <notifications@github.com> writes:

> In helm-org.el:
>
>>    (helm-build-sync-source "Org Headings"
>> -    :candidates (helm-org-get-candidates filenames min-depth max-depth)
>> +    :candidates (helm-org-get-candidates filenames min-depth max-depth parents)
>> +    :candidate-transformer (if parents 'reverse nil)
>> +    :after-init-hook 'helm-follow-mode
>
>     Did you try using :follow 1 instead of :after-init-hook ?
>    
> Yes, it does activate follow-mode, but it happens before the
> preselection, causing the point-jumping behavior.

Hmm, ok.

>
>     But anyway, I am not sure it is a good idea to enable follow-mode by default (not what I want for myself anyway).
>    
> I imagine that it's not for everyone, no. But for my use case, it
> definitely helps.

You can use C-<arrow-down/up> also to have the same behavior as
follow-mode, would it fit your needs ?

> But if you think it shouldn't be, it could be left off.

I prefer not having this by default.

> I would like to find a way to offer support for it, though. :)

If the source builder use their own classes, and then use
`helm-make-source' instead of `helm-build-sync-source', one can use
`helm-setup-user-source' in its config to enable as needed follow for
these sources.


>     Also I remember now I have created helm-after-preselection-hook
>     some days ago for this same reason, but still using :follow is
>     better IMO.
>    
> I noticed that hook, but from my limited understanding, it appears
> that it's a global hook that would apply to all Helm sources--and
> definitely not all Helm sources should have follow-mode activated.

Of course.

> Is there a way to use that hook for individual sources?

If your function is a noop for other sources, it is not a problem, e.g:

    (when (string= "my-source-name" (assq 'name (helm-get-current-source)))
       [...])

-- 
Thierry

alphapapa:
Ok, I pushed another commit without the follow-mode.  I'll leave that for my personal Emacs config.  I think it's ready for merging now.  Thanks for all your help.

thierryvolpiatto:
Ok, merged, thanks for your work!

PR #1271, opened 2015-11-21
helm-org.el: Minor cleanup

alphapapa:
This is a bit cleaner.  I should have done it this way in the first place.  :)


thierryvolpiatto:
Thanks.

PR #1286, opened 2015-12-11
Control insertion of kill-ring entry into buffer.

Jehops:
Add a defcustom to control whether an entry selected from
`helm-show-kill-ring` gets inserted into the buffer.  By default,
nothing is different for the user, but when the new defcutsom
`helm-show-kill-ring-yank` is set to nil, selecting an entry from
`helm-show-kill-ring` will not cause that entry to be inserted into the
buffer.  This is useful for users who are using, e.g., clipmon-mode to
synchronize the kill-ring and the X clipboard.  If the user wants to
choose an entry from the kill ring to paste into another application,
s/he can choose it with `helm-show-kill-ring`, then switch to the other
application to paste.

### DIFF
--- a/helm-ring.el
+++ b/helm-ring.el
@@ -57,6 +57,11 @@ If nil or zero (disabled), don't truncate candidate, show all."
   :group 'helm-ring
   :type '(alist :key-type string :value-type function))
 
+(defcustom helm-show-kill-ring-yank t
+  "*Insert entry from helm-show-kill-ring into the buffer."
+  :type '(boolean)
+  :group 'helm-ring)
+
 
 ;;; Kill ring
 ;;
@@ -116,7 +121,8 @@ replace with STR as yanked string."
   (with-helm-current-buffer
     (setq kill-ring (delete str kill-ring))
     (if (not (eq (helm-attr 'last-command helm-source-kill-ring) 'yank))
-        (insert-for-yank str)
+        (when helm-show-kill-ring-yank
+          (insert-for-yank str))
       ;; from `yank-pop'
       (let ((inhibit-read-only t)
             (before (< (point) (mark t))))


thierryvolpiatto:
Joseph Mingrone <notifications@github.com> writes:

> Add a defcustom to control whether an entry selected from
> helm-show-kill-ring gets inserted into the buffer. By default,
> nothing is different for the user, but when the new defcutsom
> helm-show-kill-ring-yank is set to nil, selecting an entry from
> helm-show-kill-ring will not cause that entry to be inserted into the
> buffer. This is useful for users who are using, e.g., clipmon-mode to
> synchronize the kill-ring and the X clipboard. If the user wants to
> choose an entry from the kill ring to paste into another application,
> s/he can choose it with helm-show-kill-ring, then switch to the other
> application to paste.

Yes but she can't yank anymore directly in emacs then ?
Why not creating another action to do this bound to another key (not
RET) ?

-- 
Thierry

thierryvolpiatto:

> Why not creating another action to do this bound to another key (not
> RET) ?

I have created a new action, it is in third position, so to select a
candidate in helm-show-kill-ring for yanking it from another application
external to emacs, just hit F3 to push it on top.
IIUC this is what you need, let me know if this fit your needs and if I
understand correctly what you want.

Thanks.

-- 
Thierry

thierryvolpiatto:

> Thierry Volpiatto <thierry.volpiatto@gmail.com> writes:
>
>> Why not creating another action to do this bound to another key (not
>> RET) ?
>
> I have created a new action, it is in third position, so to select a
> candidate in helm-show-kill-ring for yanking it from another application
> external to emacs, just hit F3 to push it on top.
> IIUC this is what you need, let me know if this fit your needs and if I
> understand correctly what you want.

I have removed the kill-new action as it is already available with `C-c
C-k`.
So the only thing you have to do is `C-c C-k` and then go to external
apps and paste.

Hope that help.

-- 
Thierry

Jehops:
This is much better.  Thanks and sorry for the noise.

PR #1289, opened 2015-12-12
helm-get-org-candidates-in-file: Widen buffer

alphapapa:
If an Org buffer is narrowed, this function will only get headings in the narrowed part of the buffer.  We need to widen the buffer first to get all headings in the buffer.


thierryvolpiatto:
Thanks.

PR #1290, opened 2015-12-13
Fix helm--follow-action's arg calling

issue-dispenser:
* helm.el (helm--follow-action): Use default arg for next/previous line.

This makes it so that `helm-move-to-line-cycle-in-source` is respected.


thierryvolpiatto:
Thanks.

Issue #1292, opened 2015-12-14
Disable confirmation of directory creation with helm-find-files


halbtuerke:
Hi,

I would like to disable the confirmation in helm-find-files when creating a file in a non-existent directory. Right now the following message occurs:

```
Directory ... does not exist; create? (y or n)
```

With Emacs' default `find-files` one can `(setq confirm-nonexistent-file-or-buffer nil)` to disable this confirmation, but this variable seems to be ignored by helm-find-files.

Is there a way to configure this behavior? A quick glance through the source code did not reveal anything promising.

Thanks in advance

halbtuerke:
I went ahead and implemented a small fix.

PR #1293, opened 2015-12-14
Comply with `confirm-nonexistent-file-or-buffer` variable.

halbtuerke:
If `confirm-nonexistent-file-or-buffer` is nil then helm-find-files will
not ask for confirmation to create directories, when trying to create a
file in a non-existent directory.

This fixes https://github.com/emacs-helm/helm/issues/1292

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -2688,7 +2688,8 @@ Called with a prefix arg open files in background without selecting them."
         (find-file-wildcards nil)
         (make-dir-fn
          (lambda (dir &optional helm-ff)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Patrick <notifications@github.com> writes:

> If confirm-nonexistent-file-or-buffer is nil then helm-find-files will
> not ask for confirmation to create directories, when trying to create a
> file in a non-existent directory.
>
> This fixes #1292

I will have a look at this at the end of the week, maybe earlier on thursday,
I have no time actually, thanks.

-- 
Thierry

thierryvolpiatto:
Thanks!

PR #1297, opened 2015-12-16
Fix additional space suffix in `helm--completion-in-region`

markus1189:
Previously, `helm--completion-in-region` always adds a trailing " " to
the input (unless `helm-completion-in-region-fuzzy-match` is true).
This leads to problems e.g. in `sbt-mode` completion, because instead of
`testOnly ` (note the trailing " "), we have `testOnly  ` (two " " at
the end).  I fixed this by adding another check whether the input
already ends with an " ", and if this is the case do not append another
" ".

-- 24pullrequests
[](http://24pullrequests.com/)


thierryvolpiatto:
Looks correct, thanks.

PR #1309, opened 2015-12-31
fix error in helm-search-match-part

cosmicexplorer:
Not sure if there's something deeper to this, but this was causing errors in the extension I was writing. `substring` accepts a string, then an initial index, and the two were transposed for some reason.


thierryvolpiatto:
Thanks!

PR #1311, opened 2016-01-01
Use consistent pretty major name

xuchunyang:
Because `helm-buffer--details` uses pretty major name. 

This should be able to reduce the width of major mode (some helm user asked about this in Emacs SE: http://emacs.stackexchange.com/q/19192/3889) on some level since pretty major name is usually shorter.


thierryvolpiatto:
Thanks!

PR #1312, opened 2016-01-01
Position point at the beginning of the headline

jagrg:
I think it makes sense to place the point at the beginning of the headline. It makes visibility cycling easier when all headlines are folded. It's also convenient for speed keys (`org-use-speed-commands`).


thierryvolpiatto:
Thanks!

Issue #1318, opened 2016-01-05
More actions for helm-apropos


xiongtx:
`helm-apropos` is a very useful command, but there are some actions that are currently not available.

For example, the sole action available for faces is `describe-face`. It is often the case that I want to customize a face after finding it in `helm-apropos`. 

Adding this functionality is fairly simple; in `helm-def-source--emacs-faces` (located in `helm-elisp.el`), change the action to the following:

```emacs-lisp
:action '(("Describe face" . helm-describe-face)
          ("Customize face" . (lambda (candidate)
                                (customize-face (helm-symbolify candidate)))))
```

Let me know if this is a good idea, or if it goes against the goals of `helm-apropos` (variables have a "customize" option, so I'm not sure why faces cannot).

thierryvolpiatto:
Tianxiang Xiong <notifications@github.com> writes:

> helm-apropos is a very useful command, but there are some actions that
> are currently not available
>
> For example, the sole action available for faces is describe-face It
> is often the case that I want to customize a face after finding it in
> helm-apropos
>
> Adding this functionality is fairly simple; in
> helm-df-source--emacs-faces (located in helm-elispel), change the
> action to the following:
>
> :action '(("Describe face" helm-describe-face) ("Customize face"
>           (lambda (candidate) (customize-face (helm-symbolify
>           candidate)))))
>
> Let me know if this is a good idea, or if it goes against the goals of
> helm-apropos (variables have a "customize" option, so I'm not sure why
> faces cannot)

Yes it is a good idea.

-- 
Thierry

xiongtx:
Hey Thierry, at what point is the fix "confirmed"? It's already been merged into `master`, so it should be out there in the wild now.

thierryvolpiatto:
Tianxiang Xiong <notifications@github.com> writes:

> Hey Thierry, at what point is the fix "confirmed"? It's already been
> merged into master, so it should be out there in the wild now.

You can close if you are satisfied with your changes (personally I am
;-)).


-- 
Thierry

PR #1320, opened 2016-01-06
Add commands to `helm-apropos` for faces (#1318)

xiongtx:
Add actions to `helm-apropos` for faces (#1318)

Change all action names for `helm-apropos` sources to be "Sentence
case" instead of "Title Case" for greater consistency.


thierryvolpiatto:
Great, thanks.

Issue #1324, opened 2016-01-07
Improving `helm-info-*`


xiongtx:
I really like the ability to quickly search through documentation with the `helm-info-*` commands. However, I think there are several issues with how the system currently works.

### Too many commands, no unified interace ###
There are lots of individual commands for different info files like `helm-info-elisp` and `helm-info-emacs` and no way to easily choose the one you want. You can narrow down your choices with `helm-M-x`, but that takes a whole bunch of keystrokes.

What we need is a single, unified interface to all the `helm-info-*` commands. I whipped together something:

```emacs-lisp
(defvar helm-info-searched (make-ring 32))

(defun helm-info-search-index (candidate)
  "Search the index of CANDIDATE's Info file using helm-info-<CANDIDATE>."
  (let ((helm-info-function
	 (intern-soft (concat "helm-info-" candidate))))
	(when (fboundp helm-info-function)
	  (funcall helm-info-function)
	  (ring-insert helm-info-searched candidate))))

(defun helm-def-source--info-files ()
  (helm-build-sync-source "Helm Info"
	:candidates (sort (copy-sequence helm-default-info-index-list)
			  #'string-lessp)
	:nomark t
	:action '(("Search index" . helm-info-search-index))))

(defun helm-info ()
  "Preconfigured `helm' for searching Info files indices."
  (interactive)
  (let ((default (ring-ref *-helm-info-searched 0)))
	(helm :sources (funcall #'helm-def-source--info-files)
	  :buffer "*helm Info*"
	  :preselect (and default
			  (concat "\\_<" (regexp-quote default) "\\_>")))))
```

Can we get it more polished and included in `helm`?

### Lots of commands don't actually work ###
Looking at `helm-info.el`, it seems that the `helm-info-*` commands are auto-generated by `helm-define-info-index-sources`, which looks at `helm-default-info-index-list`, a customizable list of info file names, to determine which commands to generate.

However, depending on the user's system, lots of commands don't actually work. `helm-info-ratpoison`, for example, does nothing because I don't have the corresponding Info file.

Although the list is customizable, I'm wondering if there is a way to automatically extract all available Info files so the user doesn't have to worry about it. 

I know the `dir` file (e.g. "/user/share/info/dir") maintains a directory of Info files; might we have a customizable list of paths to `dir` files such that we can automatically detect which Info files are valid from these `dir` files?

tuhdo:
:+1: 

[...]

PR #1326, opened 2016-01-09
Add `helm-info` for Info file to search (#1324).

xiongtx:
Info files are no longer hard-coded by default, but found by searching
through directories in `Info-directory-list`.


thierryvolpiatto:
Great, thanks!

PR #1335, opened 2016-01-13
Fix helm-file-name-sans-extension for files beginning with a dot

kleinmann:
A file beginning with a dot (`.`) and no further dots should be treated as a file without any extension.

Currently it keeps recursing until it hits `max-lisp-eval-depth`. It happened to me with a file named `._libtool.info`. The error disappeared after the change.

I'm unsure if this introduces any other problems. It seemed like an easy enough fix, so this is just a quick PR. Please tell me if I overlooked anything :-)

### DIFF
--- a/helm-lib.el
+++ b/helm-lib.el
@@ -424,7 +424,7 @@ See `kill-new' for argument REPLACE."
 (defun helm-file-name-sans-extension (filename)
   "Same as `file-name-sans-extension' but remove all extensions."
   (helm-aif (file-name-sans-extension filename)
-      (if (string-match "\\." it)
+      (if (string-match "[^\\.]+\\." it)
           (helm-file-name-sans-extension it)
           it)))
 


thierryvolpiatto:
Uwe Kleinmann <notifications@github.com> writes:

> A file beginning with a dot (.) and no further dots should be treated
> as a file without any extension.

Many thanks to point me on this, I forgot to handle this case.

> Currently it keeps recursing until it hits max-lisp-eval-depth. It
> happened to me with a file named ._libtool.info. The error disappeared
> after the change.

Yes of course.

> I'm unsure if this introduces any other problems.

Didn't try but should not.

> It seemed like an easy enough fix, so this is just a quick PR. Please
> tell me if I overlooked anything :-)

No your change looks ok, but even simpler IMO is just searching starting
at index 1 of string. See 0a3728353710de7853d75bce2d97bd7213ef0b6e.

Thanks.

-- 
Thierry

kleinmann:
I agree, your change looks better than mine. I don't know the functions all that well so far ;-)
I'm closing the PR as it is now fixed. Thanks!

PR #1337, opened 2016-01-16
Change helm-recentf-source persistent action

issue-dispenser:
* helm-files.el (helm-recentf-source): Add PA initform.

Similar to 84e7fc3


thierryvolpiatto:
Thanks.

PR #1338, opened 2016-01-16
Remove mark messages

seanirby:
Added an optional argument to ```helm-mark-all``` and ```helm-unmark-all``` to skip the message.  Considering that there is already visual feedback for these actions, I don't think the message is necessary but I figured I make it optional.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -5287,21 +5287,18 @@ Argument ACTION if present will be used as second argument of `display-buffer'."
                            (forward-line 1))
                          (forward-line 1))
                      (end-of-line))))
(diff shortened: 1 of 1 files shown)


seanirby:
I notice that the message for '''helm-mark-all''' is currently getting overwritten anyways. 
https://github.com/emacs-helm/helm/blob/master/helm.el#L5291-L5292

Updated PR to just remove these messages.

thierryvolpiatto:
Indeed we can now get rid of these messages now with have a visual info in mode-line.

Thanks!

PR #1343, opened 2016-01-19
Fix bugs in 'helm-revive-visible-mark' function

seanirby:
I've found some issues with ```helm-revive-visible-mark``` and this PR addresses them.  You can use the code below to repro.

```
;; Not a very useful transformer but its just for illustration purposes
(setq xfrmr (lambda (candidates source)
              (helm-get-candidates source)))

;; `helm-revive-visible-mark' doesn't respect the 'allow-dups slot and it
;; doesn't match candidates well for candidates with only a display value.

;; To reproduce, select all candidates and then enter a character into the pattern prompt.
;; You'll see that dups get visibly unmarked, and the "0" gets unmarked as well because
;; it matches the "0" chars within the subsequent "100" candidates
(helm :sources (helm-build-sync-source "Marking Bugs Demo"
                 :allow-dups t
                 :candidates (cons "0" (make-list 10 "100"))
                 :filtered-candidate-transformer xfrmr))
```

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -5358,22 +5358,24 @@ When key WITH-WILDCARD is specified try to expand a wilcard if some."
   (with-current-buffer helm-buffer
     (save-excursion
       (cl-dolist (o helm-visible-mark-overlays)
-        (goto-char (point-min))
-        (while (and (search-forward (overlay-get o 'string) nil t)
-                    (not (overlays-at (point-at-bol 0)))
-                    (helm-current-source-name= (overlay-get o 'source)))
-          ;; Calculate real value of candidate.
-          ;; It can be nil if candidate have only a display value.
-          (let ((real (get-text-property (point-at-bol 0) 'helm-realvalue)))
-            (if real
-                ;; Check if real value of current candidate is the same
-                ;; than the one stored in overlay.
-                ;; This is needed when some cands have same display names.
-                ;; Using equal allow testing any type of value for real cand.
-                ;; Issue (#706).
-                (and (equal (overlay-get o 'real) real)
-                     (move-overlay o (point-at-bol 0) (1+ (point-at-eol 0))))
-                (move-overlay o (point-at-bol 0) (1+ (point-at-eol 0))))))))))
+        (let ((o-str (overlay-get o 'string)))
+          (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Sean Irby <notifications@github.com> writes:

> I've found some issues with helm-revive-visible-mark and this PR addresses them. You can use the code below to repro.
>
> ;; Not a very useful transformer but its just for illustration purposes
> (setq xfrmr (lambda (candidates source)
>               (helm-get-candidates source)))
>
> ;; `helm-revive-visible-mark' doesn't respect the 'allow-dups slot and it
> ;; doesn't match candidates well for candidates with only a display value.
> ;; The test below illustrates this:
>
> ;; To reproduce, select all candidates and then enter a character into the pattern prompt.
> ;; You'll see that dups get visibly unmarked, and the "0" gets unmarked as well because
> ;; it matches the "0" chars within the subsequent "100" candidates
> (helm :sources (helm-build-sync-source "Marking Bugs Demo"
>                  :allow-dups t
>                  :candidates (cons "0" (make-list 10 "100"))
>                  :filtered-candidate-transformer xfrmr))

Thanks to catch this.
This is not related with allow-dups, the bug is reproductible also
without it.
The reason is we have indeed duplicate candidates on startup.
When helm starts the loop it find indeed the first "100" and put overlay
on it then it goes back at point-min and search again the same string,
move-overlay on next line and so on until last candidate, so we ends up
with the last marked but not the others, note that the list of marked is
reversed.
I have commited a simple fix that check if candidate have been already
marked, if so, it switch to the next etc...

-- 
Thierry

seanirby:
Cool, thanks for the explanation.   Learning how helm works has been a great way to learn elisp :)

I don't think your fix addresses the other issue with the example code.   I realize I could have given a better example to illustrate.  This time, instead of selecting all the candidates, just select the first 3.  You'll see the "0" gets unselected because it matches on "0" as well as on all the subsequent "100" elements. Not a problem if your candidates have (DISPLAY. VALUE) pairs,  but its a problem if they just have a display name.  

I think  if the candidate doesn't have a 'real' value, you should try comparing against the display value instead.  Updated PR to include this check.

thierryvolpiatto:
Good, thanks.

PR #1349, opened 2016-01-21
Clean up dependencies

syohex:
helm supports Emacs 24.3 and higher versions and cl-lib was bundled
since Emacs 24.3.


thierryvolpiatto:
Thanks.

PR #1350, opened 2016-01-21
Remember prompt value for helm-resume

seanirby:
If a user sets a custom prompt with ```:prompt``` it will not be remembered if ```helm-resume``` is called.  This PR fixes that by setting a buffer-local value for ```helm-prompt```.


thierryvolpiatto:
Thanks.

PR #1357, opened 2016-01-25
Better completion in region

thierryvolpiatto:


issue-dispenser:
Hi, thanks for this -- it fixed a bug I was having in haskell-mode.

Currently it also introduces a new bug -- if the string between start and end isn't the prefix of the completion, then I get [No matches].

For instance, when I did `helm--completion-at-point` in a Haskell repl with `:cd ` I used to get all of the directories in the current directory, (which is what `(helm--completion-at-point 313 313 '(<list of stuff>))` got me, with 313 being the cursor position).

Now it gives me [No matches].

thierryvolpiatto:
issue-dispenser <notifications@github.com> writes:

> Hi, thanks for this -- it fixed a bug I was having in haskell-mode.
>
> Currently it also introduces a new bug -- if the string between start and end isn't the prefix of the completion, then I get [No matches].
>
> For instance, when I did helm--completion-at-point in a Haskell repl with :cd I used to get all of the directories in the current directory, (which is what
> (helm--completion-at-point 313 313 '(<list of stuff>)) got me, with 313 being the cursor position).
>
> Now it gives me [No matches].

Can't try with haskell but I have commited a fix that may work when you
try to :cd with your haskell repl.
Let us know if it works.

Thanks.

-- 
Thierry

issue-dispenser:
It appears like it's working now, thanks.

thierryvolpiatto:
@michael-heerdegen Perhaps this can be merged now ?

michael-heerdegen:

> Perhaps this can be merged now ?

I still want to have a final look.  I'll merge it tomorrow, ok?

thierryvolpiatto:

>
>> Perhaps this can be merged now ?
>
> I still want to have a final look. I'll merge it tomorrow, ok?

Ok great.

Thanks.

-- 
Thierry

michael-heerdegen:
The only question that raised while I read the changes again was whether it could be worth it to honor the `display-sort-function` the completion metadata might include.

thierryvolpiatto:

> The only question that raised while I read the changes again was
> whether it could be worth it to honor the display-sort-function the
> completion metadata might include.

We could add something like this in fc-transformer:

    (append (list 'helm-cr-default-transformer)
            (list (helm-aif (completion-metadata-get metadata 'display-sort-function)
                      (lambda (candidates _source)
                        (funcall it candidates))
                    (lambda (candidates _source)
                      (sort candidates 'helm-generic-sort-fn)))))

even if in practice `display-sort-function` seems rarely provided in
metadata's.

-- 
Thierry

michael-heerdegen:
Would we have to care about the fact that candidates in helm can include conses (I guess we must)?

michael-heerdegen:
But actually, it seems to be used nowhere in vanilla Emacs.  So it's probably not worth it to support that feature.

thierryvolpiatto:

> Would we have to care about the fact that candidates in helm can include conses (I guess we must)?

Oh yes, right, in a similar way I did in the generic sort fn.

But as you said, probably it is prematurate for now to add this, let
emacs provide some completions with this first.

-- 
Thierry

PR #1361, opened 2016-01-26
Make helm-source-org-headings-for-files more versatile.

SkydiveMike:
Create var `helm-org-headings-persistent-help' which can be overriden
allowing to define custom help strings for the custom actions defined in
`helm-org-headings-actions'.

Adds functionality to fbbcb15acfee9a0c368ada0884f228562d0f6c77.

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -86,6 +86,15 @@ Note this have no effect in `helm-org-in-buffer-headings'."
   :group 'helm-org
   :type '(alist :key-type string :value-type function))
 
+(defcustom helm-org-headings-persistent-help
+  "Go to line (keeping session); <f1> Go to line; <f2> Refile to this heading; <f3> Insert link to this heading"
+  "Default Persistent Help string for
+  `helm-source-org-headings-for-files'. The contents of this
+  string should match the actions defined in
+  `helm-org-headings-actions'"
+  :group 'helm-org
+  :type 'string)
+
 (defun helm-source-org-headings-for-files (filenames &optional parents)
   (helm-build-sync-source "Org Headings"
     :candidates filenames ; Start with only filenames.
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Sorry, but as long as we don't add a specific persistent action we don't have to add such a message, the default one is perfectly fine.

Thanks anyway for your PR.

SkydiveMike:
Isn’t the whole point of https://github.com/emacs-helm/helm/commit/fbbcb15acfee9a0c368ada0884f228562d0f6c77 to add specific persistent actions? Why would helm allow adding actions, but not documenting them?

thierryvolpiatto:
SkydiveMike <notifications@github.com> writes:

> Isn’t the whole point of fbbcb15 to add specific persistent actions?

No.

-- 
Thierry

SkydiveMike:
My apologies, I pasted the wrong commit (in both my PR’s commit message and therefore here in the discussion. Clearly this caused confusion. I'm sorry!

Isn’t the whole point of https://github.com/emacs-helm/helm/commit/603fc973db0764cd1d88db6c56f86748c62d0c8c to allow the adding of specific persistent actions? Why would helm allow adding actions, but not documenting them?

For example if I customize `helm-org-headings-actions` using the new `defcustom` introduced in https://github.com/emacs-helm/helm/commit/603fc973db0764cd1d88db6c56f86748c62d0c8c , I have, indeed, added a specific persistent action to `helm-source-org-headings-for-files` and am unable to *also* create a persistent-help line to show those (new or altered) actions.

thierryvolpiatto:
The actions that have been added are not persistent actions, you have their descriptions when hitting TAB.
If you want to document something about actions, the place to do this is helm-help.el, this will appear in embeded documentation when you hit C-h m.

Issue #1366, opened 2016-01-30
helm-comp-read much slower than completing-read-default.


vibrys:
Hallo Thierry & Helm Team.

I'm trying to adapt helm to some existing emacs module (rtags).

This module uses completing-read-default to get candidates in one of the commands.

When `helm-comp-read' is used, candidates' collection takes much longer.

The reason is that `helm-comp-read-get-candidates' calls the collection function with empty string as first parameter, while `completing-read-default' puts the user input string there. As a result rtags collection function needs to do much more than it would have to do?

The call is done at helm-mode.el(3adbdfa):253.

Do You think helm could pass the input string to collection function?

best regards

thierryvolpiatto:
vibrys <notifications@github.com> writes:

> Do You think helm could pass the input string to collection function?

See #1369, (not fully finished and tested).

-- 
Thierry

vibrys:
Thank You

thierryvolpiatto:
I have reverted some changes here that break some completing-read's functions #1371 .

thierryvolpiatto:
We should not pass helm-pattern when computing function.
I think you see slowdown because you are wrongly using completing-read.
Please provide recipe to show where we experiment a slowdown, closing for now.

vibrys:
rtags backend I'm talking about handles large database of C++ source code symbols.
It uses the pattern to return the subset of elements from the database.

The following recipe does the same thing.

Please for example use "virtu" pattern for both the cases.

The result is the same for both of them,
but under the hood, things are done differently.

helm version based on empty pattern, always returns the whole 3 patterns and filtering is done at postprocessing,
while two non-helm versions do the preprocess within the lambda function provided, based on non-empty pattern.

Please see *Messages" for both the cases.

For large results from backend the emacs frontend has much more to do, thus slowdown.

<pre>
(require 'cl)
;; emacs-helm.sh
(helm-comp-read "->" (completion-table-dynamic
                      (lambda (pattern)
                        (message (concat "pattern is: " pattern))
                        (let ((result (delete-if 'null (mapcar
                                                        (lambda (param) (and (string-match (concat "^" pattern) param) param))
                                                        '("Inher::Inher" "Inher::m_field" "virtual void Inher::fun(int)")))))
                          (message (format "number of results: %d" (length result)))
                          result)))
                :requires-pattern 1
                :fuzzy nil
                )

;; emacs -Q
;; (icomplete-mode) - extra option.
[...]

PR #1369, opened 2016-01-31
Use initial input in collection in completing-read's (#1366).

thierryvolpiatto:
* helm-mode.el (helm-comp-read-get-candidates): New arg input.
(helm-comp-read): Use it.
(helm-completing-read-default-1): Use input in initial setting of collection.
(helm-completing-read-with-cands-in-buffer): Same.

### DIFF
--- a/helm-mode.el
+++ b/helm-mode.el
@@ -250,7 +250,7 @@ If COLLECTION is an `obarray', a TEST should be needed. See `obarray'."
                            collect (concat (file-name-as-directory
                                             (helm-basedir helm-pattern)) f)))
                  ((functionp collection)
-                  (funcall collection "" test t))
+                  (funcall collection input test t))
                  ((and alistp test)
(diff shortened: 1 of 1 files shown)


helm-mode.el
```diff
@@ -254,7 +254,7 @@ If COLLECTION is an `obarray', a TEST should be needed. See `obarray'."
                  ((and alistp test)
```
vibrys: input param missing in line above.

vibrys:
Thierry,
thank You for the fix. I left the comment above.

helm-mode.el
thierryvolpiatto: vibrys <notifications@github.com> writes:

> In helm-mode.el:
>
>> @@ -254,7 +254,7 @@ If COLLECTION is an `obarray', a TEST should be needed. See `obarray'."
>>                   ((and alistp test)
>
> input param missing in line above.

Indeed, fixed, thanks.

-- 
Thierry

Issue #1376, opened 2016-02-05
Using custom keymaps with custom actions?


alphapapa:
I'm trying to make `C-RET` activate a custom action in a custom source.  The action works fine when activated with `F2` or from the action list, but I can't get it to work correctly from the keymap.

I have it almost working: it activates the action, but it leaves the helm buffer open.  

I've tried to dig through the helm source and follow the path of execution, but I got lost around the call to `(exit-minibuffer)`.  I guess somewhere a hook is added to helm's exit-minibuffer hook that actually calls the selected action, but I'm getting a bit worn out with this.  :)  So I would greatly appreciate any help.

Maybe the ideal solution would be a way to define a key directly in the `:action` list, instead of only separately in a custom keymap...

Here's what I have so far:
```el
(defun my/helm-org-open-heading-in-indirect-buffer (marker)
  (interactive)  ; Is this necessary here?

  (helm-org-goto-marker marker)

  ;; Open in current frame, not a new one
  (let ((org-indirect-buffer-display 'current-window))
    (org-tree-to-indirect-buffer))

  ;; Put the non-indirect buffer at the bottom of the prev-buffers
  ;; list so it won't be selected when the indirect buffer is killed
  (set-window-prev-buffers nil
                           (append (cdr (window-prev-buffers))
                                   (car (window-prev-buffers)))))

(cl-defun my/helm-source-org-headings-for-files (filenames
                                                  &optional (min-depth 1) (max-depth 8) (parents nil))
  (helm-build-sync-source "Org Headings"
    :candidates (helm-org-get-candidates filenames min-depth max-depth parents)
    :candidate-transformer (and parents 'reverse)
    :action '(("Go to heading" . helm-org-goto-marker)
              ("Open in indirect buffer" . my/helm-org-open-heading-in-indirect-buffer)
              ("Refile to this heading" . helm-org-heading-refile)
              ("Insert link to this heading" . helm-org-insert-link-to-heading-at-marker))
              :keymap my/helm-org-map))

(defun my/helm-org-in-buffer-headings ()
  "Preconfigured helm for org buffer headings."
  (interactive)
  (let ((helm-org-headings--nofilename t))
    (helm :sources (my/helm-source-org-headings-for-files (list (current-buffer)))
          :candidate-number-limit 99999
          :buffer "*helm org inbuffer*")))

(setq my/helm-org-map
      (let ((map (make-sparse-keymap)))
        (set-keymap-parent map helm-map)
[...]

PR #1380, opened 2016-02-07
Helm org headings enhancements

alphapapa:
As discussed in https://github.com/emacs-helm/helm/issues/1376, here are a few minor enhancements.  I have one question, which I added a comment about...


thierryvolpiatto:
Quoting the actions variable have no effect as the source is build at runtime inside a function, but as soon we will make a new class or bind the source to a variable, we will need the variable to be quoted to be evaluated at each time it is modified. (I would like a class for this).

Thanks for your work.

Issue #1386, opened 2016-02-10
Is `helm-exit-minibuffer-hook` working correctly?


xiongtx:
I'd like to turn `popwin-mode` off when using Helm, and turn it back on after leaving Helm. This is because Helm's persistent action displays Help buffers in the other window. When `popwin-mode` is active, both the persistent action's Help buffer *and* the `popwin` Help buffer are shown.

I've done the following:

```emac-lisp
(defun *-popwin-help-mode-off ()
  (when (boundp 'popwin:special-display-config)
    (customize-set-variable 'popwin:special-display-config
			    (delq 'help-mode popwin:special-display-config))))

(defun *-popwin-help-mode-on ()
  (when (boundp 'popwin:special-display-config)
    (customize-set-variable 'popwin:special-display-config
                            (push 'help-mode popwin:special-display-config))))

(add-hook 'helm-minibuffer-set-up-hook #'*-popwin-help-mode-off)
(add-hook 'helm-exit-minibuffer-hook #'*-popwin-help-mode-on)
```

Turning `popwin-mode` off for Helm works fine. Turning it back on does not work, however, when exiting Helm with <kbd>C-g</kbd>. It *does* work when exiting Helm by selecting a candidate.

This is not what I would expect from the documentation:

> Run just before exiting minibuffer.

Quitting seems as valid a method of "exiting" as selecting a candidate.

I've gotten around this by using `helm-cleanup-hook` instead. Its documentation is:

> Run after helm minibuffer is closed. IOW this hook is executed BEFORE performing action.

Is this the right hook to use? What does "IOW" mean?

thierryvolpiatto:
Tianxiang Xiong <notifications@github.com> writes:
    
> Is this the right hook to use? 

Yes.

> What does "IOW" mean?

In Other Words.

-- 
Thierry

xiongtx:
Can we update the documentation of `helm-exit-minibuffer-hook` to:

> Run just before exiting the minibuffer. This hook is run if the minibuffer is exited normally (e.g. via candidate selection). It is NOT run if the minibuffer is exited abnormally (e.g. via `helm-keyboard-quit`).

And that of `helm-cleanup-hook` to:

> Run after exiting the minibuffer and before performing an action. This hook is run even if the minibuffer is exited abnormally (e.g. via `helm-keyboard-quit`).

PR #1387, opened 2016-02-10
Add documentation to hook variables.

xiongtx:
Address #1386 by adding documentation to `helm-exit-minibuffer-hook` and
`helm-cleanup-hook`.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -708,8 +708,11 @@ This is very similar to `helm-update-hook' but selection is not moved.
 It is useful to select a particular object instead of the first one.")
 
 (defvar helm-cleanup-hook nil
-  "Run after helm minibuffer is closed.
-IOW this hook is executed BEFORE performing action.")
+  "Run after exiting the minibuffer and before performing an
+action.
+
+This hook is run even if the minibuffer is exited
+abnormally (e.g. via `helm-keyboard-quit`).")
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks, more generally don't modify indentation, trailing white spaces etc... in same commit, it make hard the read of the diff.

Thanks.

xiongtx:
OK, but I have `whitespace-cleanup` in my `before-save-hook` to get rid of excess whitespace in my files before they are committed. 

This seems like a good idea in general. Is it something we can adopt as a standard for the project?

thierryvolpiatto:

> OK, but I have whitespace-cleanup in my before-save-hook to get rid of
> excess whitespace in my files before they are committed.
>
> This seems like a good idea in general. Is it something we can adopt
> as a standard for the project?

No thanks.

-- 
Thierry

xiongtx:
*Mea culpa*.

By the way, you can see the [diff with whitespace ignored](https://github.com/blog/967-github-secrets) by appending a `?w=1` do the url.

thierryvolpiatto:

> By the way, you can see the diff with whitespace ignored by appending a ?w=1 do the url.

Good to know, thanks.

-- 
Thierry

PR #1392, opened 2016-02-15
Remove trailing newline in helm-copy-to-buffer

xuchunyang:
I found the trailing newline is annoying mainly when there is only one selected candidate, so I want to get rid of it.


thierryvolpiatto:
Merged, sorry for late reply, I just forget about it.

Thanks.

thierryvolpiatto:
Small inconvenient with this, when pressing C-c C-i  more than one time from a source where marking is not allowed e.g helm-apropos, there is no separator between the items inserted (was a newline before).

PR #1403, opened 2016-02-22
add space in front of buffer names (emacs conv. for ephemeral buffers)

abred:
According to the GNU emacs manual the names of ephemeral buffers should start with a space
(https://www.gnu.org/software/emacs/manual/html_node/elisp/Buffer-Names.html)


abred:
would cause other problems, helm buffers should not be ephemeral

PR #1409, opened 2016-02-29
Update guide to enable helm-esh-pcomplete

fm4d:


thierryvolpiatto:
Thanks, I will update the wiki accordingly.
`pcomplete` was the command to use on old emacs when I wrote helm-eshell.el.

Issue #1414, opened 2016-03-03
describe-* not describing


TheBB:
In latest master, the describe functions don't appear to work. They open a helm session but when picking a candidate no help is displayed.

Git bisect blames https://github.com/emacs-helm/helm/commit/d72d7dcd3e8bcfd83c0cf1c58bacfcfec631f4fa

TheBB:
It looks like this affects more than just describe, which is not surprising since that commit changes some very central things.

xenodium:
helm-M-x is affected also.

Reverting https://github.com/emacs-helm/helm/commit/d72d7dcd3e8bcfd83c0cf1c58bacfcfec631f4fa worked around it for me.

KZiemian:
I also have problem with helm-M-x, when I press RET it say ''Quit'' in minibuffer. I also try to reverse changes in helm.el, but don't work in my case. Maybe I just do something wrong when I was rewriting code? I will try to find my way out.

thierryvolpiatto:
Fixed by PR #1416 

Thanks to report.

Issue #1412, opened 2016-03-03
`RET` with `helm-find-files` and `ag`


xiongtx:
In Helm 1.9.2, the `Grep current directory with AG` action for `helm-find-files` does not allow jumping to the candidate's position with `RET`. Instead, `RET` clears the minibuffer.

Invoking the `Find file` action in the `*helm ag*` buffer does the right thing. So does pressing `RET` when calling `helm-do-grep-ag` directly.

thierryvolpiatto:
Tianxiang Xiong <notifications@github.com> writes:

> In Helm 1.9.2, the Grep current directory with AG action for
> helm-find-files does not allow jumping to the candidate's position
> with RET. Instead, RET clears the minibuffer.

Seems you have catched a very old bug, it happens not only with AG, but
with any actions that runs in this context.

Will fix ASAP.

Thanks.

-- 
Thierry

cwhatley:
The commit seems to have broken helm-M-x.

thierryvolpiatto:
Now fixed by #1416

aiguofer:
whew! took me forever to track down that the problem was in `helm-M-x` how long does it take for changes like this to make it into melpa?

xiongtx:
Confirmed as fixed.

*Tip*: to see the changes after upgrading the package, it's necessary to restart Emacs (or manually evaluate the relevant Emacs Lisp code...but restarting is probably easier!).

thierryvolpiatto:
Tianxiang Xiong <notifications@github.com> writes:

> Confirmed as fixed.
>
> Tip: to see the changes after upgrading the package, it's necessary to
> restart Emacs

It is always true not only for helm but all other packages after upgrade.

Thanks.

-- 
Thierry

thierryvolpiatto:
Diego Fernandez <notifications@github.com> writes:

> how long does it take for changes like this to make it into melpa?

See http://melpa.org/

You will see "Current build started: an hour ago, last took 5 hours"

Builds are made by alphabetically order helm is at "h"...

-- 
Thierry

PR #1416, opened 2016-03-03
Fix return value of helm-execute-selection-action

### LINKED ISSUE #1415: helm-M-x broken
helm-M-x suddenly stopped working for me, when I select command in helm-M-x window nothing happens, window dissapears as usually but command is not run. There is nothing in messages or backtrace. Other helm commands and helmized commands seems to be fine, only helm-M-x and helm-run-external-command are behaving like this. Runing helm with `emacs-helm.sh` leads to same result. I updated helm from MELPA after 2-3 days so it must be something recent I guess.

xuchunyang:
The bug is introduced in commit d72d7dcd3e8bcfd83c0cf1c58bacfcfec631f4fa which is made to solve issue #1412.

Fixes #1415 and #1414.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -3678,8 +3678,8 @@ function."
   ;; Position can be change when `helm-current-buffer'
   ;; is split, so jump to this position before executing action.
   (helm-current-position 'restore)
-  (helm-execute-selection-action-1)
-  (helm-log-run-hook 'helm-after-action-hook))
+  (prog1 (helm-execute-selection-action-1)
+    (helm-log-run-hook 'helm-after-action-hook)))
 
 (defun helm-execute-selection-action-1 (&optional
                                         selection action


mordocai:
Can confirm manually applying this to my local helm fixed #1415 for me

thierryvolpiatto:
Thanks to fix this, didn't have the time to fully test my last changes this morning.

arecker:
Updated this morning and was affecting me as well.  Defining this function applying the patch manually worked for me

(or just defining this function somewhere after helm is loaded)

```lisp
  (defun helm-execute-selection-action ()
    "Hack to fix a helm bug"
    "Execute current action."
    (helm-log-run-hook 'helm-before-action-hook)
    ;; Position can be change when `helm-current-buffer'
    ;; is split, so jump to this position before executing action.
    (helm-current-position 'restore)
    (prog1 (helm-execute-selection-action-1)
      (helm-log-run-hook 'helm-after-action-hook)))
```

Issue #530, opened 2014-05-28
How to properly enable helm-follow-mode


tuhdo:
I want to enable `helm-follow-mode` in `helm-occur` and `helm-multi-occur` by doing so:

``` lisp
(add-hook 'helm-after-initialize-hook
          #'(lambda () (helm-attrset 'follow 1 helm-source-occur)))
(add-hook 'helm-after-initialize-hook
          #'(lambda () (helm-attrset 'follow 1 helm-source-moccur)))
```

It works fine in my running Emacs. However, after restarting Emacs, this error occurs:

> Wrong type argument: consp, nil

How do I setup properly?

skparkes:
For anyone else that ends up here, not all of the helm sources have initialization functions.  In particular, it can make a lot of sense to turn on `follow-mode` for `helm-source-grep`, but `helm-source-grep` is initialized to `nil` instead of a dummy source.  I ended up going with this to get always-on follow-mode for greps:
```elisp
(setq helm-source-grep (helm-build-dummy-source "init_grep" :follow 1))
(add-hook 'helm-before-initialize-hook
          (lambda () (helm-attrset 'follow 1 helm-source-grep)))
```
I imagine the `:follow 1` is superfluous, but I'm too worn out after chasing this down to confirm/understand that well enough.

thierryvolpiatto:
skparkes <notifications@github.com> writes:

> For anyone else that ends up here, not all of the helm sources have
> initialization functions. In particular, it can make a lot of sense to
> turn on follow-mode for helm-source-grep, but helm-source-grep is
> initialized to nil instead of a dummy source.

It is not the way to do, dummy sources are not for this.

> I ended up going with this to get always-on follow-mode for greps:
>
> (setq helm-source-grep (helm-build-dummy-source "init_grep" :follow 1))
> (add-hook 'helm-before-initialize-hook
>           (lambda () (helm-attrset 'follow 1 helm-source-grep)))

This hack would work only for grep, please see:

https://github.com/emacs-helm/helm/wiki/FAQ#why-is-a-customizable-helm-source-nil

For the documentation of helm-follow-mode, yes it is deprecated and
should be fixed.

For the helm grep case, I will commit a change that will allow you to
do only:

(defmethod helm-setup-user-source ((source helm-grep-class))
  (set-slot-value source 'follow 1))

Will be ready soon.

-- 
Thierry

thierryvolpiatto:
See also helm-follow-mode-persistent.

-- 
[...]

PR #1419, opened 2016-03-05
Expand follow-mode doc string.

skparkes:
Do this because it was not clear to me at first why the suggested hooks were not working in my initialization scripts.

Googling eventually led me to emacs-helm/helm#530, and more code reading led me to understand the issue was all about when `helm-source` objects are usefully initialized.

I would like to try and save future hackers that period of similar frustration.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -5537,7 +5537,13 @@ e.g:
 \(add-hook 'helm-before-initialize-hook
           (lambda () (helm-attrset 'follow 1 helm-source-buffers-list)))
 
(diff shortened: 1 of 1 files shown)


skparkes:
Depending on your appetite for this, I was considering another issue/pull request to propose swapping:

```elisp
(defvar helm-source-grep nil)
```
to
```elisp
(setq helm-source-grep (helm-build-dummy-source "init_grep"))
```

To better allow setting attributes from the get-go.

skparkes:
@thierryvolpiatto has kindly updated the errant docstring in the linked commits.  I'm closing down this pull request.

PR #1426, opened 2016-03-12
Fixing typo in use of cl-typecase.

malisper:
Previously instead of using 'string' the code used 'stringp'. The
extra 'p' caused cl-typecase to expand incorrectly. This is obvious
when macroexpanding the old definition and the new one.

### DIFF
--- a/helm-lib.el
+++ b/helm-lib.el
@@ -350,9 +350,9 @@ ARGS is (cand1 cand2 ...) or ((disp1 . real1) (disp2 . real2) ...)
   "Return the representation of ELM as a string.
 ELM can be a string, a number or a symbol."
   (cl-typecase elm
-    (stringp elm)
-    (numberp (number-to-string elm))
-    (symbolp (symbol-name elm))))
+    (string elm)
+    (number (number-to-string elm))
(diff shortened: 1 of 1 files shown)


malisper:
I'm getting a different result:

```
(macroexpand
 '(cl-typecase elm
   (stringp elm)
   (numberp (number-to-string elm))
   (symbolp (symbol-name elm))))

(cond
  ((stringp-p elm)
   elm)
  ((numberp-p elm)
   (number-to-string elm))
  ((symbolp-p elm)
   (symbol-name elm)))
```

This pretty much broke emacs for me.

I'm using emacs 24.3.1.

thierryvolpiatto:
Michael Malis <notifications@github.com> writes:

> I'm using emacs 24.3.1.

Hmm, so that's the reason, indeed the both syntax are not supported on
24.3 (among other things).

Will merge tomorrow, still at least one cl-typecase to fix also for
24.3.

Thanks.

-- 
Thierry

thierryvolpiatto:
Merged, thanks.
I have modified some more cl-typecase for compatibility.

Issue #1421, opened 2016-03-07
helm-org breaking on call to org-get-outline-path


NJBS:
It looks like this line specifies 3 arguments to org-get-outline-path, while org-get-outline-path only accepts two, https://github.com/emacs-helm/helm/blob/master/helm-org.el#L194.

thierryvolpiatto:
NJBS <notifications@github.com> writes:

> It looks like this line specifies 3 arguments to org-get-outline-path,
> while org-get-outline-path only accepts two,
> https://github.com/emacs-helm/helm/blob/master/helm-org.el#L194.

Probably your org version is outdated, last versions takes 3 args.
On emacs-24.5 it takes 3 args.

-- 
Thierry

NJBS:
Ah, there appears to be a clash.

http://orgmode.org/cgit.cgi/org-mode.git/tree/lisp/org.el#n11628
http://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/org/org.el#n11430

TheBB:
So it's not an outdated org version, it's too new.

Would helm accept a forward compatibility patch? Presumably this must be fixed sooner or later anyway.

thierryvolpiatto:
Eivind Fonn <notifications@github.com> writes:

> So it's not an outdated org version, it's too new.
>
> Would helm accept a forward compatibility patch? Presumably this must be fixed sooner or later anyway.

No, please update your org copy.

-- 
Thierry

TheBB:
I have. I'm not sure you understood me correctly. Helm works with the *current* org but not the *new* (8.3.4 in my case). I'm proposing a patch to make it work with both. At some point the new org will become the current and this bug must be addressed.

TheBB:
I should also note that 8.3.4 is a stable release and not a development version or anything like that.

thierryvolpiatto:
Eivind Fonn <notifications@github.com> writes:

> I have. I'm not sure you understood me correctly. Helm works with the
> current org but not the new (8.3.4 in my case).

Ah, ok for misunderstanding.

> I'm proposing a patch to make it work with both.  At some point the
> new org will become the current and this bug must be addressed.

Patch is welcome, but I see the arguments have not the same meaning as
before so I am afraid we will break helm-org if we don't understand the
meaning of the new arguments (I don't personally, I have to read the new
org code before and get it of course).

Every two or three years org break helm-org.el...boring!

-- 
Thierry

TheBB:
[...]

PR #1429, opened 2016-03-16
Adapt helm-org to recent changes in org-get-outline-path

aroig:
The signature and behaviour of `org-get-outline-path` has changed in org mode's master branch, and was last touched by this [commit](http://orgmode.org/cgit.cgi/org-mode.git/commit/lisp/org.el?id=105a44669716abd50377fdcfd03a9a11dc388b16).

This PR adapts to the new behaviour, making use of the newly introduced cache to avoid backtracking. This addresses #1421 for org-mode's current master only. I'm not sure how to be compatible with both new and old `org-get-outline-path` at the same time. Should we use `condition-case` to handle the error on wrong signature? @TheBB feel free to finish the patch! ;)

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -174,6 +174,9 @@ Note this have no effect in `helm-org-in-buffer-headings'."
         (save-restriction
           (widen)
           (unless parents (goto-char (point-min)))
+          ;; clear cache for new version of org-get-outline-path
+          (and (boundp 'org-outline-path-cache)
+               (setq org-outline-path-cache nil))
           (cl-loop with width = (window-width (helm-window))
                    while (funcall search-fn)
                    for beg = (point-at-bol)
@@ -189,10 +192,16 @@ Note this have no effect in `helm-org-in-buffer-headings'."
                    collect `(,(propertize
                                (if helm-org-format-outline-path
                                    (org-format-outline-path
-                                    (append (apply #'org-get-outline-path
-                                                   (unless parents
-                                                     (list t level heading)))
-                                            (list heading))
+                                    ;; org-get-outline-path changed in signature and behaviour since org's
+                                    ;; commit 105a4466971. Let's fall-back to the new version in case
+                                    ;; of wrong-number-of-arguments error.
+                                    (condition-case nil
+                                        (append (apply #'org-get-outline-path
+                                                       (unless parents
+                                                         (list t level heading)))
+                                                (list heading))
+                                      (wrong-number-of-arguments
+                                       (org-get-outline-path t t)))
                                     width file)
                                    (if file
                                        (concat file (funcall match-fn  0))


TheBB:
Hey thanks for this, I was too busy and was starting to feel bad about my promise.

I would assume condition-case is the way to go. Something like this

```elisp
(condition-case nil
    ;; call old org function
  (wrong-number-of-arguments
    ;; call new org function
    ))
```

aroig:
I force-pushed the `condition-case` thing. Now the PR is complete, and should work either with org master branch or older stable versions.

I have only tested in with org from master branch. Please make sure it works with older org-mode before merging.

helm-org.el
```diff
@@ -189,10 +191,16 @@ Note this have no effect in `helm-org-in-buffer-headings'."
...
+                                    ;; org-get-outline-path changed in signature and behaviour since org's
+                                    ;; commit 105a4466971. Let's fall-back to the new version in case
+                                    ;; of wrong-number-of-arguments error.
+                                    (condition-case nil
+                                        (append (apply #'org-get-outline-path
+                                                       (unless parents
+                                                         (list t level heading)))
+                                                (list heading))
+                                      (wrong-number-of-arguments
+                                       (apply #'org-get-outline-path (list t t))))
```
TheBB: Actually there's no reason this can't be `(org-get-outline-path t t)`

aroig:
I force-pushed again addressing a comment by @TheBB.The comment was about using `(org-get-outline-path t t)` instead of `(apply #'org-get-outline-path (list t t))`

thierryvolpiatto:
The questions is how do we have the same behavior of  `(org-get-outline-path arg1 arg2 arg3)`
and  `(org-get-outline-path)` with the new version ?

IOW Are you sure:

`(org-get-outline-path t t)` == `(org-get-outline-path arg1 arg2 arg3)`

And I don't see in the patch correspondance for `(org-get-outline-path)` i.e called with no args.

thierryvolpiatto:
So what is the same behavior of:

`(org-get-outline-path)` with new version ?

helm-org.el
```diff
@@ -174,6 +174,8 @@ Note this have no effect in `helm-org-in-buffer-headings'."
         (save-restriction
           (widen)
           (unless parents (goto-char (point-min)))
+          ;; clear cache for new version of org-get-outline-path
+          (setq org-outline-path-cache nil)
```
thierryvolpiatto: Please use `(and (boundp 'org-outline-path-cache) (setq ...`.

thierryvolpiatto:
So as soon I am sure we have the same behavior as before with new org I will merge.

Thanks for your work.

aroig:
It looks like this function has changed considerably. I'm not really sure how it worked before. What was the point in passing `level` and `heading` arguments? The new version removes the old three arguments, and adds two for totally unrelated things: including current headline in the output, and enabling a cache.

The patch I wrote is based on my understanding of the new function, and checking that `helm-org-in-buffer-headings` and `helm-org-parent-headings` work as expected. I may be missing something though.

aroig:
Force-pushed with the `bound-p` check for `org-outline-path-cache`.

aroig:
To be clear, as far as I understand, the new version called as `(org-get-outline-path t t)` returns a list containing the path to the current outline entry. The first `t` argument makes it include the current heading at the end, and the second argument enables using a cache in `org-outline-path-cache` which is supposed to improve performance on sequential reads of an entire file.

Does helm-org require a different behaviour from this function?

thierryvolpiatto:
Abdó Roig-Maranges <notifications@github.com> writes:

> It looks like this function has changed considerably.

Yes it is why I am asking about the new behavior.

> The patch I wrote is based on my understanding of the new function,
> and checking that helm-org-in-buffer-headings and
> helm-org-parent-headings work as expected.

I guess that if you tried the both and they are working as expected it
is all fine and we can merge.

I can't tell you more what the old version does exactly, I just made a
rewrite of the function based on the knowledge of original author.
My org knowledge is limited.

> I may be missing something though.

I have not the last org installed, for my basic org usage the default version
is enough, so I didn't check how works the new version.

Let's merge your changes and we will see what happen (crossing fingers
:-).

Thanks.

-- 
Thierry

thierryvolpiatto:
Merged thanks.

StreakyCobra:
Seems to work as expected here :+1: Thanks.

PR #1431, opened 2016-03-17
Debian and Ubuntu installation instructions

spwhitton:
This ELPA package is now available from the Debian and Ubuntu package repositories.


thierryvolpiatto:
Thanks.

PR #1434, opened 2016-03-18
Do not modify kill ring when switching to eshell

maio:
Also simplify how actual cd is done.


thierryvolpiatto:
Works great, I was wondering why I was losting the head of kill-ring in eshell but didn't
had the time to investigate, glad you fixed it, thanks.

PR #1439, opened 2016-03-21
Org tag completion improvements

alphapapa:
Thanks to [Anders Johansson](https://groups.google.com/d/msg/emacs-helm/tA6cn6TUdRY/G1S3TIdzBwAJ), here's a fix for Org tag completion in Helm.  It allows completion of multiple tags, whereas the current code only allows completing one tag at a time, requiring multiple invocations of `org-set-tags-command` (`C-c C-q`).  I have been wanting to fix this for years, but Mr. Johansson here finally figured it out for us.  :)

Now, you can press `C-c C-q`, begin typing a tag, hit `RET`, then begin typing another tag, hit `RET`, then hit `RET` again, and both tags will be set on the current heading.

However, while preparing this for Helm upstream, I came across a very strange problem: when the new function used in `helm-completing-read-handlers-alist` is called, it fails with an error saying, `apply: Wrong number of arguments...`.  I nearly pulled my hair out trying to understand it, and then I happened upon the strangest workaround: if the function's name begins with `helm`, it fails, but if it starts with anything else, it works fine.  (The function originally started with `aj/`, and I removed that prefix.)  I'm afraid my elisp-fu is too weak to figure this out.

The first commit has the bug, while the second and third commits in this pull request work properly but have the function renamed to `not-helm-org-completing-read-tags`.

Thanks for your help.

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -273,6 +273,61 @@ current heading."
         :truncate-lines helm-org-truncate-lines
         :buffer "*helm org capture templates*"))
 
+;;; Org tag completion
+
+;; Based on code from Anders Johansson posted on 3 Mar 2016 at
+;; <https://groups.google.com/d/msg/emacs-helm/tA6cn6TUdRY/G1S3TIdzBwAJ>
+
+(defun helm-org-completing-read-tags (prompt coll pred req initial hist def inh)
+  (let* ((initial (and (stringp initial)
+                       (not (string= initial ""))
+                       initial))
+         (curr (when initial
+                 (org-split-string initial ":")))
+         (table (org-uniquify (mapcar 'car org-last-tags-completion-table)))
+         (table (if curr
+                    ;; Remove current tags from list
+                    (cl-delete-if (lambda (x)
+                                    (member x curr))
+                                  table)
+                  table))
+         (prompt (if initial
+                     (concat "Tags " initial)
+                   prompt)))
+    (concat initial
+            (mapconcat 'identity
+                       (nreverse (helm-org-completing-read-multiple
+                                  prompt table pred nil nil hist def
+                                  t "Org tags" "*Helm org tags*" ":"))
+                       ":"))))
+
+(defun helm-org-completing-read-multiple (prompt choices
+                                                 &optional predicate require-match initial-input hist def
+                                                 inherit-input-method name buffer sentinel)
+  "Read multiple items with `helm-completing-read-default-1'. Reading stops
+when the user enters SENTINEL. By default, SENTINEL is
+\"*done*\". SENTINEL is disambiguated with clashing completions
+by appending _ to SENTINEL until it becomes unique. So if there
+are multiple values that look like SENTINEL, the one with the
+most _ at the end is the actual sentinel value. See
+documentation for `ido-completing-read' for details on the
+other parameters."
+  (let ((sentinel (or sentinel "*done*"))
+        this-choice res done-reading)
+    ;; Uniquify the SENTINEL value
+    (while (cl-find sentinel choices)
+      (setq sentinel (concat sentinel "_")))
+    (setq choices (cons sentinel choices))
+    ;; Read some choices
+    (while (not done-reading)
+      (setq this-choice (helm-completing-read-default-1 prompt choices
+                                                        predicate require-match initial-input hist def
+                                                        inherit-input-method name buffer nil t))
+      (if (equal this-choice sentinel)
+          (setq done-reading t)
+        (setq res (cons this-choice res))
+        (setq prompt (concat prompt this-choice ":"))))
+    res))
 
 (provide 'helm-org)
 
--- a/helm-org.el
+++ b/helm-org.el
@@ -278,7 +278,10 @@ current heading."
 ;; Based on code from Anders Johansson posted on 3 Mar 2016 at
 ;; <https://groups.google.com/d/msg/emacs-helm/tA6cn6TUdRY/G1S3TIdzBwAJ>
 
-(defun helm-org-completing-read-tags (prompt coll pred req initial hist def inh)
+;; FIXME: This function fails with a "Wrong number of arguments" error
+;; if its name starts with "helm"?!  If its name starts with anything
+;; else, it works fine!
+(defun not-helm-org-completing-read-tags (prompt coll pred req initial hist def inh)
   (let* ((initial (and (stringp initial)
                        (not (string= initial ""))
                        initial))
--- a/helm-org.el
+++ b/helm-org.el
@@ -282,27 +282,36 @@ current heading."
 ;; if its name starts with "helm"?!  If its name starts with anything
 ;; else, it works fine!
 (defun not-helm-org-completing-read-tags (prompt coll pred req initial hist def inh)
-  (let* ((initial (and (stringp initial)
-                       (not (string= initial ""))
-                       initial))
-         (curr (when initial
-                 (org-split-string initial ":")))
-         (table (org-uniquify (mapcar 'car org-last-tags-completion-table)))
-         (table (if curr
-                    ;; Remove current tags from list
-                    (cl-delete-if (lambda (x)
-                                    (member x curr))
-                                  table)
-                  table))
-         (prompt (if initial
-                     (concat "Tags " initial)
-                   prompt)))
-    (concat initial
-            (mapconcat 'identity
-                       (nreverse (helm-org-completing-read-multiple
-                                  prompt table pred nil nil hist def
-                                  t "Org tags" "*Helm org tags*" ":"))
-                       ":"))))
+  (if (not (string= "Tags: " prompt))
+      ;; Not a tags prompt.  Use normal completion by calling
+      ;; `org-icompleting-read' again without this function in
+      ;; `helm-completing-read-handlers-alist'
+      (let ((helm-completing-read-handlers-alist (rassq-delete-all
+                                                  'not-helm-org-completing-read-tags
+                                                  helm-completing-read-handlers-alist)))
+        (org-icompleting-read prompt coll pred req initial hist def inh))
+    ;; Tags prompt
+    (let* ((initial (and (stringp initial)
+                         (not (string= initial ""))
+                         initial))
+           (curr (when initial
+                   (org-split-string initial ":")))
+           (table (org-uniquify
+                   (mapcar 'car org-last-tags-completion-table)))
+           (table (if curr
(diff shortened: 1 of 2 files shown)


thierryvolpiatto:
I made some fixes.  You will find them in exp branch.
Now instead of entering `sentinel` i.e ":", just enter empty string to exit.

thierryvolpiatto:
Thanks, merged.

helm-org.el
```diff
@@ -282,27 +282,36 @@ current heading."
...
-         (prompt (if initial
-                     (concat "Tags " initial)
-                   prompt)))
-    (concat initial
-            (mapconcat 'identity
-                       (nreverse (helm-org-completing-read-multiple
-                                  prompt table pred nil nil hist def
-                                  t "Org tags" "*Helm org tags*" ":"))
-                       ":"))))
+  (if (not (string= "Tags: " prompt))
```
thierryvolpiatto: In which case is this happening ?

alphapapa: That happens when, for example, entering a heading in an org-capture prompt.  Without this change, it would try to complete tags when entering a heading.

thierryvolpiatto: alphapapa <notifications@github.com> writes:

> That happens when, for example, entering a heading in an org-capture
> prompt.

Hmm, I have no prompt in org-capture, perhaps because I am using
org-capture-templates ?

> Without this change, it would try to complete tags when entering a
> heading.

Ok.

Thanks.

-- 
Thierry

alphapapa: > Hmm, I have no prompt in org-capture, perhaps because I am using org-capture-templates ?

Depending on the template, yes.  For example, this template prompts for a heading ("Description"):

```el
("i" "Inbox: general entry" entry
  (file "")
  "* %^{Description} %^G\n%U %?")
```

thierryvolpiatto: alphapapa <notifications@github.com> writes:

> Depending on the template, yes. For example, this template prompts for a heading ("Description"):
>
> ("i" "Inbox: general entry" entry
>   (file "")
>   "* %^{Description} %^G\n%U %?")

Hmm, I am indeed prompted, but in a helm completion, I couldn't succeed
at being prompted with the org-icompleting-read stuff we have in
helm-org-completing-read-tags, I am probably missing something.

Thanks.

-- 
Thierry

helm-org.el
```diff
@@ -273,6 +273,73 @@ current heading."
...
+
+;; FIXME: This function fails with a "Wrong number of arguments" error
+;; if its name starts with "helm"?!  If its name starts with anything
+;; else, it works fine!
+(defun not-helm-org-completing-read-tags (prompt coll pred req initial hist def inh)
+  (if (not (string= "Tags: " prompt))
+      ;; Not a tags prompt.  Use normal completion by calling
+      ;; `org-icompleting-read' again without this function in
+      ;; `helm-completing-read-handlers-alist'
+      (let ((helm-completing-read-handlers-alist (rassq-delete-all
```
vmfhrmfoaj: Sorry, I'm not fluent in English.

`rassq-delete-all` function can modify `helm-completing-read-handlers-alist`.
I think commiter have intended temporary changing `helm-completing-read-handlers-alist` in `let` binding.

> https://www.gnu.org/software/emacs/manual/html_node/elisp/Association-Lists.html
> — Function: rassq-delete-all value alist
> This function deletes from alist all the elements whose cdr is eq to value. It returns the shortened alist, and **often modifies the original list structure of alist**. rassq-delete-all is like assq-delete-all except that it compares the cdr of each alist association instead of the car.

1. `org-set-tags` (working) 
2. `org-do-capture` (`helm-completing-read-handlers-alist` changed) 
3. `org-set-tags` (not working)

I tested as following:
```emacs-lisp
(advice-add #'helm-org-completing-read-tags
            :around (lambda (f &rest args)
                      (cl-flet ((rassq-delete-all
                                 (val alist)
                                 (--remove (eq (cdr it) val) alist)))
                        (apply f args))))
```

vmfhrmfoaj: or
```emacs-lisp
(rassq-delete-all
              'helm-org-completing-read-tags
              (copy-alist helm-completing-read-handlers-alist))
```

alphapapa: Hi there, that's a great catch, sorry about that.  

It's interesting that it says that it "often modifies" the original.  I haven't noticed any problems in the...14 months since this commit, and I guess no one else has either.

Which version of Emacs, Org, and Helm are you using?  I wonder if something changed to cause it to modify the alist more "often".

vmfhrmfoaj: I using  [Emacs-mac](https://bitbucket.org/mituharu/emacs-mac) with [Spacemacs](https://github.com/syl20bnr/spacemacs).


**(emacs-version)**:
GNU Emacs 25.2.1 (x86_64-apple-darwin16.6.0, Carbon Version 157 AppKit 1504.83)

**org-version**:
9.0.7

**helm**:
Latest; I update packages daily.
(current: elpa/helm-20170520.1308)

---
`rassq-delete-al` in My Emacs:
```emacs-lisp
(defun rassq-delete-all (value alist)
  "Delete from ALIST all elements whose cdr is `eq' to VALUE.
Return the modified alist.
Elements of ALIST that are not conses are ignored."
  (while (and (consp (car alist))
              (eq (cdr (car alist)) value))
    (setq alist (cdr alist)))
  (let ((tail alist) tail-cdr)
    (while (setq tail-cdr (cdr tail))
      (if (and (consp (car tail-cdr))
               (eq (cdr (car tail-cdr)) value))
          (setcdr tail (cdr tail-cdr))
        (setq tail tail-cdr))))
  alist)
```

vmfhrmfoaj: Sorry, I'm not fluent in English.

I digging it.
In my case, `setcdr` update the `cons` without returning a new `cons`.

`rassq-delete-all`:
```emacs-lisp
(defun rassq-delete-all (value alist)
  "Delete from ALIST all elements whose cdr is `eq' to VALUE.
Return the modified alist.
Elements of ALIST that are not conses are ignored."
  (while (and (consp (car alist))
              (eq (cdr (car alist)) value))
    (setq alist (cdr alist)))
   (let ((tail alist) tail-cdr)
    (while (setq tail-cdr (cdr tail))
      (if (and (consp (car tail-cdr))
               (eq (cdr (car tail-cdr)) value))
          (setcdr tail (cdr tail-cdr))
        (setq tail tail-cdr))))
  alist)
```

test `setcdr`:
```emacs-lisp
(let ((al '((a . A) (b . B) (c . C))))
  (setcdr (car al) 'a)
  al)
;;=> ((a . a) (b . B) (c . C))
```

`setcdr`:
```c
DEFUN ("setcdr", Fsetcdr, Ssetcdr, 2, 2, 0,
       doc: /* Set the cdr of CELL to be NEWCDR.  Returns NEWCDR.  */)
  (register Lisp_Object cell, Lisp_Object newcdr)
{
  CHECK_CONS (cell);
  CHECK_IMPURE (cell, XCONS (cell));
  XSETCDR (cell, newcdr);
  return newcdr;
}
```

`XSETCDR`:
```c
INLINE void
XSETCDR (Lisp_Object c, Lisp_Object n)
{
  *xcdr_addr (c) = n; // USE POINTER!!
}
```

thierryvolpiatto:
Please can someone explain what's going wrong here, I understand nothing.
Thanks.

Also:
> I using Emacs-mac with Spacemacs.
When reporting bugs don't use Spacemacs but ./emacs-helm.sh.
Spacemacs is too much hacked to reproduce reliably bugs.

vmfhrmfoaj:
The problem is `helm-completing-read-handlers-alist` is changed after `helm-org-completing-read-tags` is called.

I tested it using `emacs-helm.sh` and it was reproduced.

```emacs-lisp
;; Make sure `helm-org-completing-read-tags` is registered.
(assoc 'org-set-tags helm-completing-read-handlers-alist)
;; => (org-set-tags . helm-org-completing-read-tags)

;; do `org-capture` with the following settings.
(setq org-capture-templates `(("t" "Todo" entry (file+headline ,(concat default-directory "task.org") "Tasks") "* TODO %^{Task}")))

;; In my case, `helm-org-completing-read-tags` has been modified.
(assoc 'org-set-tags helm-completing-read-handlers-alist)
;; => nil
```

thierryvolpiatto:
Jinseop Kim <notifications@github.com> writes:

> The problem is helm-completing-read-handlers-alist is changed after helm-org-completing-read-tags is called.
>
> I tested it using emacs-helm.sh and it was reproduced.
>
> ;; Make sure `helm-org-completing-read-tags` is registered.
> (assoc 'org-set-tags helm-completing-read-handlers-alist)
> ;; => (org-set-tags . helm-org-completing-read-tags)
>
> ;; do `org-capture` with the following settings.
> (setq org-capture-templates `(("t" "Todo" entry (file+headline ,(concat default-directory "task.org") "Tasks") "* TODO %^{Task}")))
>
> ;; In my case, `helm-org-completing-read-tags` has been modified.
> (assoc 'org-set-tags helm-completing-read-handlers-alist)
> ;; => nil

Thanks for clarifications, is the following patch prevents
`helm-completing-read-handlers-alist` to be modified when calling
`org-capture` ?

diff --git a/helm-org.el b/helm-org.el
index d4e3e2d0..8dde1c2c 100644
--- a/helm-org.el
+++ b/helm-org.el
@@ -325,7 +325,7 @@ current heading."
       (let ((helm-completing-read-handlers-alist
              (rassq-delete-all
               'helm-org-completing-read-tags
-              helm-completing-read-handlers-alist)))
+              (copy-alist helm-completing-read-handlers-alist))))
         (org-icompleting-read
          prompt collection pred req initial hist def inherit-input-method))
     ;; Tags prompt

-- 
Thierry

vmfhrmfoaj:
Yes

thierryvolpiatto:
@alphapapa  `rassq-delete-all` is a destructive function, you have to use it on a copy of alist otherwise it alter the original one:

```elisp
(defvar foo '((a . 1) (b . 2) (c . 3)))
(let ((foo (rassq-delete-all 2 foo)))
  foo)
=>((a . 1)
   (c . 3))
foo
=>((a . 1)
   (c . 3))

(let ((foo (rassq-delete-all 2 (copy-alist foo))))
  foo)
=>((a . 1)
   (c . 3))
foo
=>((a . 1)
   (b . 2)
   (c . 3))
```

thierryvolpiatto:
@vmfhrmfoaj Ok, so I installed the patch, thanks to catch this.

Issue #1448, opened 2016-03-26
Common functionality in `helm-M-x` and `helm-apropos` for commands


xiongtx:
## Request

It would be nice have access to the same actions in `helm-M-x` as in `helm-apropos`, such as:

- Find function
- Info lookup

I often use `helm-M-x` to discover commands for new packages; once I find a command, I can use persistent action to get its documentation, but I cannot jump to source. I then have to use `helm-apropos` to search for the command again.

Conversely, I often want to execute a command I find through `helm-apropos`, but no action is available to do that. `helm-apropos` also does not show the keybindings of commands next to their names like `helm-M-x` does.

## Suggested Solution

Define a Helm source for Emacs commands that can be used by both `helm-M-x` and `helm-apropos`. This can reduce code duplication and increase consistency.

## Potential Issues

Looking at the `helm-command.el` code, I see that the candidates are populated through `helm-M-x-read-extended-command`, which uses `helm-comp-read` instead of creating a Helm source.

I don't know whether this is because a normal Helm source is insufficient, or the `helm-command.el` code is old, or of some other reason. `helm-comp-read` does not allow an `:action` keyword, so we cannot add actions to `helm-M-x` easily.

thierryvolpiatto:
Tianxiang Xiong <notifications@github.com> writes:

> I often use helm-M-x to discover commands for new packages; once I
> find a command, I can use persistent action to get its documentation,
> but I cannot jump to source.  I then have to use helm-apropos to
> search for the command again.

Sometimes helm users ask me to write a new feature that save them one or
two keystrokes to achieve what they want and would requires me to
rewrite a whole piece of code or more.
So to save 1 or 2 seconds to the user it would take me hours of work for
really a few benefit for the user and zero for me.

Thanks to understand, that I am not paid for helm maintainance and I do
that only on my free time, but your suggestions and bug reports are
always welcome and I appreciate your participation.

-- 
Thierry

thierryvolpiatto:
However, I can add a `:actions` keyword to `helm-comp-read` which allow adding more actions to `helm-M-x` without too much work from me (should be easy).
So you could have an action to switch to `helm-apropos` from `helm-M-x` .

xiongtx:
> Sometimes helm users ask me to write a new feature that save them one or two keystrokes to achieve what they want and would requires me to rewrite a whole piece of code or more.
>
>So to save 1 or 2 seconds to the user it would take me hours of work for really a few benefit for the user and zero for me.

[...]

PR #1449, opened 2016-03-26
Add :actions keyword to helm-comp-read and use it in helm-M-x (#1448).

thierryvolpiatto:
* helm-command.el (helm-M-x-read-extended-command): Add actions with new keyword.
* helm-elisp.el (helm-apropos): Allow passing default arg for non--interactive calls.
* helm-mode.el (helm-comp-read): Add actions keyword.


thierryvolpiatto:
Just a first shot quickly done to provide an extra action to `helm-M-x`.

PR #1452, opened 2016-03-27
Fix #1451 by storing external commands with paths.

### LINKED ISSUE #1451: `helm-open-file-externally` does not work the first time
## Problem

The first time that `helm-open-file-externally` is run, it returns the external program to be used without actually opening the file.

## Analysis

Starting from `emacs-helm.sh`, evaluating the following on a Fedora system gives:

```emacs-lisp
(helm-open-file-externally "foo/bar/doc.pdf")

;; => ("/usr/bin/xdg-open")
```

Evaluating it a second time opens the file and gives the correct result:

```emacs-lisp
;; => /usr/bin/xdg-open /foo/bar/doc.pdf process...Finished
```

The problem line seems to be:

```emacs-lisp
(helm-run-or-raise program file)
```

Calling this this function the first time results `("/usr/bin/xdg-open")`.

**Update**: the issue seems to be with `helm-external-commands-list`. See [comment below](https://github.com/emacs-helm/helm/issues/1451#issuecomment-202013343).

**Update 2**: the issue the use of a path-qualified command, e.g. `/usr/bin/xdg-open`, vs. a non-qualified command (e.g. `xdg-open`). See [comment below](https://github.com/emacs-helm/helm/issues/1451#issuecomment-202125876).

xiongtx:
Now, it no longer requires two attempts to open a file of a particular
type externally.

`helm-external-commands-list` now stores commands with fully qualified
paths (e.g. `/usr/bin/xdg-open` instead of `xdg-open`). The only
exception to this is user-invoked commands (through interactive
`helm-run-external-command`). For a good user experience, we should not
qualify the commands the user sees with paths.

Various functions (e.g. `helm-run-and-raise`) no longer require the
command to be a format string (e.g. `/usr/bin/xdg-open %s`). They now
take the command by itself (e.g. `/usr/bin/xdg-open`). This eliminates
several repetitive calls to `replace-regexp-in-string`. This should not
cause any problems, because we were already assuming that the commands
are in the form `COMMAND %s`.

### DIFF
--- a/helm-external.el
+++ b/helm-external.el
@@ -24,7 +24,7 @@
 
 
 (defgroup helm-external nil
-  "External related Applications and libraries for Helm." 
+  "External related Applications and libraries for Helm."
   :group 'helm)
 
 (defcustom helm-raise-command nil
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:

> Now, it no longer requires two attempts to open a file of a particular
> type externally.
>
> helm-external-commands-list now stores commands with fully qualified
> paths (e.g. /usr/bin/xdg-open instead of xdg-open). The only
> exception to this is user-invoked commands (through interactive
> helm-run-external-command). For a good user experience, we should not
> qualify the commands the user sees with paths.
>
> Various functions (e.g. helm-run-and-raise) no longer require the
> command to be a format string (e.g. /usr/bin/xdg-open %s). They now
> take the command by itself (e.g. /usr/bin/xdg-open). This eliminates
> several repetitive calls to replace-regexp-in-string. This should not
> cause any problems, because we were already assuming that the commands
> are in the form COMMAND %s.

Thanks for your PR, but you didn't answer my questions:

What is your PR trying to fix ?

IIUC the only problem you have to fix is that you try to use
helm-open-file-externally from elisp and it is not working because helm
find the full path of a default program in your .mailcap, correct me if
I am wrong.

If I am right we just have to ensure we return the basename of this
program (xdg-open) from .mailcap file.

Actually your PR is not working as expected, returning a list of
full-paths and we endup with a mix of full-path and basename in both
history and completion list.
Also on the interactive command you have to finally remove-dups which
looks wrong.

Generally I don't want to make such changes on helm stuff that work
correctly since years now, perhaps a refactorization is needed but
no more IMO.

I will make a change that ensure the returned patch from a mailcap is a
basename, this is the only change needed IMO for now (apart some
refactorization not affecting the current behavior).

Thanks.

-- 
Thierry

xiongtx:
Yes, I think your change works, although I agree the code could use some refactoring. 

The issue of mixing path-qualified and non-qualified commands paths already existed prior to this change.

I'll close this then.

PR #1455, opened 2016-03-30
Ag types

thierryvolpiatto:


thierryvolpiatto:
Fixme: handle ag versions not handling types (just do nothing when types returns nil).

PR #1458, opened 2016-04-01
Allow multiple ag types

xuchunyang:


thierryvolpiatto:
Thanks, but this is already implemented, the `mapconcat` happen in `helm-grep-ag-prepare-cmd-line`.
Isn't it working for you ?

xuchunyang:
Oops, it is working.  I was using some part of helm from melpa and some part from git, don't know how this happened.

PR #1464, opened 2016-04-08
Overhaul helm-fuzzy-default-highlight-match.

CyberShadow:
This is a revision of the `helm-fuzzy-default-highlight-match` function, with the following fixes:

- Highlights all matches, not just the first.
- Fixes sources' `match-part` property being ignored when falling back to multi-word or per-character highlighting.
- Fixes interpreting individual characters as individual regular expressions when falling back to per-character highlighting.
- Adds comments.

Screenshot:

Note that because of the `match-part` fix, a visual "regression" is introduced:

This is because the sources return a `match-part` that does not match the part of the string they actually match. Of course, it does not affect the result list, only highlighting. Still, I think it is an improvement, especially if `match-part` will be fixed accordingly. Ideally, though, I think that the highlighting should be done by the source itself, and the `match-part` property deprecated / removed.

I am relatively new to Emacs and unfamiliar with Helm internals, so please let me know if there is something that can be improved in this PR, and I will revise it.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -3092,7 +3092,7 @@ Returns non-nil if all PATTERNS were found at least once."
 
        ;; Otherwise, highlight occurrences of each character, in that order.
        ;; We assume that helm-pattern is not a regex in this case.
-       ((funcall matcher (mapcar 'rx-form (split-string helm-pattern "" t)))))
+       ((funcall matcher (mapcar 'regexp-quote (split-string helm-pattern "" t)))))
 
       (setq display (concat mp-prefix (buffer-string) mp-suffix)))
     (if real (cons display real) display)))
--- a/helm.el
+++ b/helm.el
@@ -3023,41 +3023,81 @@ It is meant to use with `filter-one-by-one' slot."
   (let* ((pair (and (consp candidate) candidate))
          (display (helm-stringify (if pair (car pair) candidate)))
          (real (cdr pair))
-         (regex (helm-aif (and helm-migemo-mode
-                               (assoc helm-pattern
-                                      helm-mm--previous-migemo-info))
-                    (cdr it)
-                  helm-pattern))
-         ;; FIXME This is called at each turn, cache it to optimize.
-         (mp (helm-aif (helm-attr 'match-part (helm-get-current-source))
-                 (funcall it display))))
+         start-point end-point
+         (matcher
+          (lambda (patterns)
+            "Repeatedly attempts to highlight all PATTERNS.
+PATTERNS is a list of patterns to be highlighted, in that order,
+but the matches will be repeated. On the second and successive iterations,
+if not all patterns are found, then none are highlighted.
+Returns non-nil if all PATTERNS were found at least once."
+            (let
+                ((regexps (mapcar       ; Patterns converted to regexps
+                           (lambda (p)
+                             (or (and helm-migemo-mode
+                                      (assoc
+                                       p helm-mm--previous-migemo-info))
+                                 p))
+                           patterns))
+                 found                  ; Did we find anything yet?
+                 done                   ; Are we done?
+                 matches)               ; cons list of match bounds
+
+              (goto-char start-point)
+
+              (when regexps      ; Bail on empty patterns
+                ;; Repeat until the end of the buffer
+                (while (not done)
+                  (setq matches '())
+                  (mapc (lambda (re)
+                          (if (and
+                               (re-search-forward re end-point t) ; We found something
+                               (not (equal (match-beginning 0) (match-end 0)))) ; Non-empty
+                              (push (cons (match-beginning 0) (match-end 0)) matches) ; Save result
+                            (setq done t))) ; No match? We're done
+                        regexps)
+
+                  (when (or
+                         (not found)      ; First iteration?
+                         (not done))      ; Found all patterns?
+                    ;; Apply highlights to results
+                    (mapc (lambda (m)
+                            (add-text-properties
+                             (car m) (cdr m)
+                             '(face helm-match)))
+                          matches))
+                  (when matches
+                    (setq found t)))
+                found)))))
+
     (with-temp-buffer
       (insert (propertize display 'read-only nil)) ; Fix (#1176)
-      (goto-char (point-min))
-      (when mp
-        ;; FIXME the first part of display may contain an occurrence of mp.
-        ;; e.g "helm-adaptive.el:27:(defgroup helm-adapt"
-        (search-forward mp nil t)
-        (goto-char (match-beginning 0)))
-      (if (re-search-forward regex (and mp (+ (point) (length mp))) t)
-          (add-text-properties
-           (match-beginning 0) (match-end 0) '(face helm-match))
-          (cl-loop with multi-match
-                   with patterns = (if (string-match-p " " helm-pattern)
-                                       (prog1 (split-string helm-pattern)
-                                         (setq multi-match t))
-                                       (split-string helm-pattern "" t))
-                   for p in patterns
-                   for re = (or (and helm-migemo-mode
-                                     (assoc-default
-                                      p helm-mm--previous-migemo-info))
-                                p)
-                   do
-                   (when (re-search-forward re nil t)
-                     (add-text-properties
-                      (match-beginning 0) (match-end 0)
-                      '(face helm-match)))
-                   (when multi-match (goto-char (point-min)))))
+
+      ;; FIXME This is called at each turn, cache it to optimize.
+      (helm-aif (helm-aif (helm-attr 'match-part (helm-get-current-source))
+                    (funcall it display))
+          (progn
+            ;; FIXME the first part of display may contain an occurrence of mp.
+            ;; e.g "helm-adaptive.el:27:(defgroup helm-adapt"
+            (goto-char (point-min))
+            (search-forward it nil t)
+            (setq start-point (match-beginning 0)
+                  end-point (match-end 0)))
+        (setq start-point (point-min)
+              end-point nil))
+
+      (cond
+       ;; First, try to find helm-pattern as a regex pattern as-is. Repeat if found.
+       ((funcall matcher (list helm-pattern)))
+
+       ;; Otherwise, if it contains a space, highlight all words, in any order.
+       ((string-match-p " " helm-pattern)
+        (mapc (lambda (s) (funcall matcher (list s))) (split-string helm-pattern)))
+
+       ;; Otherwise, highlight occurrences of each character, in that order.
+       ;; We assume that helm-pattern is not a regex in this case.
+       ((funcall matcher (mapc 'rx-form (split-string helm-pattern "" t)))))
+
       (setq display (buffer-string)))
     (if real (cons display real) display)))
 
--- a/helm.el
+++ b/helm.el
@@ -3092,7 +3092,7 @@ Returns non-nil if all PATTERNS were found at least once."
 
        ;; Otherwise, highlight occurrences of each character, in that order.
        ;; We assume that helm-pattern is not a regex in this case.
-       ((funcall matcher (mapc 'rx-form (split-string helm-pattern "" t)))))
+       ((funcall matcher (mapcar 'rx-form (split-string helm-pattern "" t)))))
 
       (setq display (concat mp-prefix (buffer-string) mp-suffix)))
     (if real (cons display real) display)))
--- a/helm.el
+++ b/helm.el
@@ -3034,7 +3034,7 @@ Returns non-nil if all PATTERNS were found at least once."
                 ((regexps (mapcar       ; Patterns converted to regexps
                            (lambda (p)
                              (or (and helm-migemo-mode
-                                      (assoc
+                                      (assoc-default
                                        p helm-mm--previous-migemo-info))
                                  p))
                            patterns))
--- a/helm.el
+++ b/helm.el
@@ -3017,57 +3017,37 @@ real part."
                        (< len1 len2))
                       ((> scr1 scr2)))))))))
 
+(defun helm--maybe-get-migemo-pattern (pattern)
+  (or (and helm-migemo-mode
+           (assoc-default pattern helm-mm--previous-migemo-info))
+      pattern))
+
 (defun helm-fuzzy-default-highlight-match (candidate)
   "The default function to highlight matches in fuzzy matching.
 It is meant to use with `filter-one-by-one' slot."
   (let* ((pair (and (consp candidate) candidate))
          (display (helm-stringify (if pair (car pair) candidate)))
          (real (cdr pair))
-         (matcher
-          (lambda (patterns)
-            "Repeatedly attempts to highlight all PATTERNS.
-PATTERNS is a list of patterns to be highlighted, in that order,
-but the matches will be repeated. On the second and successive iterations,
-if not all patterns are found, then none are highlighted.
-Returns non-nil if all PATTERNS were found at least once."
-            (let
-                ((regexps (mapcar       ; Patterns converted to regexps
-                           (lambda (p)
-                             (or (and helm-migemo-mode
-                                      (assoc-default
-                                       p helm-mm--previous-migemo-info))
-                                 p))
-                           patterns))
-                 found                  ; Did we find anything yet?
-                 done                   ; Are we done?
-                 matches)               ; cons list of match bounds
 
-              (goto-char (point-min))
+         (highlight-once
+          ;; Highlight first occurrence from RE in point.
+          ;; Returns nil if not found.
+          (lambda (re)
+            (when (re-search-forward re nil t)
+              (add-text-properties
+               (match-beginning 0) (match-end 0)
+               '(face helm-match)))))
 
-              (when regexps      ; Bail on empty patterns
-                ;; Repeat until the end of the buffer
-                (while (not done)
-                  (setq matches '())
-                  (mapc (lambda (re)
-                          (if (and
-                               (re-search-forward re nil t) ; We found something
-                               (not (equal (match-beginning 0) (match-end 0)))) ; Non-empty
-                              (push (cons (match-beginning 0) (match-end 0)) matches) ; Save result
-                            (setq done t))) ; No match? We're done
-                        regexps)
+         (highlight-all
+          ;; Highlight all occurrences of RE.
+          ;; Returns nil if nothing was found, t otherwise.
+          (lambda (re)
+            (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
This is a good idea to rework this, however I don't think highlighting all occurences of characters in fuzzy matching is so nice, it quickly looks confusing, but for multi-match yes it is ok, it is what we already do in (m)occur (and grep before using ansi).
I have commited some changes on exp branch, please have a look.

Thanks.

CyberShadow:
Well, I am a little sad that you chose to improve the existing function while addressing only a part of the problems addressed by my patch. It would have been much easier for me to remove the functionality you did not like from my PR.

What exactly is "fuzzy matching"? Which part of the code covers it? Is it where the string is split by individual characters?

The function as it is in the exp branch still has the problems:

- First of all, multi-highlighting still doesn't work as in my screenshot. You have not handled the simplest case, where `helm-pattern` is matched whole. This is what I wanted to fix in the first place, everything else came along the way.

- When splitting `helm-pattern` by individual characters, they are still treated as individual regular expressions.

- The new code assumes that the result of `match-part` will always be at the end of the string, is that correct? I noticed you removed the TODO item, but I don't see how the issue is fixed. If it is always at the end, then it does indeed remove the ambiguity, but the code still does not exploit it.

- Actually, the `match-part` functionality does not work for me at all, neither in `master` or `exp` ([screenshot with exp branch](http://dump.thecybershadow.net/c0ff9d09b1bdc8738c99020e504d1ccc/Spectacle.T16203.png). I do not understand why. But it does work in my patch, as you can see in my screenshot above. I do not understand why.

- The code remains undocumented, and for example `helm-migemo-mode` is still called from two places.

thierryvolpiatto:

> Well, I am a little sad that you chose to improve the existing
> function while addressing only a part of the problems addressed by my
> patch.

Not sure to know exactly what you want to fix, but as you can see I have
modified code on exp branch so that your PR is still valid on master.

> It would have been much easier for me to remove the functionality you
> did not like from my PR.

We can remove complexity of your code by starting from the one on exp branch.

> What exactly is "fuzzy matching"?

M-x shcmd match shell-command 

> Which part of the code covers it?

Well, see the code about fuzzy matching in helm and helm-lib.

> Is it where the string is split by individual characters?

For the part that is related to this PR yes.

> The function as it is in the exp branch still has the problems:
>
>   * First of all, multi-highlighting still doesn't work as in my
>     screenshot. You have not handled the simplest case, where
>     helm-pattern is matched whole. This is what I wanted to fix in the
>     first place, everything else came along the way.

I don't know what you are matching in your screenshots.
    
>   * When splitting helm-pattern by individual characters, they are
>   still treated as individual regular expressions.

Yes, this is what I expect, what's wrong with this ?

>    
>   * The new code assumes that the result of match-part will always be
>     at the end of the string, is that correct? I noticed you removed
>     the TODO item, but I don't see how the issue is fixed. If it is
>     always at the end, then it does indeed remove the ambiguity, but
>     the code still does not exploit it.

Right, I have fixed this, thanks.
    
>   * Actually, the match-part functionality does not work for me at
>     all, neither in master or exp (screenshot with exp branch. I do
>     not understand why. But it does work in my patch, as you can see
>     in my screenshot above. I do not understand why.

Please elaborate on what you are expecting from match-part.
    
>   * The code remains undocumented,

This is not the final code.

>   and for example helm-migemo-mode is still called from two places.

This is needed AFAIK.

> —


-- 
Thierry

CyberShadow:
> I don't know what you are matching in your screenshots.

Just "helm"

> Yes, this is what I expect, what's wrong with this ?

If it contains special characters, then the behavior will be less useful.

> Please elaborate on what you are expecting from match-part.

Maybe I don't understand what it's for. But if the candidate is `"/foo/bar/baz"`, and `match-part` returns `"baz"`, and `helm-pattern` is `"ba"`, then shouldn't the `ba` in `baz` be highlighted, and not the one in `bar`?

> This is needed AFAIK.

Why? I don't mean that it is called twice, but that the two calls are done from two places, so the code is copy-pasted. This is not the case in my version.

CyberShadow:
> We can remove complexity of your code by starting from the one on exp branch.

OK, I will incorporate the change with placing only `match-part` in the buffer. It will eliminate the `start-point` and `end-point` variables.

thierryvolpiatto:
Some screenshots using exp branch:

helm-locate:

screenshot - 09042016 - 15 20 35

helm-etags (the match-part is 'tag)

screenshot - 09042016 - 15 21 10

helm-M-x:

screenshot - 09042016 - 15 22 19

CyberShadow:
It is correct in your screenshots. But your screenshots only show cases where it is correct :)

> helm-locate:

Try a `helm-pattern` without a space with `helm-locate`. You will see that multiple occurrences are not highlighted. You will also see that as soon as you add a space, multiple occurrences will start being highlighted.

> helm-etags (the match-part is 'tag)

It is not visible from this screenshot whether `match-part` is working or not, because your search pattern does not contain any words that would be outside of `match-part`.

Instead, try the "Files from Current Directory" source. It has a `match-part` for just the file name (not directory). Currently, you will see that `helm-fuzzy-default-highlight-match` highlights parts of the directory name too, thus ignoring `match-part`.

> helm-M-x:

That works as expected. But I think that if there was a function, hypothetically, `shell-command-execute-shell-command`, the letters "shcmd" should be highlighted twice. But only if the letters all appear. For example, for a hypothetical `shell-command-bash`, the "sh" in "bash" should not be highlighted again, because it is an incomplete match group. This is what my implementation does.

CyberShadow:
> OK, I will incorporate the change with placing only match-part in the buffer. It will eliminate the start-point and end-point variables.

OK, done. As a bonus it also searches for `match-part` backwards, thus eliminating the problem where the file name has the same name as a directory component. It also behaves correctly if `match-part` for some reason is not found in the candidate.

Please tell me what else needs to be improved to make this PR mergeable.

CyberShadow:
@thierryvolpiatto I see it is not customary in your project (or elisp in general?) to use local functions. Now that it no longer refers to `helm-fuzzy-default-highlight-match`'s closure, I could extract `matcher` to a global function if you like.

thierryvolpiatto:

> Try a helm-pattern without a space with helm-locate. You will see that
> multiple occurrences are not highlighted. You will also see that as
> soon as you add a space, multiple occurrences will start being
> highlighted.

Indeed, fixed.

>     helm-etags (the match-part is 'tag)
>    
> It is not visible from this screenshot whether match-part is working
> or not, because your search pattern does not contain any words that
> would be outside of match-part.

Yes it is working, we have "helm ada" in both sides.

> Instead, try the "Files from Current Directory" source. It has a
> match-part for just the file name (not directory). Currently, you will
> see that helm-fuzzy-default-highlight-match highlights parts of the
> directory name too, thus ignoring match-part.

You misunderstand how things are working here match-part match only
basename when display show only basename, otherwise match all.

>     helm-M-x:
>    
> That works as expected. But I think that if there was a function,
> hypothetically, shell-command-execute-shell-command, the letters
> "shcmd" should be highlighted twice.

This is what your code does, not mine until to now, however I think
finally it is not too bad to show all.

> —


-- 
Thierry

thierryvolpiatto:

> @thierryvolpiatto I see it is not customary in your project (or elisp
> in general?) to use local functions. Now that it no longer refers to
> helm-fuzzy-default-highlight-match's closure, I could extract matcher
> to a global function if you like.

No there is no closure here, only a global variable bound to a function.
Don't know what you mean here.

-- 
Thierry

CyberShadow:
> You misunderstand how things are working here match-part match only basename when display show only basename, otherwise match all.

Sorry, I don't understand. Could you please elaborate?

> No there is no closure here, only a global variable bound to a function. Don't know what you mean here.

I meant the `start-point` and `end-point` variables. Sorry if I'm using the wrong terminology, I mean "local variable" as specified in the [Emacs manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Local-Variables.html). These variables are defined using `let*`, which is why I said they're local. You're right that there is no closure here.

Thank you for your improvements, but I don't understand why you would rather make only the smallest necessary changes to your code when my PR also contain other fixes. Could you please answer my question ("Please tell me what else needs to be improved to make this PR mergeable"), because at the moment I feel like I am wasting my time and that my contributions to this project are unwelcome.

thierryvolpiatto:

> Please tell me what else needs to be improved to make this PR
> mergeable.

There is too much code actually, see the exp branch which behave the
same as your branch.
Too much use of mapc for my taste, also you use rx where it is not
needed (in fuzzy no regexp is used).
Too much complexity with the `while (not done) [...]` for accumulating
candidates.

Also there is an occurence of mapc which is wrong (use mapcar instead).

By using the cl-loop (we use it widely in helm) you could simplify your
code.

Otherwise your code is working fine and I appreciate your effort.

Thanks.

> —


-- 
Thierry

CyberShadow:
> There is too much code actually, see the exp branch which behave the same as your branch.

Well, there are minor issues that are still present in the `exp` branch. But I guess that since the issue I came here for is fixed, I have no reason to feel strongly about it.

> Too much use of mapc for my taste

Sorry, what is wrong with mapc? I have noticed that `mapc` is an Emacs Lisp builtin and `cl-loop` is not, therefore `mapc` should perform better, thus should be preferred unless `cl-loop` can simplify the code greatly.

> also you use rx where it is not needed (in fuzzy no regexp is used).

It is only used to escape special characters. `rx` is already used somewhere else in `helm` for this purpose, I think.

> Too much complexity with the `while (not done) [...]` for accumulating candidates.

It is the shortest code I could come up with while at the same time remaining correct (for my understanding of what is correct behavior). For example, your code in the `exp` branch will fail the `shell-command-bash` test I have proposed. If you do not care about this case, then I agree that it can be made much simpler.

> Otherwise your code is working fine and I appreciate your effort.

Thanks. Is it realistic to merge this PR with more communication and work from my side?

CyberShadow:
> Also there is an occurence of mapc which is wrong (use mapcar instead).

Well spotted. Fixed.

thierryvolpiatto:

> Sorry, what is wrong with mapc? I have noticed that mapc is an Emacs
> Lisp builtin and cl-loop is not, therefore mapc should perform better,

It doesn't, when compiled loop is very fast.
Please let me choose what I feel the best for helm, I want to be able to
review code easily and do not want archaic code like you may find in
emacs itself.

> thus should be preferred unless cl-loop can simplify the code greatly.
>
>     also you use rx where it is not needed (in fuzzy no regexp is used).
>    
> It is only used to escape special characters.

You don't need to escape charecters here, if you want to do so use regexp-quote.

> rx is already used somewhere else in helm for this purpose, I think.

Not for this purpose.

> It is the shortest code I could come up with while at the same time remaining correct (for my understanding of what is correct behavior). For example, your code in
> the exp branch will fail the shell-command-bash test I have proposed.

It doesn't, yours yes.


-- 
Thierry

CyberShadow:
> Please let me choose what I feel the best for helm, I want to be able to review code easily and do not want archaic code like you may find in emacs itself.

OK. So should I rewrite all uses of `mapc` and `mapcar` into `cl-loop` or something else?

> You don't need to escape charecters here,

I will agree that it is not important, but here is why it is "needed":

```
mkdir /tmp/test
touch '/tmp/test/abc^def'
emacs '/tmp/test/abc^def'
M-x helm-for-files
```

Then type `a^d`.

In the `exp` branch, the `^` will not be highlighted.

It is a very minor bug, but it is a bug.

> if you want to do so use regexp-quote.

Thanks, I forgot that function's name. Done.

> It doesn't, 

I don't understand what you mean by this. If you mean that the test "succeeds" on `exp`, well, only because you reverted the feature entirely :) Now the `shell-command-execute-shell-command` test I proposed fails.

> yours yes.

I don't understand what you meant by this either, but in my branch both tests pass as I have proposed them.

thierryvolpiatto:

>     Please let me choose what I feel the best for helm, I want to be able to review code easily and do not want archaic code like you may find in emacs itself.
>    
> OK. So should I rewrite all uses of mapc and mapcar into cl-loop or something else?

If you really want to use a classic while loop use something like this
instead:

    (while (and regexps (re-search-forward (car regexps) nil t))
      (add-text-properties
       (match-beginning 0) (match-end 0)
       '(face helm-match))
      (setq regexps (cdr regexps)))

which avoid all the complexity of your flags done, matches etc...

>     You don't need to escape charecters here,
>    
> I will agree that it is not important, but here is why it is "needed":
>
> mkdir /tmp/test
> touch '/tmp/test/abc^def'
> emacs '/tmp/test/abc^def'
> M-x helm-for-files
>
> Then type a^d.
>
> In the exp branch, the ^ will not be highlighted.
>
> It is a very minor bug, but it is a bug.

Ok.

>     if you want to do so use regexp-quote.
>    
> Thanks, I forgot that function's name. Done.

Thanks it is the one to use.

> I don't understand what you mean by this. If you mean that the test
> "succeeds" on exp, well, only because you reverted the feature
> entirely :) Now the shell-command-execute-shell-command test I
> proposed fails.

I want that on shell-command-execute-shell-command only 

shell-command-execute-shell-command
^^    ^ ^   ^

highlight and not

shell-command-execute-shell-command
^^    ^ ^^  ^         ^^    ^ ^^  ^

when using as pattern "shcmd"

> —
> You are receiving this because you were mentioned.


-- 
Thierry

CyberShadow:
> If you really want to use a classic while loop

I have no preference. The code was so long mainly to support repeated fuzzy search correctly.

> I want that on shell-command-execute-shell-command only

OK, done. This allowed simplifying the code greatly. The code is now ever shorter than the `exp` variant, if you were to remove comments and blank lines.

I still left in some uses of `mapc` / `mapcar`, because the alternative would have resulted in longer code, but let me know if you don't like them and I will rewrite it.

Sorry, I still don't understand the deal with `match-part` (you haven't answered my question). I *think* it works correctly now, but I did not understand your explanation for why it does not.

thierryvolpiatto:

> OK, done. This allowed simplifying the code greatly. The code is now
> ever shorter than the exp variant, if you were to remove comments and
> blank lines.

You still use lambda's with not so nice code (well from my point of
view).

> I still left in some uses of mapc / mapcar, because the alternative
> would have resulted in longer code, but let me know if you don't like
> them and I will rewrite it.

That's ok for now, I will probably switch to the cl-loop version though
which is simpler.

> Sorry, I still don't understand the deal with match-part (you haven't
> answered my question).

See documentation of `match-part` slot in `helm-source`

e.g in helm-etags with a pattern like "helm ada"

you want to match only:

helm-adaptive.el:23:(defun helm-adaptive-[...]
                           ^^^^ ^^^

and not

helm-adaptive.el:23:(defun helm-adaptive-[...]
^^^^ ^^^                   ^^^^ ^^^

so with your code you actually remove the non matching parts:

"helm-adaptive.el:23:" and nothing at end

then in your temp buffer highlight as needed:

"(defun helm-adaptive-[...]"

and finally when done

reconstruct the whole candidate with the highlighted part:

"helm-adaptive.el:23:" + "(defun helm-adaptive-[...]" + ""

> I think it works correctly now, but I did not understand your
> explanation for why it does not.

It is working yes, but why inserting the whole text and then extracting
it instead of just replace-matching when done like in exp branch ?

i.e here:
    (when (search-backward it nil t)
       (setq mp-suffix (delete-and-extract-region (match-end 0) (point-max))
             mp-prefix (delete-and-extract-region (point-min) (match-beginning 0))))))
     [...]
     (setq display (concat mp-prefix (buffer-string) mp-suffix))

But your code is correct anyway, it is a detail.

Thanks, merging now.

> —
> You are receiving this because you were mentioned.


-- 
Thierry

CyberShadow:
> It is working yes, but why inserting the whole text and then extracting it instead of just replace-matching when done like in exp branch ?

Oh, OK. The only reason I did that is to search backwards. `string-match` did not seem to have an option or variant for backwards search. Searching backwards is needed when the match-part (base name) occurs earlier in the string. I understand that usually it is located at the end or near the end, so I thought it makes sense to search backwards. This fixes the problem that a comment in the original code was referring to:

```
          ;; FIXME the first part of display may contain an occurrence of mp.
          ;; e.g "helm-adaptive.el:27:(defgroup helm-adapt"
```

> Thanks, merging now.

Thank you :) Sorry for wasting so much of your time over trivialities.

CyberShadow:
Sorry, but why did you merge this PR if you were going to overwrite the code with your version anyway? I thought you only wanted to refactor the loops...

Your code still has problems:

- It does not call `regexp-quote` as per your own recommendation.
- Searching for `match-part` from the end of the string is gone, so the highlighting will be wrong again when `match-part` occurs earlier in the string.
- The `match-part` fix for `match-part` being the basename in `helm-recentf` etc. from my version is gone again, though I still can't explain it.

By merging the `exp` branch you have effectively wasted both of our time and introduced three regressions... Sorry, but I don't know what I should think of this.

I have proposed to improve my code to your liking, just say the word.

thierryvolpiatto:

> Your code still has problems:

All the issues you mention have been fixed.

> By merging the exp branch you have effectively wasted both of our time
> and introduced three regressions... Sorry, but I don't know what I
> should think of this.

We didn't waste our time, it helped a lot fixing this issue with your
feedback even if prefer on the end my version which does the same as
yours.

> I have proposed to improve my code to your liking, just say the word.

There is still parts of your code where we disagree even if your code
works fine.

Thanks you anyway for your help on this.

> —
> You are receiving this because you modified the open/close state.


-- 
Thierry

CyberShadow:
OK. I don't know if you saw because you replied by email, but I edited the comment you quoted.

However you should also know that your version (in git master) is broken right now:
```
Error running timer: (wrong-type-argument integer-or-marker-p nil)
```

CyberShadow:
Here is a screenshot of the backtrace:



Sorry for not pasting text, the backtrace window is cleared as soon as I touch anything, I'm not immediately sure how to get the backtrace text.

thierryvolpiatto:

> Here is a screenshot of the backtrace:
>
> [687474]
>
> Sorry for not pasting text, the backtrace window is cleared as soon as
> I touch anything, I'm not immediately sure how to get the backtrace

Thanks, I have reverted all changes we did, I will not have the time now
to fix this.
You will find the code in highlight_matches branch, I created a PR.
I reused (partially) in this branch the code you used to remove/reinsert
the non match-part parts.
This branch is also storing match-part in a special property, so the
function is no more called at each turn when matching and highlighting.
This branch is working for me...

> —
> You are receiving this because you modified the open/close state.


-- 
Thierry

PR #1466, opened 2016-04-10
Fix loosing init face by using the deletion insertion method.

thierryvolpiatto:
* helm.el (helm-fuzzy-default-highlight-match): Do it.


CyberShadow:
Yep, this fixes the error for me :)

Issue #1467, opened 2016-04-11
`wrong-type-argument stringp nil`


CyberShadow:
I got this when I opened `helm-for-files`:

![](http://dump.thecybershadow.net/671e08c1e69e34af529bb923b35beb3a/helmbug.png)

It looks like the backtrace buffer has no `default-directory`, so calling `abbreviate-file-name` fails.

CyberShadow:
https://github.com/emacs-helm/helm/pull/1468

thierryvolpiatto:
Vladimir Panteleev <notifications@github.com> writes:

> I got this when I opened helm-for-files:
>
> [687474]
>
> It looks like the backtrace buffer has no default-directory, so calling abbreviate-file-name fails.

Why your backtrace buffer doesn't have a default-directory ?
It have always a default-directory like all emacs buffers.

> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub*


-- 
Thierry

CyberShadow:
Sorry, I don't know. It just started happening.

michael-heerdegen:
We need to be sure about the reason.  Maybe the next time `default-directory' is bound to a number, or to a string for that `abbreviate-file-name' fails...

AFAIK, default-directory should not be bound to nil normally.

CyberShadow:
> We need to be sure about the reason.

I reproduced it. It happens when something is causing Emacs to be stuck, and you send a SIGUSR2 to the emacs process. This causes a backtrace buffer to appear. The thus-created backtrace buffer seems to lack a default-directory.

michael-heerdegen:
Does this depend on this "something"?

When I send SIGUSR2 to my Emacs, the debugger buffer has `default-directory -> "~/"`.

thierryvolpiatto:
As this is very specific and more related to an emacs bug and also not reproductible, I am closing, feel free to reopen if more infos.

CyberShadow:
It hasn't happened since my last comment, so OK. I'll let you know if I find a way to reproduce it reliably.

CyberShadow:
It happened again :/ It definitely happens after I send a SIGUSR2 (I think when Emacs is waiting on a subprocess), but it only seems to happen sometimes.

The patch I proposed in #1467 is small and personally I don't see a problem with merging it.

thierryvolpiatto:
Vladimir Panteleev <notifications@github.com> writes:

[...]

PR #1468, opened 2016-04-11
Fix issue #1467 (fix handling buffers with nil default-directory)

CyberShadow:
This fixes it for me.

### DIFF
--- a/helm-buffers.el
+++ b/helm-buffers.el
@@ -324,10 +324,10 @@ See `ido-make-buffer-list' for more infos."
          (size (propertize (helm-buffer-size buf)
                            'face 'helm-buffer-size))
          (proc (get-buffer-process buf))
-         (dir (with-current-buffer buffer (abbreviate-file-name default-directory)))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Merged but IMO this should be fixed upstream.

thierryvolpiatto:
Thanks.

CyberShadow:
I agree, unfortunately as it's difficult to reproduce and analyze, it's difficult to find its source.

thierryvolpiatto:

> I agree, unfortunately as it's difficult to reproduce and analyze,
> it's difficult to find its source.

Perhaps report the bug to emacs ?

> —
> You are receiving this because you modified the open/close state.


-- 
Thierry

CyberShadow:
What use is a bug report without solid reproduction instructions? I was wrong to report it to helm in the first place, because I twice thought it could be easily reproduced, and twice I was wrong.

Considering that no one encountered it before me, it's probably caused by another package (or possibly even something in my `init.el`).

PR #1469, opened 2016-04-11
Make sure pass a string to kill-new

xuchunyang:
In buffer source, C-u C-c C-k ('helm-kill-selection-and-quit') raises a
wrong-type-argument error because the real value of candidate is not a
string.  To fix this, (format "%s" ...) is used to convert any kind of
value into a string (its printed representation).

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -5357,6 +5357,8 @@ With a prefix arg set to real value of current selection."
   (interactive "P")
   (with-helm-alive-p
     (let ((str (helm-get-selection nil (not arg))))
+      (when str
+        (setq str (format "%s" str)))
       (kill-new str)
       (helm-set-pattern str))))
 (put 'helm-yank-selection 'helm-only t)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:

> In buffer source, C-u C-c C-k ('helm-kill-selection-and-quit') raises a
> wrong-type-argument error because the real value of candidate is not a
> string. To fix this, (format "%s" ...) is used to convert any kind of
> value into a string (its printed representation).

Indeed, thanks, but we have the same problem with markers (imenu) and we
may have the same problem in others place too if the real candidate is
not a string, could you apply your change directtly to
`helm-kill-selection-and-quit` instead ?

Thanks.

> —


-- 
Thierry

thierryvolpiatto:
Ok fixed.

>> —
>> You are receiving this because you are subscribed to this thread.
>> Reply to this email directly or view it on GitHub*


-- 
Thierry

xuchunyang:
All right, just don't forget to do the same to `helm-yank-selection`.  And `helm-get-selection` can return nil but `(format "%s" nil`) gives `"nil"`, though this should be a problem in practice.

thierryvolpiatto:

> All right, just don't forget to do the same to
> helm-yank-selection. And helm-get-selection can return nil but (format
> "%s" nil) gives "nil", though this should be a problem in practice.

Indeed, done.

Thanks.

> —


-- 
Thierry

PR #1470, opened 2016-04-12
Simplify argument list in helm-get-selection

xuchunyang:
It looks like `(buffer nil buffer-s)` is not necessary to me, so I made this change.


thierryvolpiatto:
Yes, thanks.

PR #1481, opened 2016-04-23
Introduce same feature as "cp --backup=numbered from to".

thierryvolpiatto:
* helm-files.el (helm-find-files-backup): New action.
(helm-find-files-action-transformer): Add it to actions when
using helm-file-globstar and basename is a recursive wildcard.
(helm-dired-action): Add backup action.


thierryvolpiatto:
To try this feature use https://github.com/jwiegley/emacs-async/pull/62

PR #1483, opened 2016-04-25
Introducing optional input string into helm-occur

wongdao:
Making other scripts can preset the searching words like `(helm-occur "hello")`


thierryvolpiatto:
Sorry but I don't want to use input in helm-occur.

PR #1487, opened 2016-04-28
Modify helm-fuzzy-default-highlight-match for migemo

momomo5717:
Hello,

It seems fuzzy highlight doesn't work with `helm-migemo-mode` after recent update.
So I modified it for using literal patterns independently of `helm-migemo-mode`.

Thank you.


thierryvolpiatto:
Thanks.

momomo5717:
Thank you for merging!

PR #1503, opened 2016-05-16
Support truncated index lines in info manuals

Lompik:
Issue: selecting certain lines in helm-info-* takes us to the top of the info node but not to the actual selected item .

Cause: long lines in info index are truncated in info file (as per attached screenshot) and `helm-info-init` only insert lines starting with '* '. (line numbers are propertize with invisible in emacs.)

2016-05-16-18 32 04_1085x151 9 739

Fix: I propose the attached patch. I do not see any case where another two `* Menu:` would be in the same info file index so I insert the whole index at once. let me know if thats not the case.

### DIFF
--- a/helm-info.el
+++ b/helm-info.el
@@ -64,11 +64,12 @@ files with `helm-info-at-point'."
           (goto-char (point-min))
           (while (search-forward "\n* " nil t)
             (unless (search-forward "Menu:\n" (1+ (point-at-eol)) t)
-              (setq start (point-at-bol)
-                    end (point-at-eol))
-              (with-current-buffer tobuf
-                (insert-buffer-substring infobuf start end)
-                (insert "\n")))))))))
+              (let ((info-index (buffer-substring
+                                 (point-at-bol)
+                                 (point-max))))
+                (with-current-buffer tobuf
+                  (insert (replace-regexp-in-string "\n   " " " info-index) )))
+              (goto-char (point-max)))))))))
 
 (defun helm-info-goto (node-line)
   (Info-goto-node (car node-line))


thierryvolpiatto:
I am not sure to understand the problem, can you provide a recipe to reproduce the issue you describe.
Where the screenshot you sent comes from ?

Lompik:
sorry I wasn't clear. To reproduce, use `helm-info-emacs` and goto to item `add-file-local-variable`. The problem is that helm won't put the cursor on line 80 like it normally does for short index items (try with `add-global-abbrev`). Indeed `helm-info-display-to-real` returns the wrong line number (1) because the line information is not there. The screenshot shows the `Command Index` of the emacs info manual with `buffer-invisibility-spec` set to `nil` where `helm-info-emacs` sources candidates. Currently it is wrongly skipping the line with `(line 80)` .

thierryvolpiatto:
Thanks for explanations, and good catch, I never realized some lines was displaying the invisible spec on next line.
The problem of inserting the whole buffer at once like you propose
is it may insert some unwanted lines like newlines.
I have commited a different fix, have a try and let me know if it is working fine for you.

Thanks.

Lompik:
Yes thanks it works. I'm closing this.

I just measured how long `helm-info-init` took in both cases : ~0.6s for inserting whole buffer and ~0.75 seconds for yours (on a i7-4650U). I was supprised it wasn't that much  !!

thierryvolpiatto:
Stephane Fontaine <notifications@github.com> writes:

> Yes thanks it works. I'm closing this.
>
> I just measured how long helm-info-init took in both cases : ~0.6s for
> inserting whole buffer and ~0.75 seconds for yours (on a i7-4650U). I
> was supprised it wasn't that much !!

Be sure both version are compiled (or not) for comparison.  In many
places in emacs, it is faster printing/inserting items one by one
instead of inserting the whole buffer-string.
Try for example describe-variable on a huge variable, you will see it
takes a while because describe-variable is printing the whole string at
once (You will find an optimized version here:
https://github.com/thierryvolpiatto/emacs-tv-config/blob/master/.emacs.el#L244).

Thanks.

-- 
Thierry

Issue #1532, opened 2016-06-08
Failed to set `helm-top-command' to  "env COLUMNS=%s top -b -n 1 -c")


xor3:
Failed to set `helm-top-command' to  "env COLUMNS=%s top -b -n 1 **-c**").
I just want to add the 'c' flag.

```
(defun helm-top-command-set-fn (var _value)
  (set var
       (cl-case system-type
         (darwin "env COLUMNS=%s ps -axo pid,user,pri,nice,ucomm,tty,start,vsz,%%cpu,%%mem,etime,command")
         (t      "env COLUMNS=%s top -b -n 1"))))

(defcustom helm-top-command "env COLUMNS=%s top -b -n 1"
  "Top command used to display output of top.
To use top command, a version supporting batch mode (-b option) is needed.
On Mac OSX top command doesn't support this, so ps command
is used by default instead.
If you modify this the number and order of elements displayed
should be the same as top command to have the sort commands
working properly, that is 12 elements with the 2 first being
PID and USER and the last 4 being %CPU, %MEM, TIME and COMMAND.
A format string where %s will be replaced with `frame-width'."
  :group 'helm-sys
  :type 'string
  :set  'helm-top-command-set-fn)
```

Is  the  `helm-top-command-set-fn'  ok?

PR #1533, opened 2016-06-08
Fix helm-top-command (#1532)

xuchunyang:
To make users be able customize this variable


thierryvolpiatto:

> To make users be able customize this variable

I commited a change while you were commiting a similar fix at the same
time, I have fixed conflicts with my changes and pushed your patch
which is better instead of mine.

Thanks.

-- 
Thierry

PR #1534, opened 2016-06-10
Run helm-check-minibuffer-input within helm window

xuchunyang:
Otherwise I get one quick refresh every time after I call
`helm-find-files'.

The issue was introduced in 7c92237

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -1598,7 +1598,8 @@ or when `helm-pattern' is equal to \"~/\"."
                  "\\`\\([.]\\|\\s-\\)\\{2\\}[^/]+"
                  (helm-basename helm-pattern))
                 (string-match-p "/\\'" helm-pattern))
-           (helm-ff-recursive-dirs helm-pattern))
+           (helm-ff-recursive-dirs helm-pattern)
+           (with-helm-window (helm-check-minibuffer-input)))
           ((and (string-match
                  "/?\\$.*/\\|/\\./\\|/\\.\\./\\|/~.*/\\|//\\|\\(/[[:alpha:]]:/\\|\\s\\+\\)"
                  helm-pattern)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:

> Otherwise I get one quick refresh every time after I call
> `helm-find-files'.
>
> The issue was introduced in 7c92237

Thanks for catching this, however your change is wrong, you have to wrap
the whole cond block inside with-helm-window otherwise only one clause
(the second) receive the helm-check-minibuffer-input.
I will commit the change tomorrow morning unless you do it in this PR.

Thanks.

-- 
Thierry

thierryvolpiatto:
Using with-helm-window on top doesn't help, here a fix that works:

```elisp

(defun helm-ff-auto-expand-to-home-or-root ()
  "Allow expanding to home/user directory or root or text yanked after pattern."
  (when (helm-file-completion-source-p)
      (cond ((and (null (file-exists-p helm-pattern))
                  (string-match-p
                   "\\`\\([.]\\|\\s-\\)\\{2\\}[^/]+"
                   (helm-basename helm-pattern))
                  (string-match-p "/\\'" helm-pattern))
             (helm-ff-recursive-dirs helm-pattern)
             (with-helm-window (helm-check-minibuffer-input)))
            ((and (string-match
                   "/?\\$.*/\\|/\\./\\|/\\.\\./\\|/~.*/\\|//\\|\\(/[[:alpha:]]:/\\|\\s\\+\\)"
                   helm-pattern)
                  (with-current-buffer (window-buffer (minibuffer-window)) (eolp))
                  (not (string-match helm-ff-url-regexp helm-pattern)))
             (let* ((match (match-string 0 helm-pattern))
                    (input (cond ((string= match "/./")
                                  (expand-file-name default-directory))
                                 ((string= helm-pattern "/../") "/")
                                 ((string-match-p "\\`/\\$" match)
                                  (let ((sub (substitute-in-file-name match)))
                                    (if (file-directory-p sub)
                                        sub (replace-regexp-in-string "/\\'" "" sub))))
                                 (t (expand-file-name
                                     (helm-substitute-in-filename helm-pattern)
                                     ;; [Windows] On UNC paths "/" expand to current machine,
                                     ;; so use the root of current Drive. (i.e "C:/")
                                     (and (memq system-type '(windows-nt ms-dos))
                                          (getenv "SystemDrive")) ; nil on Unix.
                                     )))))
               (if (file-directory-p input)
                   (setq helm-ff-default-directory
                         (setq input (file-name-as-directory input)))
                   (setq helm-ff-default-directory (file-name-as-directory
                                                    (file-name-directory input))))
               (with-helm-window
                 (helm-set-pattern input)
                 (helm-check-minibuffer-input)))))))

```

xuchunyang:
I have included your patch and force push.

thierryvolpiatto:
Great thanks.

PR #1535, opened 2016-06-11
Don't append space if input is empty

syohex:
There are some candidates but `helm--completion-in-region` says `No matches` if input string is empty.

#### Original code
orig

#### Fixed version
fixed

See also
- https://github.com/syohex/emacs-helm-robe/issues/2


thierryvolpiatto:
Thanks.

PR #1537, opened 2016-06-14
Add helm-quit-and-helm-mini feature

zw963:


helm-buffers.el
```diff
@@ -899,6 +899,14 @@ displayed with the `file-name-shadow' face if available."
         :ff-transformer-show-only-basename nil
         :truncate-lines helm-buffers-truncate-lines))
 
+(defun helm-quit-and-helm-mini ()
+  "Drop into `helm-mini' from `helm'."
+  (interactive)
+  (with-helm-alive-p
+    (helm-run-after-exit
+     (lambda ()
+       (helm-mini)))))
```
xuchunyang: No needs to wrap `helm-mini`, `(helm-run-after-exit 'helm-mini)` is enough

zw963:
Thank @xuchunyang .

I will fix things which need adjust to satisfied demand here.

After work well, I will rebase and squish only one commit here.

thierryvolpiatto:
Thanks!

zw963:
......

I still not `rebase -i' those two commit into one commit ...

Fine, in fact, I like keep those branch details like this, though, some maintainer hope more tidy code ...

thierryvolpiatto:
Billy.Zheng <notifications@github.com> writes:

> ......
>
> I still not `rebase -i' those two commit into one commit ...

No problem it's ok like this with two commits, ;-) thanks.

-- 
Thierry

PR #1541, opened 2016-06-19
Use Helm source instead of `completing-read`

xiongtx:
This change lays the groundwork for having a common Helm source of commands for
execution and apropos.

See issue #1448.

### DIFF
--- a/helm-command.el
+++ b/helm-command.el
@@ -24,6 +24,8 @@
 (require 'helm-elisp)
 
 
+;;; Customize
+
 (defgroup helm-command nil
   "Emacs command related Applications and libraries for Helm."
   :group 'helm)
@@ -49,10 +51,19 @@ Show all candidates on startup when 0 (default)."
   :group 'helm-command
   :type 'boolean)
 
+(defcustom helm-M-x-actions
+  '(("Execute" . helm-M-x-execute)
+    ("Find command" . helm-find-function)
+    ("Info lookup" . (lambda (candidate)
+                       (helm-info-lookup-symbol (helm-stringify candidate))))
+    ("Describe command" . helm-describe-function))
+  "Actions for `helm-M-x'."
+  :group 'helm-command
+  :type '(alist :key-type string :value-type function))
+
 
 ;;; Faces
-;;
-;;
+
 (defgroup helm-command-faces nil
   "Customize the appearance of helm-command."
   :prefix "helm-"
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:

> This change lays the groundwork for having a common Helm source of commands for
> execution and apropos.
>
> See issue #1448.

Thanks, 

but I don't think it is a good idea to change helm-M-x and make
it doing something else than just running a command and what it does
actually, that is allowing user to see the doc before pressing RET.
Lets keep this as it is.

I would accept a change using helm and not helm-comp-read, but not
implemented as you did:

- Big loss of performances using mapatoms and then a loop transforming
  the candidates in cons, then another adding bindings etc...
  (use all-completions instead of mapatoms among other things)
  This also doesn't fit well with lexical-binding.
- Big loss of performances using sync source.

So, such changes would make helm-M-x not better than the
execute-extended-command helmized in terms of performances.  Also
make-obsolete is called with wrong number of args but this is a detail.

-- 
Thierry

xiongtx:
> but I don't think it is a good idea to change helm-M-x and make it doing something else than just running a command and what it does actually, that is allowing user to see the doc before pressing RET. Lets keep this as it is.

Not sure what you mean by this...the goal of this is to eventually use a single Helm source for `helm-M-x` and `helm-apropos` so we're not duplicating the work, and can conveniently jump to source if we want to know more about a command before executing it, or execute a command after we've apropo'ed it. That's what we addressed in issue #1448.

As far as performance is concerned, fast enough is fast enough. `helm-build-sync-source` is supposed to be the standard way to build a source, right? A potential performance concern is the alphabetical sorting of commands--which I find useful and logical, but which may not be necessary.

And can you elaborate on the lexical binding issue?

thierryvolpiatto:

> Not sure what you mean by this...the goal of this is to eventually use
> a single Helm source for helm-M-x and helm-apropos so we're not
> duplicating the work, and can conveniently jump to source if we want
> to know more about a command before executing it, or execute a command
> after we've apropo'ed it. That's what we addressed in issue #1448.

Yes but I think it is not a good idea to unify apropos and M-x which are
two different things.

> As far as performance is concerned, fast enough is fast
> enough.

Because you are maybe using a fast machine, but in the past I had a lot
of bug reports about helm-M-x too slow, idem for helm-apropos.

> helm-build-sync-source is supposed to be the standard way to build a
> source, right?

No, not necessarily, look at the helm sources you will see a lot of
sources build with in-buffer method, like helm-M-x or helm-apropos. 

> A potential performance concern is the alphabetical sorting of
> commands--which I find useful and logical, but which may not be
> necessary.

In your patch the sorting have changed, and no the main performance
issue is in your candidate function(s), i.e (you loop on the result of
maphash, then loop again at least one time for filtering)

> And can you elaborate on the lexical binding issue?

Compile your files to see warnings.

-- 
Thierry

PR #1542, opened 2016-06-21
Make helm-resume honor helm-full-frame

dandavison:
Currently, under `helm-resume`, when there is no buffer-local value of `helm-full-frame`, the value `nil` is used rather than the global value of `helm-full-frame`. This PR changes that.

##### Testing
I've tested this manually by setting `helm-full-frame` to `t` globally and confirming that on this branch the resume buffer is in a full frame window.


thierryvolpiatto:
Thanks, I will just modify by using `(default-value 'helm-full-frame)` which make the intention more clear.

PR #1543, opened 2016-06-21
Add helm-M-x-allow-prefix-argument option

xuchunyang:
When the option is non-nil (defaults to nil), prefix argument is allowed
before calling helm-M-x, but it will not be used if you specify prefix
argument right before executing command, for example,

    M-3 helm-M-x foo RET      => (foo 3)
    M-3 helm-M-x foo M-4 RET  => (foo 4)

The current (default) behaviour of helm-M-x remains the same.

### DIFF
--- a/helm-command.el
+++ b/helm-command.el
@@ -49,6 +49,14 @@ Show all candidates on startup when 0 (default)."
   :group 'helm-command
   :type 'boolean)
 
+(defcustom helm-M-x-allow-prefix-argument nil
+  "Allow specifying prefix argument before `helm-M-x' when non--nil.
+Note if you set it to non--nil and specify a prefix argument
+before `helm-M-x', then you will NOT be able to cancel it though
+you can change it to another vaule."
+  :group 'helm-command
+  :type 'boolean)
+
 
 ;;; Faces
 ;;
@@ -187,16 +195,29 @@ than the default which is OBARRAY."
                        do (set-text-properties 0 (length c) nil c)
                        and collect c))
         (unwind-protect
-             (let ((msg "Error: Specifying a prefix arg before calling `helm-M-x'"))
-               (when current-prefix-arg
-                 (ding)
-                 (message "%s" msg)
-                 (while (not (sit-for 1))
-                   (discard-input))
-                 (user-error msg))
+            (let ((msg "Error: Specifying a prefix arg before calling `helm-M-x'")
+                  (prompt "M-x "))
+              (if helm-M-x-allow-prefix-argument
+                  (setq prompt
+                        (concat (cond
+                                 ((eq current-prefix-arg '-) "- ")
+                                 ((and (consp current-prefix-arg)
+                                       (eq (car current-prefix-arg) 4)) "C-u ")
+                                 ((and (consp current-prefix-arg)
+                                       (integerp (car current-prefix-arg)))
+                                  (format "%d " (car current-prefix-arg)))
+                                 ((integerp current-prefix-arg)
+                                  (format "%d " current-prefix-arg)))
+                                "M-x "))
+                (when current-prefix-arg
+                  (ding)
+                  (message "%s" msg)
+                  (while (not (sit-for 1))
+                    (discard-input))
+                  (user-error msg)))
                (setq current-prefix-arg nil)
                (helm-comp-read
-                "M-x " (or collection obarray)
+                prompt (or collection obarray)
                 :test 'commandp
                 :requires-pattern helm-M-x-requires-pattern
                 :name "Emacs Commands"


thierryvolpiatto:
When you do `M-3 M-x foo` `helm-M-x` should start with the mode-line updated with 3 as prefix value.

xuchunyang:
In a helm-M-x session, mode-line now displays prefix argument which will be used by any next command (such as `a` and `C-n`), not just `RET`, so displaying 3 initially for `M-3 M-x` will break the rule.

xuchunyang:
Now it's done.

Besides, instead of providing this option (and disabled by default), can we just allow prefix argument before `helm-M-x` since specifying prefix argument after helm-M-x still works.

thierryvolpiatto:

> Now it's done.

Thanks, but still not the behavior expected, now when you do C-u M-x the
prefix arg is kept for all the session i.e C-n should remove the prefix
arg.

-- 
Thierry

xuchunyang:
the problem is if C-n remove the prefix arg, the prefix arg can't pass to C-n, so the initial prefix arg is special. And if user doesn't specify prefix arg before RET, for example, `C-u M-x foo RET`, the initial prefix arg C-u passes to RET but doesn't display in the mode line.

xuchunyang:
I have pushed another commit to do what you want.

thierryvolpiatto:

> I have pushed another commit to do what you want.

Still not working:

C-u M-x C-n
The mode-line doesn't show the prefix arg but it is still active.
You can verify by pressing RET on e.g emacs-version, the emacs version
is printed in current-buffer.

-- 
Thierry

xuchunyang:
if we clear prefix argument before helm-M-x after every command such as `a` and `C-n`, then there is no need to support prefix argument before helm-M-x, since it works only for `M-x RET` and nothing else.

thierryvolpiatto:
Hmm, you are right, so I think allowing prefix arg before M-x is just confusing.
Also once started there is no way to disable it.

xuchunyang:
how about treating prefix arg before helm-M-x as the default prefix arg (i.e., it will be used if user doesn't specify prefix arg right before RET, otherwise it will be overridden) and displaying it in prompt like Emacs's built M-x ?

thierryvolpiatto:

> how about treating prefix arg before helm-M-x as the default prefix
> arg (i.e., it will be used if user doesn't specify prefix arg right
> before RET, otherwise it will be overridden) and displaying it in
> prompt like Emacs's built M-x ?

No, I don't believe in prefix arg before M-x, even if it is the vanilla
emacs behavior since decades, it is IMO non sense to prepare a prefix
arg for a command you don't know in advance.
The emacs behavior assume you know exactly which command you will run
before starting M-x, once you have entered the prefix arg you can't go
back if you change your mind, helm behavior at opposite assume you don't
know exactly the command you will run and allow changing while in the
session.

Actually from your last patch the behavior stay really confusing:

C-u M-x C-u C-u C-n
The prefix arg is still present, it should be disabled.

OTOH:
C-h M-x C-n, like we conclude should not disable prefix arg.

IOW As soon as we hit C-u from M-x session, the initial arg should be
disabled.

Perhaps a good behavior would be:

C-u M-x C-u => First C-u disable all prefix arg (num pref arg == 0).
C-u => Second C-u start again setting prefix-arg starting at '(4).

-- 
Thierry

xuchunyang:
> No, I don't believe in prefix arg before M-x, even if it is the vanilla
> emacs behavior since decades, it is IMO non sense to prepare a prefix
> arg for a command you don't know in advance.
> The emacs behavior assume you know exactly which command you will run
> before starting M-x, once you have entered the prefix arg you can't go
> back if you change your mind, helm behavior at opposite assume you don't
> know exactly the command you will run and allow changing while in the
> session.

I agree that it is a nice feature that helm-M-x allows setting/changing prefix arg afterwards, but I often know what command and what prefix arg that I need before hitting M-x, so when I use helm-M-x, I might forget to specify prefix arg before RET especially when filtering command takes me much effort. So inhibiting setting prefix arg before M-x is inconvenient for me.

> IOW As soon as we hit C-u from M-x session, the initial arg should be
> disabled.
> 
> Perhaps a good behavior would be:
> 
> C-u M-x C-u => First C-u disable all prefix arg (num pref arg == 0).
> C-u => Second C-u start again setting prefix-arg starting at '(4).

I don't set prefix argument before M-x by chance so I don't like to disable it easily, and even so typing a C-g to start again is easy to me.

xuchunyang:
Oops, I probably misunderstood you, at least I find my code is not correct at all.

xuchunyang:
> Oops, I probably misunderstood you, at least I find my code is not correct at all.

Never mind, I neglected helm-M-x's first argument, but the code worked like my expect. I have forced push one new commit, it should be easy to review.

It is true that if user turns this option on, then he won't be able to cancel initial prefix argument (though he can change it to another value).

thierryvolpiatto:

> I don't set prefix argument before M-x by chance so I don't like to
> disable it easily, and even so typing a C-g to start again is easy to
> me.

I have pushed a branch based on your patch, let me know what you think
of it.


-- 
Thierry

xuchunyang:
It works like your description and I can accept it. Thanks.

One small issue is after using first C-u to disable initial prefix, the prompt still displays the initial prefix arg.

thierryvolpiatto:

> It works like your description and I can accept it. Thanks.
>
> One small issue is after using first C-u to disable initial prefix,
> the prompt still displays the initial prefix arg.

Feel free to modify directly on this branch and merge when you feel it
is ready, I gave you write access to helm (let me know if it is not
working).

-- 
Thierry

xuchunyang:
It looks like I still don't have permission to push. or do I need to setting anything?

```
~/Projects/helm $ git push --verbose upstream xuchunyang_patch_modified 
Pushing to https://github.com/emacs-helm/helm.git
remote: Permission to emacs-helm/helm.git denied to xuchunyang.
fatal: unable to access 'https://github.com/emacs-helm/helm.git/': The requested URL returned error: 403
```

xuchunyang:
> It looks like I still don't have permission to push. or do I need to setting anything?

It's working now. I accepted the invitation.

thierryvolpiatto:

>     It looks like I still don't have permission to push. or do I need to setting anything?
>    
> It's working now. I accepted the invitation.

Great, I made a small modification to your change and now modifying
prompt works.

Just test it a bit, if looks ok to you perhaps we can merge ?

Maybe remove helm-M-x-allow-prefix-argument to use this new behavior
unconditionally ?

-- 
Thierry

PR #1545, opened 2016-06-22
Xuchunyang patch modified (allow prefix arg before M-x)

thierryvolpiatto:
Continue #1543  here.


xuchunyang:
> Maybe remove helm-M-x-allow-prefix-argument to use this new behavior
unconditionally ?

I agree

xuchunyang:
I tried it for a few times and it worked as expected, so I merged this.

thierryvolpiatto:

> I tried it for a few times and it worked as expected, so I merged this.

Great, good job.

Thanks.

-- 
Thierry

PR #1548, opened 2016-06-23
Rework helm-M-x-prefix-argument (#1545)

xuchunyang:
Set helm-M-x-prefix-argument within `interactive' form so it works for
keyboard macro, and before (helm-M-x-read-extended-command) so it works
for C-g too.

To avoid calling `called-interactively-p', inhibit calling helm-M-x from
lisp explicitly by declaring helm-M-x interactive-ony. Anyway, I don't
think being able to call helm-M-x from lisp is necessary.

### DIFF
--- a/helm-command.el
+++ b/helm-command.el
@@ -210,48 +210,39 @@ than the default which is OBARRAY."
                        and collect c))
         (unwind-protect
              (progn
-               (setq helm-M-x-prefix-argument current-prefix-arg)
                (setq current-prefix-arg nil)
-               (condition-case nil
-                   (helm-comp-read
-                    (concat (cond
-                             ((eq helm-M-x-prefix-argument '-) "- ")
-                             ((and (consp helm-M-x-prefix-argument)
-                                   (eq (car helm-M-x-prefix-argument) 4)) "C-u ")
-                             ((and (consp helm-M-x-prefix-argument)
-                                   (integerp (car helm-M-x-prefix-argument)))
-                              (format "%d " (car helm-M-x-prefix-argument)))
-                             ((integerp helm-M-x-prefix-argument)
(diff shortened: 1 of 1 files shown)


xuchunyang:
#1545

helm-command.el
```diff
@@ -264,7 +255,11 @@ Unlike regular `M-x' emacs vanilla `execute-extended-command' command,
 the prefix args if needed, are passed AFTER starting `helm-M-x'.
 
 You can get help on each command by persistent action."
-  (interactive (list current-prefix-arg (helm-M-x-read-extended-command)))
+  (declare (interactive-only command-execute))
```
thierryvolpiatto: Using declare here doesn't work, use (put ...) instead.
BTW why putting interactive-only prop on `command-execute`, did I miss something ?

helm-command.el
```diff
@@ -264,7 +255,11 @@ Unlike regular `M-x' emacs vanilla `execute-extended-command' command,
...
+  (interactive
+   (progn
+     (setq helm-M-x-prefix-argument current-prefix-arg)
```
thierryvolpiatto: Even simpler if perhaps using something like this which avoid the unused arg ARG at compile time
and is simpler IMO (correct me if I am wrong, I am tired tonight).

```elisp
  (interactive
   (progn
     current-prefix-arg
     (list current-prefix-arg (helm-M-x-read-extended-command))))
  (setq helm-M-x-prefix-argument arg)
```

thierryvolpiatto:
Forget comments added yesterday, verify if all is good for you, for me it is ready to merge.

Thanks.

xuchunyang:
Just tested and it worked good for me.

PR #1551, opened 2016-06-27
Avoid a quick refresh of Echo Area in helm-describe-function

xuchunyang:
When I execute persistent-action in helm-M-x, helm-apropos and C-h f, I
notice a quick refresh of Echo Area (this is annoying).  This patch is a
workaround.


thierryvolpiatto:

> When I execute persistent-action in helm-M-x, helm-apropos and C-h f, I
> notice a quick refresh of Echo Area (this is annoying). This patch is a
> workaround.

Looks good, please merge.

Thanks.

-- 
Thierry

PR #1588, opened 2016-08-23
Expand default-directory in helm-do-grep-ag.

blorbx:
If the `default-directory` has a "~" in it, it will be escaped later with `shell-quote-argument` in `helm-grep-ag-prepare-cmd-line`. This causes issues trying to use `helm-do-grep-ag` in the home directory.


thierryvolpiatto:
Merged thanks.

PR #1591, opened 2016-08-28
Fuzzy matching for helm-eshell

bb2020:
Fuzzy is benefitical to helm-eshell especially when searching for files (e.g. by extension). One doesn't have to leave spaces between path and the query.

### DIFF
--- a/helm-eshell.el
+++ b/helm-eshell.el
@@ -45,6 +45,12 @@
   :group 'helm)
 
 
+(defcustom helm-eshell-fuzzy-match nil
+  "Enable fuzzy matching in `helm-esh-pcomplete' when non--nil."
+  :group 'helm-eshell
+  :type 'boolean)
+
+
 (defvar helm-eshell-history-map
   (let ((map (make-sparse-keymap)))
     (set-keymap-parent map helm-map)
@@ -71,6 +77,7 @@
    (nomark :initform t)
    (persistent-action :initform 'ignore)
    (nohighlight :initform t)
+   (fuzzy-match :initform (eval helm-eshell-fuzzy-match))
    (filtered-candidate-transformer
     :initform
     (lambda (candidates _sources)


helm-eshell.el
```diff
@@ -71,6 +77,7 @@
    (nomark :initform t)
    (persistent-action :initform 'ignore)
    (nohighlight :initform t)
+   (fuzzy-match :initform (eval helm-eshell-fuzzy-match))
```
thierryvolpiatto: Instead of adding a slot here with `eval`, I would prefer adding a method here:

```elisp
(defmethod helm--setup-source ((source helm-esh-source))
  (setf (slot-value source 'fuzzy-match helm-eshell-fuzzy-match)))
```

thierryvolpiatto:
Thanks, I have added a comment, have a look.

helm-eshell.el
thierryvolpiatto: There is a typo in code above, read instead:

```elisp
(defmethod helm--setup-source ((source helm-esh-source))
  (setf (slot-value source 'fuzzy-match) helm-eshell-fuzzy-match))
```
Sorry.

bb2020: I think it is better to run `:before`. Otherwise it breaks classical helm completion.

```
(defmethod helm--setup-source :before ((source helm-esh-source))
  (setf (slot-value source 'fuzzy-match) helm-eshell-fuzzy-match))
```

bb2020:
Answered in comment.

helm-eshell.el
thierryvolpiatto: bb2020 <notifications@github.com> writes:

> In helm-eshell.el:
>
>> @@ -71,6 +77,7 @@
>>     (nomark :initform t)
>>     (persistent-action :initform 'ignore)
>>     (nohighlight :initform t)
>> +   (fuzzy-match :initform (eval helm-eshell-fuzzy-match))
>
> I think it is better to run :before. Otherwise it breaks classical helm completion.

And did you try adding:

(fuzzy-match :initform helm-eshell-fuzzy-match)

I think eval isn't needed.

> (defmethod helm--setup-source :before ((source helm-esh-source))
>   (setf (slot-value source 'fuzzy-match) helm-eshell-fuzzy-match))
>
> —


-- 
Thierry

bb2020: Apologizing. I've missed the fist comment.

bb2020: Without `eval` fuzzy-matching is always enabled. It treats `helm-eshell-fuzzy-match` as `t` even it has value of `nil`.

thierryvolpiatto:
Ok thanks, merged.

PR #1609, opened 2016-09-18
allow user to choose a different emacs when running ``make''.

sameerds:
EMACS_COMMAND indicates the emacs to be used in the Makefile. Invoking make
with ``EMACS_COMMAND=/path/to/custom/emacs'' overrides the default.

* Makefile: Do it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@
 ## Floor, Boston, MA 02110-1301, USA.
 
 # Emacs invocation
-EMACS_COMMAND   := emacs
+EMACS_COMMAND   ?= emacs
 
 EMACS		:= $(EMACS_COMMAND) -Q -batch
 


xuchunyang:
hmm, I am not a Makefile expert. But without the change, `make EMACS_COMMAND=/path/to/emacs` works for me, though `EMACS_COMMAND=/path/to/emacs make` doesn't.

sameerds:
You are right, I  just checked the Make manual (should have done it first!)

https://www.gnu.org/software/make/manual/make.html#Overriding

It seems you can over-ride a variable by specifying it as "make EMACS_COMMAND=/path/to/emacs" ... you don't need the "?=" to make this happen!

I suppose we don't need this change then. My bad!

PR #1614, opened 2016-09-26
Added at-point google and wikipedia suggest defuns

jatimix:
* helm-net.el : (helm-google-suggest-at-point) : New. (helm-wikipedia-suggest-at-point) : New.
I'm not sure its relevant, but its a small function that I use 150 times a day.


thierryvolpiatto:
jatimix <notifications@github.com> writes:

>   * helm-net.el : (helm-google-suggest-at-point) :
>   New. (helm-wikipedia-suggest-at-point) : New.

You don't need this, add the google and wikipedia sources to
`helm-sources-using-default-as-input`.

--
Thierry

PR #1628, opened 2016-10-20
Use correct slot name `candidate-transformer`

xiongtx:
There is no slot called `candidates-transformer`.


thierryvolpiatto:
Thanks, merged.

PR #1634, opened 2016-10-29
Change boring buffer regexp

fice-t:
It only needs to check the beginning of each buffer name.

Also remove the redundant regexp *helm-mode.

### DIFF
--- a/helm-buffers.el
+++ b/helm-buffers.el
@@ -37,7 +37,7 @@
   :group 'helm)
 
 (defcustom helm-boring-buffer-regexp-list
-  '("\\` " "\\*helm" "\\*helm-mode" "\\*Echo Area" "\\*Minibuf")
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
fice-t <notifications@github.com> writes:

> It only needs to check the beginning of each buffer name.

Is this fixing a particular issue ?

-- 
Thierry

fice-t:
Not anything that I experienced personally, but it does eliminate false positives and makes it a bit more efficient.

thierryvolpiatto:
fice-t <notifications@github.com> writes:

> Not anything that I experienced personally, but it does eliminate
> false positives and makes it a bit more efficient.

Ok, just wanted to know if you had a particular problem, and yes
specifying the beginning of string makes the regexp more efficient.

Thanks.

--
Thierry

PR #1639, opened 2016-11-11
Speed up helm-for-files by 680%

CyberShadow:
`helm-skip-entries` was being wasteful in testing each regular expression in the blacklist and whitelist. Optimizing it by pre-concatenating the regular expression lists results in a drastic speed-up (480 ms to 70 ms on my machine).

### DIFF
--- a/helm-lib.el
+++ b/helm-lib.el
@@ -357,10 +357,14 @@ Default is `eq'."
         unless (gethash elm cont)
         collect (puthash elm elm cont)))
 
+(defsubst helm--string-join (strings &optional separator)
+  "Join all STRINGS using SEPARATOR."
+  (mapconcat 'identity strings separator))
+
 (defun helm--concat-regexps (regexp-list)
   "Return a regexp which matches any of the regexps in REGEXP-LIST."
   (if regexp-list
-      (concat "\\(?:" (string-join regexp-list "\\)\\|\\(?:") "\\)")
+      (concat "\\(?:" (helm--string-join regexp-list "\\)\\|\\(?:") "\\)")
     "\\<\\>"))                          ; Match nothing
 
 (defun helm-skip-entries (seq black-regexp-list &optional white-regexp-list)
--- a/helm-lib.el
+++ b/helm-lib.el
@@ -357,17 +357,22 @@ Default is `eq'."
         unless (gethash elm cont)
         collect (puthash elm elm cont)))
 
+(defun helm--concat-regexps (regexp-list)
+  "Return a regexp which matches any of the regexps in REGEXP-LIST."
+  (if regexp-list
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Ok, can you use something else than `string-join` (not in 24.5)?

Thanks.

xiongtx:
We should probably just defined `helm-string-join`:

```emacs-lisp
(defun helm-string-join (strings &optional separator)
  "Join all STRINGS using SEPARATOR."
  (mapconcat #'identity strings separator))
```

[CIDER has this](https://github.com/clojure-emacs/cider/blob/v0.14.0/cider-util.el#L323) and other compatibility functions in a special `cider-utils.el` file.

CyberShadow:
OK, added `helm--string-join`.

xuchunyang:
`string-join` is provided by `subr-x.el`, which is part of Emacs since 24.4. You should require that feature before using the function, since some version of Emacs doesn't preload `subr-x.el` such as Emacs 24.5.

CyberShadow:
@xuchunyang I already replaced the usage of `string-join` as per the above comments.

helm-lib.el
```diff
@@ -357,17 +357,22 @@ Default is `eq'."
         unless (gethash elm cont)
         collect (puthash elm elm cont)))
 
+(defun helm--concat-regexps (regexp-list)
+  "Return a regexp which matches any of the regexps in REGEXP-LIST."
+  (if regexp-list
+      (concat "\\(?:" (string-join regexp-list "\\)\\|\\(?:") "\\)")
```
thierryvolpiatto: And BTW, isn't this the same as `(regexp-opt regexp-list)`

thierryvolpiatto:
Isn't this the same as your patch ?

```elisp
(defun helm-skip-entries (seq black-regexp-list &optional white-regexp-list)
  "Remove entries which matches one of REGEXP-LIST from SEQ."
  (cl-loop with black-regexp = (regexp-opt black-regexp-list)
           with white-regexp = (and white-regexp-list
                                    (regexp-opt white-regexp-list))
           for i in seq
           unless (and (stringp i)
                       (string-match-p black-regexp i)
                       (or (null white-regexp)
                           (null (string-match-p white-regexp i))))
           collect i))
```

CyberShadow:
> Isn't this the same as your patch ?

No.

`regexp-opt` takes a list of *strings*. `helm--concat-regexps` takes a list of *regexps*.

thierryvolpiatto:

>     Isn't this the same as your patch ?
>    
> No.
>
> regexp-opt takes a list of strings. helm--concat-regexps takes a list of regexps.

Oh, right, so ok merging now.

Thanks.

-- 
Thierry

PR #1647, opened 2016-11-26
Search compressed files by helm-tags

tasuwo:
`find-tag` command will automatically check for the existence of the compressed versions of files if that is enabled. But helm-tags have colundn't. So I modified it in reference to [emacs/etags.el](https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/etags.el).

### DIFF
--- a/helm-tags.el
+++ b/helm-tags.el
@@ -56,14 +56,6 @@ one match."
   :type 'boolean
   :group 'helm-tags)
 
-(defcustom helm-etags-compression-info-list
-  (purecopy '("" ".Z" ".bz2" ".gz" ".xz" ".tgz"))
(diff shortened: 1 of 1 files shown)


helm-tags.el
```diff
@@ -56,6 +56,14 @@ one match."
   :type 'boolean
   :group 'helm-tags)
 
+(defcustom helm-etags-compression-info-list
```
thierryvolpiatto: Why not reusing `tags-compression-info-list` instead of creating a new var ?

tasuwo: There are no particular reason. Modified it.

thierryvolpiatto:
Thanks, merged.

PR #1662, opened 2017-01-15
Kill ring delete

xiongtx:
`M-D` to delete candidates is a good convention established by `helm-find-files`.


thierryvolpiatto:
Thanks!

PR #1664, opened 2017-01-16
Ring preview

xiongtx:
Rename `helm-show-kill-ring` to `helm-kill-ring`.

Add preview persistent action to `helm-show-kill-ring`.

Fix preview for `helm-mark-ring`, which was broken due to missing `switch-to-buffer`.

### DIFF
--- a/helm-ring.el
+++ b/helm-ring.el
@@ -206,15 +206,19 @@ This is a command for `helm-kill-ring-map'."
              collect m into recip
              finally return recip)))
 
+(defun helm--mark-ring-preview (candidate)
+  "Preview mark ring position at CANDIDATE."
+  (switch-to-buffer helm-current-buffer)
+  (helm-goto-line (string-to-number candidate))
+  (helm-highlight-current-line))
+
 (defvar helm-source-mark-ring
   (helm-build-sync-source "mark-ring"
     :candidates #'helm-mark-ring-get-candidates
     :action '(("Goto line"
                . (lambda (candidate)
-                   (helm-goto-line (string-to-number candidate))))) 
-    :persistent-action (lambda (candidate)
-                         (helm-goto-line (string-to-number candidate))
-                         (helm-highlight-current-line))
+                   (helm-goto-line (string-to-number candidate)))))
+    :persistent-action #'helm--mark-ring-preview
     :persistent-help "Show this line"))
 
 ;;; Global-mark-ring
--- a/helm-ring.el
+++ b/helm-ring.el
@@ -73,12 +73,12 @@ If nil or zero (disabled), don't truncate candidate, show all."
     :candidates #'helm-kill-ring-candidates
     :filtered-candidate-transformer #'helm-kill-ring-transformer
     :action 'helm-kill-ring-actions
-    :persistent-action 'ignore
-    :persistent-help "DoNothing"
+    :persistent-action #'helm--kill-ring-preview
+    :persistent-help "Preview entry"
     :keymap helm-kill-ring-map
     :migemo t
     :multiline t)
-  "Source for browse and insert contents of kill-ring.")
+  "Helm source for browsing and inserting contents of `kill-ring.'")
 
 (defun helm-kill-ring-candidates ()
   (cl-loop for kill in (helm-fast-remove-dups kill-ring :test 'equal)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Sorry but I don't see in what it is helping to have a preview buffer for kill ring.
For mark ring, can you describe what is broken, I didn't find.

xiongtx:
It's helpful to have a buffer for previewing the kill ring because kill ring entries longer than `helm-kill-ring-max-lines-number` lines are truncated.

The mark ring preview doesn't work without the change. Try it--it does nothing.

thierryvolpiatto:

> It's helpful to have a buffer for previewing the kill ring because
> kill ring entries longer than helm-kill-ring-max-lines-number lines
> are truncated.

If you really need to see the whole contents of candidates, perhaps a
better approach is writing a function to toggle truncated/not-truncated
view in the helm-buffer.
It's really simple to do with something like this:

    (setq flag (not flag))
    (if flag 
        (setq helm-kill-ring-max-lines-number 9999)
        (setq helm-kill-ring-max-lines-number initial-value))
    (helm-update)

An other option is to always have a high value to
helm-kill-ring-max-lines-number.

>
> The mark ring preview doesn't work without the change. Try it--it does nothing.

Yes IIRC persistent-action is disabled in kill ring.

--
Thierry

xiongtx:
Not sure why you oppose previewing contents of `kill-ring`: there are previews for `mark-ring`, and the solution you suggest would have to be included in everyone's init files.

> An other option is to always have a high value to `helm-kill-ring-max-lines-number`.

No matter how high you set this value, you could always end up w/ even longer text.

> Yes IIRC persistent-action is disabled in kill ring.

Persistent action works for `global-mark-ring`, but not `mark-ring`. This PR just addresses that inconsistency. Not sure what kill ring has to do w/ it.

thierryvolpiatto:

> Not sure why you oppose previewing contents of kill-ring: there are
> previews for mark-ring,

No they is no preview in mark-ring, just a persistent action which is
showing a line in a buffer.
The preview you are proposing create a new buffer only to show the
candidate contents, this buffer will add complexity to handle
follow-mode, different windows layout etc...

> and the solution you suggest would have to be included in everyone's
> init files.

No you misundertood what I proposed, the function would be provided in
helm with its keybinding, so nothing to add to init files.

> Persistent action works for global-mark-ring, but not mark-ring.

I don't see this here, what is not working ?

--
Thierry

xiongtx:
> I don't see this here, what is not working ?

Set `'helm-split-window-default-side` to `'same` and try persistent action on the mark ring (not global mark ring) source.

thierryvolpiatto:

>     I don't see this here, what is not working ?
>
> Set 'helm-split-window-default-side to 'same and try persistent action
> on the mark ring (not global mark ring) source.

Ok, now I see, I have comitted a simple fix.

Try to make your PR's related to the same issue, this PR is treating two
issues at the same time which create confusion.

Thanks.

--
Thierry

thierryvolpiatto:

> Not sure why you oppose previewing contents of kill-ring: there are
> previews for mark-ring, and the solution you suggest would have to be
> included in everyone's init files.

Here the solution I suggest (just eval and try):

    (defvar helm-kill-ring-truncated-flag nil)
    (defun helm-kill-ring-toggle-truncated ()
      (interactive)
      (setq helm-kill-ring-truncated-flag (not helm-kill-ring-truncated-flag))
      (if helm-kill-ring-truncated-flag
          (let ((helm-kill-ring-max-lines-number 99999))
            (helm-update))
          (helm-update)))
    (define-key helm-kill-ring-map (kbd "C-]") 'helm-kill-ring-toggle-truncated)

--
Thierry

xiongtx:
I think your solution is OK, except that it doesn't toggle truncation of the *current* candidate, only the first that could be truncated.

Can you add this feature to Helm when this issue is fixed?

thierryvolpiatto:

> I think your solution is OK, except that it doesn't toggle truncation
> of the current candidate, only the first that could be truncated.

No all candidates are toggled, but the preselection is not implemented,
so you are back to first candidate with this example.
I have a version working with preselection, however this raise another
issue which is the candidates with one single huge line; They are not
truncated because we are using a number of lines instead of using an offset.

> Can you add this feature to Helm when this issue is fixed?

Yes as soon as the issue above is fixed.

--
Thierry

thierryvolpiatto:
Feature have been now provided, closing.

PR #1673, opened 2017-01-24
add fuzzy-match support for session source

lewang:
Instead of making a new variable, I'm re-using `helm-recentf-fuzzy-match`.  Open to changes, let me know.


thierryvolpiatto:
Thanks.

Issue #749, opened 2014-12-12
`helm-buffer--details` too slow on TRAMP buffers


Fuco1:
In my day job I work mostly over TRAMP and I often have 100+ TRAMP buffers opened.  `helm-buffer--details` calls a lot of file-system hitting functions (`file-exists-p`, `verify-visited-file-modtime`, etc...).  It takes about 200ms to process one buffer.  This makes `helm` time out (or something similar) and `helm-buffers-list` is completely unusable.

The simple solution is disable these checks on tramp files (`tramp-tramp-file-p`) or `file-remote-p` (not sure if this checks the same thing).  As a temporary fix, I have simply commented out the slow branches of the `cond` in the above mentioned function.

The above could maybe be controlled with some user-customizable flag---I'd rather have usable buffer switching without all the extra details than none at all :)

pclouds:
I'm using 20150504 from melpa and helm-mini is still slow for me when tramp buffers are opened. It takes a few seconds (even with C-] to turn off detail mode). That's too long for fast switching between buffers. I have recentf in helm-mini but I don't think it matters here. Is there any switch to avoid tramp in helm-mini?

tuhdo:
@pclouds You can set `helm-buffer-skip-remote-checking` to `t` and no remote checking for buffer is performed and there will be no more lag.

pclouds:
Perfect! Thank you!

pclouds:
Actually I it's still slow once in a while, then the following helm-mini is fast again. If I kill all those "*tramp/ssh <hostname>* buffers then it's always fast. Is it possible that helm-mini look into these buffers or something, that makes tramp contact the remote side?

thierryvolpiatto:
Set `tramp-verbose` to 6.
Set `helm-buffer-skip-remote-checking` to t.
Now switch to one of the tramp debug buffers ("*debug tramp/ssh user@host*") in a full window.
Move the point at end of buffer.
Now run helm-mini.
If tramp is involved, you will see the buffer growing, and you will know what is going on.

pclouds:
There are tramp activities even when helm-buffer-ski.. is t..

    11:45:30.513828 tramp-do-file-attributes-with-stat (5) # file attributes with stat: /path/to/file
[...]

PR #1679, opened 2017-01-26
delay highlighting file-attribute retrieval

aeruder:
This is a change that was discussed heavily in #749, fixed
incorrectly in c26d3a9 (this change ignored whether the file
in question was remote or not) and then essentially reverted
in 46687d1.

Here is the correct fix and speeds up my heavy TRAMP usage a ton
(from ~5 seconds pulling up helm-mini to instantaneous).

In order to recreate this, populate your recentf with remote files,
open one of these remote files (this part is 100% necessary), and
then do a helm-mini.  Since the connection to the remote files is open
the file-attribute will retrieve all the attributes of the files.

* helm-files.el (helm-highlight-files): delay file-attribute check

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -3470,28 +3470,28 @@ Colorize only symlinks, directories and files."
                                          (string-match ffap-url-regexp i)))
                                (not (string-match helm-ff-url-regexp i)))
                           (helm-basename i) i)
-           for type = (car (file-attributes i))
            collect
-           (cond ((and helm-ff-tramp-not-fancy
-                       (string-match helm-tramp-file-name-regexp i))
-                  (cons disp i))
-                 ((stringp type)
-                  (cons (propertize disp
-                                    'face 'helm-ff-symlink
-                                    'match-part (funcall mp-fn disp)
-                                    'help-echo (expand-file-name i))
-                        i))
-                 ((eq type t)
-                  (cons (propertize disp
-                                    'face 'helm-ff-directory
-                                    'match-part (funcall mp-fn disp)
-                                    'help-echo (expand-file-name i))
-                        i))
-                 (t (cons (propertize disp
-                                      'face 'helm-ff-file
-                                      'match-part (funcall mp-fn disp)
-                                      'help-echo (expand-file-name i))
-                          i)))))
+           (if (and helm-ff-tramp-not-fancy
+                    (string-match helm-tramp-file-name-regexp i))
+               (cons disp i)
+             (let ((type (car (file-attributes i))))
+               (cond ((stringp type)
+                      (cons (propertize disp
+                                        'face 'helm-ff-symlink
+                                        'match-part (funcall mp-fn disp)
+                                        'help-echo (expand-file-name i))
+                            i))
+                     ((eq type t)
+                      (cons (propertize disp
+                                        'face 'helm-ff-directory
+                                        'match-part (funcall mp-fn disp)
+                                        'help-echo (expand-file-name i))
+                            i))
+                     (t (cons (propertize disp
+                                          'face 'helm-ff-file
+                                          'match-part (funcall mp-fn disp)
+                                          'help-echo (expand-file-name i))
+                              i)))))))
 
 (defclass helm-files-in-current-dir-source (helm-source-sync helm-type-file)
   ((candidates :initform (lambda ()


helm-files.el
```diff
@@ -3470,28 +3470,28 @@ Colorize only symlinks, directories and files."
                                          (string-match ffap-url-regexp i)))
                                (not (string-match helm-ff-url-regexp i)))
                           (helm-basename i) i)
-           for type = (car (file-attributes i))
```
thierryvolpiatto: Perhaps just check if file (i) is remote and `helm-ff-tramp-not-fancy` e.g
for type = (and helm-ff-tramp-not-fancy (not (file-remote-p i)) (car (file-attributes i)...

thierryvolpiatto:
Yes you are right file-attributes should not be callled needlessly. Added some comment in diff.

Thanks.

helm-files.el
aeruder: Was attempting to not duplicate the check, but I was actually wondering if (file-remote-p) would be better than string-match basically everywhere?

thierryvolpiatto:
Andrew Ruder <notifications@github.com> writes:

> @aeruder commented on this pull request.
>
>
> In helm-files.el:
>
>> @@ -3470,28 +3470,28 @@ Colorize only symlinks, directories and files."
>                                           (string-match ffap-url-regexp i)))
>                                 (not (string-match helm-ff-url-regexp i)))
>                            (helm-basename i) i)
> -           for type = (car (file-attributes i))
>
> Was attempting to not duplicate the check, but I was actually
> wondering if (file-remote-p) would be better than string-match
> basically everywhere?

No they are two different things i.e ffap-url-regexp and
helm-ff-url-regexp doesn't match what's called remote files, they match
only urls.

--
Thierry

xiongtx:
I think @aeruder was talking about [`helm-tramp-file-name-regexp`](https://github.com/emacs-helm/helm/pull/1679/files#diff-c30ab41edecc9d4b288cf5765f90e290L3476) specifically.

thierryvolpiatto:
Ah yes, for some reason `helm-tramp-file-name-regexp` don't show up in my mailer.
So yes and no, `(string-match helm-tramp-file-name-regexp)` can be replaced by `file-remote-p` in some places but not other places, be careful.
Anyway this is another issue that should be treated separately.

thierryvolpiatto:
And yes in the case of `helm-highlight-files` , `(string-match helm-tramp-file-name-regexp)` can be replaced safely by `file-remote-p`.

xiongtx:
What's the difference b/t the two?

thierryvolpiatto:
So something like this should work as expected:

```diff
diff --git a/helm-files.el b/helm-files.el
index 6934adab..bb6cc02a 100644
--- a/helm-files.el
+++ b/helm-files.el
@@ -3470,11 +3470,12 @@ Colorize only symlinks, directories and files."
                                          (string-match ffap-url-regexp i)))
                                (not (string-match helm-ff-url-regexp i)))
                           (helm-basename i) i)
-           for type = (car (file-attributes i))
+           for isremote = (file-remote-p i)
+           for type = (and (or (and (null helm-ff-tramp-not-fancy) isremote)
+                               (null isremote))
+                           (car (file-attributes i)))
            collect
-           (cond ((and helm-ff-tramp-not-fancy
-                       (string-match helm-tramp-file-name-regexp i))
-                  (cons disp i))
+           (cond ((null type) (cons disp i))
                  ((stringp type)
                   (cons (propertize disp
                                     'face 'helm-ff-symlink
```

thierryvolpiatto:

> What's the difference b/t the two?

There is a subtle difference in HFF where `file-remote-p` will kick in as
soon as it see e.g "/ssh:" and fail because host is not provided yet,
and the check with `helm-tramp-file-name-regexp` will invoke remote
connection only when the second ":" is found.

-- 
Thierry

thierryvolpiatto:
I have finally merged your change, which is more readable than what I proposed previously.
I did some small modifications using if/else loop and using as suggested `file-remote-p`.

Thanks.

thierryvolpiatto:
@aeruder I have reworked this to handle the case where `helm-ff-tramp-not-fancy` is nil.
In this case `file-attributes` was called  and tramp connection initialized.
Not sure this was really happening as we bind `non-essential` at helm initialization, but better be safe.
Let me know if this change is ok for you.
Thanks.

PR #1681, opened 2017-01-30
Fix helm-man-woman in failures of man and woman

danielxia:
Changes:
- Pass absolute man file path to `man' without -l flag.
- When woman fails, the selected man section is retained. This fixes
the undesired behaviour that only the topic (without the section) is 
passed to `man' after such a failure.

The rationale behind the first change is that in BSD, man has no -l
option. Since the file name passed to man is generated from a woman
helper function, it is extremely unlikely to resemble a page
name. (see Help on `man')

### DIFF
--- a/helm-man.el
+++ b/helm-man.el
@@ -61,17 +61,20 @@ source.")
 (defun helm-man-default-action (candidate)
   "Default action for jumping to a woman or man page from helm."
   (let ((wfiles (mapcar #'car (woman-file-name-all-completions candidate))))
-    (let ((file (if (> (length wfiles) 1)
-                    (helm-comp-read "ManFile: " wfiles :must-match t)
-                  (car wfiles))))
-      (if (eq helm-man-or-woman-function 'Man-getpage-in-background)
-          (man file)
-        (condition-case nil
-            (woman-find-file file)
-          ;; If woman is unable to format correctly
-          ;; use man instead.
-          (error (kill-buffer)
-                 (man file)))))))
+    (condition-case nil
+        (let ((file (if (> (length wfiles) 1)
+                        (helm-comp-read "ManFile: " wfiles :must-match t)
+                      (car wfiles))))
+          (if (eq helm-man-or-woman-function 'Man-getpage-in-background)
+              (man (format helm-man-format-switches file))
+            (condition-case nil
+                (woman-find-file file)
+              ;; If woman is unable to format correctly
+              ;; use man instead.
+              (error (kill-buffer)
+                     (man file)))))
+      (error (kill-buffer)
+             (man candidate)))))
 
 (defun helm-man--init ()
   (require 'woman)
--- a/helm-man.el
+++ b/helm-man.el
@@ -62,19 +62,19 @@ source.")
   "Default action for jumping to a woman or man page from helm."
   (let ((wfiles (mapcar #'car (woman-file-name-all-completions candidate))))
     (condition-case nil
-        (let ((file (if (> (length wfiles) 1)
+        (let ((file (if (cdr wfiles)
                         (helm-comp-read "ManFile: " wfiles :must-match t)
                       (car wfiles))))
           (if (eq helm-man-or-woman-function 'Man-getpage-in-background)
-              (man (format helm-man-format-switches file))
+              (manual-entry (format helm-man-format-switches file))
             (condition-case nil
                 (woman-find-file file)
               ;; If woman is unable to format correctly
               ;; use man instead.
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Daniel Xia <notifications@github.com> writes:

> Changes:
>
>   * Pass absolute man file path to `man' without -l flag.

Isn't (setq helm-man-format-switches "%s") working?

>   * When woman fails, the selected man section is retained. This fixes
>     the undesired behaviour that only the topic (without the section) is
>     passed to `man' after such a failure.

IIRC Woman or man may fail due to a non existing or not valid file exists or
a file not supported by woman, in this case topic is used to allow
letting man find an alternative.
Can you describe the undesired behavior ?

> The rationale behind the first change is that in BSD, man has no -l
> option.

See issue #1035.

> Since the file name passed to man is generated from a woman helper
> function, it is extremely unlikely to resemble a page name. (see Help
> on `man')

Sorry I don't understand what you mean here.
In your patch you reuse a file that have been already tried and have
failed, a topic should be used as an alternative.

-- 
Thierry

danielxia:
Thank you Thierry for the comment. Sorry I didn't realize customizing `helm-man-format-switches' would work the moment I ran into the issue. You are right on this point.

> IIRC Woman or man may fail due to a non existing or not valid file exists or
a file not supported by woman, in this case topic is used to allow
letting man find an alternative.
Can you describe the undesired behaviour ?

`woman' cannot decode the man page files on my mac and consequently `Man-getpage-in-background' tries on CANDIDATE, i.e. the topic. However, it seems more friendly to me that before this last resort, we can take a chance on the file itself, thus the inner `condition-case'.

P.S. It's my first time to contribute to a open source project. My apology in advance if I have missed anything :).

thierryvolpiatto:
Daniel Xia <notifications@github.com> writes:

> Thank you Thierry for the comment. Sorry I didn't realize customizing
> `helm-man-format-switches' would work the moment I ran into the
> issue. You are right on this point.
>
>     IIRC Woman or man may fail due to a non existing or not valid file exists or
>     a file not supported by woman, in this case topic is used to allow
>     letting man find an alternative.
>     Can you describe the undesired behaviour ?
>
> woman' cannot decode the man page files on my mac and
> consequently Man-getpage-in-background' tries on CANDIDATE, i.e. the
> topic. However, it seems more friendly to me that before this last
> resort, we can take a chance on the file itself, thus the inner
> `condition-case'.

So ok for the additional `condition-case` for woman call, looks good.

I don't remember exactly the details, but I think I used
`Man-getpage-in-background` and not `man` directly, because `man` don't
work properly when called from lisp with `Man-notify-method`.
Perhaps there is anothers reason but I can't remember.  So I
would prefer you use for now `Man-getpage-in-background`.
In the first call use `manual-entry`.

So the only thing needed seems the additional condition-case.
While you are at it, can you can change original `(> (length wfiles) 1)`
which is ugly to `(cdr wfiles)` ?

Thanks.

--
Thierry

danielxia:
I've had your comments addressed in the latest commit.

Thanks.

helm-man.el
```diff
@@ -62,19 +62,19 @@ source.")
   "Default action for jumping to a woman or man page from helm."
   (let ((wfiles (mapcar #'car (woman-file-name-all-completions candidate))))
     (condition-case nil
-        (let ((file (if (> (length wfiles) 1)
+        (let ((file (if (cdr wfiles)
                         (helm-comp-read "ManFile: " wfiles :must-match t)
                       (car wfiles))))
           (if (eq helm-man-or-woman-function 'Man-getpage-in-background)
-              (man (format helm-man-format-switches file))
+              (manual-entry (format helm-man-format-switches file))
```
thierryvolpiatto: Isn't this simpler and doing the same thing:
```diff
@@ -69,3 +69,5 @@ source.")
-              (if (eq helm-man-or-woman-function 'Man-getpage-in-background)
-                  (manual-entry (format helm-man-format-switches file))
-                  (woman-find-file file)))
+              (if (eq helm-man-or-woman-function 'woman)
+                  (condition-case nil
+                      (woman-find-file file)
+                    (error nil))
+                  (manual-entry (format helm-man-format-switches file))))
```

thierryvolpiatto:
Once review applied, can you rebase ? And then we merge this change.
Thanks.

thierryvolpiatto:
Ah, doing many things at the same time is no good, may comment above is wrong, sorry, merging now, doesn't matter for the extras commits.
Thanks.

PR #1700, opened 2017-03-01
Improve docstring of `helm`

xiongtx:
The docstring contains:

```
\(&optional sources input prompt resume
            preselect buffer keymap default history\).
```

It's unclear whether `allow-nest` is accepted as a positional parameter. Seems strange that it's the only basic keyword parameter not also used as a positional parameter.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -1785,108 +1785,130 @@ only."
 (defun helm (&rest plist)
   "Main function to execute helm sources.
 
-Keywords supported:
-:sources :input :prompt :resume :preselect
-:buffer :keymap :default :history :allow-nest
+PLIST is a list like
 
-Extra LOCAL-VARS keywords are supported, see below.
+\(:key1 val1 :key2 val2 ...\)
 
-PLIST is a list like \(:key1 val1 :key2 val2 ...\) or
+ or
+
 \(&optional sources input prompt resume
             preselect buffer keymap default history\).
 
+** Keywords
+
+Keywords supported:
+
+- :sources
+- :input
+- :prompt
+- :resume
+- :preselect
+- :buffer
+- :keymap
+- :default
+- :history
+- :allow-nest
+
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
It is supported as positional parameter as well.
You have changed many things in this PR unrelated to this, please make you PR's focused on one thing at the time.
Some notes:
Describing that positional parameters are here just for backward compatibility and deprecated is good.
You have modified `:default`, removing some important stuff.
`:history` is wrong, well not really wrong but confusing.

xiongtx:
The PR is intended to overhaul the docstring of the `helm` command. The question about `allow-nest` as a position parameter is just a question, not the full intent of the PR.

I don't think anything important has been removed from the description of `default`:

Old description:

> A default argument that will be inserted in minibuffer \ with
> \\<minibuffer-local-map>\\[next-history-element]. When nil or not
> present `thing-at-point' will be used instead. If
> `helm--maybe-use-default-as-input' is non-`nil' display will be
> updated using :default arg as input unless :input is specified,
> which in this case will take precedence over :default. This is a
> string or a list. If list, car of the list becomes initial
> default input. \\<minibuffer-local-map>\\[next-history-element]

New description:

> Default value inserted into the minibuffer \ with
> \\<minibuffer-local-map>\\[next-history-element].
> 
> One of the following:
> 
> - String
> - List
>   - \\<minibuffer-local-map>\\[next-history-element] cycles through
> the list items, starting with the first.
> 
> If nil, `thing-at-point' is used.
> 
> If `helm--maybe-use-default-as-input' is non-`nil', display is
> updated using this value, unless :input is specified, in which
> case that value is used instead.

I have restored the description of `:history`.

`allow-nest` has been included as a positional parameter.

thierryvolpiatto:

> I don't think anything important has been removed from the description of default:

I badly read the diff so.

Thanks for your work.

-- 
Thierry

PR #1702, opened 2017-03-03
Fix #1701.

### LINKED ISSUE #1701: helm-for-files stat()s every local file in recentf
It seems that any time `helm-for-files` is invoked, it will invoke `file-attributes` for every file in the current directory and `recentf`. This causes some problems:

- If `recentf-max-saved-items` is large, Helm will still `stat` every item in it every time it is invoked, even if it will show a small subset of them.
- If some files reside on a mounted remote filesystem, the Helm invocation will be slow or even cause Emacs to hang, merely by being invoked, as `file-remote-p` can't tell whether a mounted filesystem is a network one.

Some ways in which this could be improved:

- Highlight file names asynchronously
- Only call `helm-highlight-files` for files that are immediately visible
- Cache the result of `helm-highlight-files` on a per-filename basis in a hashmap

For now, my workaround is to stub out `helm-highlight-files`:

```lisp
(defun helm-highlight-files (files)
  files)
```

This made `helm-for-files` instant again.

thierryvolpiatto:
Try to improve performances by using filtered-candidate-transformer instead of candidates-transformer in type-files.
Should compute much faster unless the value of candidates-number-limit is very hight.

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -282,6 +282,16 @@ in `current-buffer'."
   "Dired commands that are allowed moving to first real candidate."
   :group 'helm-files
   :type '(repeat (choice symbol)))
+
+(defcustom helm-mounted-network-directories nil
+  "A list of directories used for mounting remotes filesystem.
+
+When nil `helm-file-on-mounted-network-p' always return nil otherwise
+it checks if a file is in one of these directories.
+
+Remote filesystem are generally mounted with sshfs."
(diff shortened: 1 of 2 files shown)


CyberShadow:
Works great, thank you very much! Even without configuration I see a big improvement in performance, and `helm-mounted-network-directories` seems to work as expected.

Issue #1705, opened 2017-03-08
`helm-imenu-in-all-buffers` should have each buffer as separate source


xiongtx:
## Expected behavior

`helm-imenu-in-all-buffers` should have symbols from each buffer under a separate source, preferably with indication of the buffer's major mode.

## Actual behavior from emacs-helm.sh if possible

All symbols from all buffers are in a single source, with no indication as to where each came from.

E.g. the following symbols came from 3 buffers: 2 `enh-ruby-mode`, 1 `js-mode`. This is not at all clear.

![helm-imenu-same-source](https://cloud.githubusercontent.com/assets/1624090/23698967/52dae84c-03a2-11e7-99a4-f564625f9c14.png)

## Steps to reproduce (recipe)

Call `helm-imenu-in-all-buffers` with multiple buffers in the same major mode.

## Backtraces if some (M-x toggle-debug-on-error)

N/A

## Describe versions of helm, emacs, operating system etc.

- Helm: `2.5.1`
- Emacs: `25.1.1`
- OS: `Fedora 24`

thierryvolpiatto:
Tianxiang Xiong <notifications@github.com> writes:


> All symbols from all buffers are in a single source, with no
> indication as to where each came from.
>
> E.g. the following symbols came from 3 buffers: 2 enh-ruby-mode, 1
> js-mode. This is not at all clear.

You can use helm-popup-tip-mode.

--
Thierry

thierryvolpiatto:
![popup-tip-mode](https://cloud.githubusercontent.com/assets/1533939/23702874/7a57fea4-03fd-11e7-9f6a-bb47d18d11d0.png)

thierryvolpiatto:
Tianxiang Xiong <notifications@github.com> writes:

> Expected behavior
>
> helm-imenu-in-all-buffers should have symbols from each buffer under a
> separate source,

This have been addressed now, let me know if you are happy with this.

> preferably with indication of the buffer's major mode.

For this one it is easy to fix now, though I am still unsure how to
present it.

-- 
Thierry

xiongtx:
:+1: LGTM.

PR #1707, opened 2017-03-08
helm-imenu-in-all-buffers in separate sources (#1705).

thierryvolpiatto:
* helm-imenu.el (helm-imenu-in-all-buffers-separate-sources): New user var.
(helm-imenu-collect-sources-from-all-buffers): New fn.
(helm-imenu-in-all-buffers): Use it.

### DIFF
--- a/helm-imenu.el
+++ b/helm-imenu.el
@@ -178,45 +178,31 @@ only '((foo . bar)) is needed."
                         (delete (assoc "*Rescan*" index) index))))
           (setq helm-cached-imenu-tick tick))))))
 
-(defun helm-imenu-candidates-in-all-buffers ()
+(defun helm-imenu-candidates-in-all-buffers (&optional build-sources)
   (let* ((lst (buffer-list))
          (progress-reporter (make-progress-reporter
                              "Imenu indexing buffers..." 1 (length lst))))
     (prog1
-        (cl-loop for b in lst
-                 for count from 1
-                 when
-                 (and (with-current-buffer b
-                        (derived-mode-p 'prog-mode))
-                      (with-current-buffer b
-                        (helm-same-major-mode-p helm-current-buffer
-                                                helm-imenu-all-buffer-assoc)))
-                 do (progress-reporter-update progress-reporter count)
-                 and
-                 append (with-current-buffer b
-                          (helm-imenu-candidates b)))
-      (progress-reporter-done progress-reporter))))
-
-(defun helm-imenu-collect-sources-from-all-buffers ()
-  (let* ((lst (buffer-list))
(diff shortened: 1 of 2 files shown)


helm-imenu.el
```diff
@@ -191,6 +197,28 @@ only '((foo . bar)) is needed."
...
 
+(defun helm-imenu-collect-sources-from-all-buffers ()
+  (let* ((lst (buffer-list))
+         (progress-reporter (make-progress-reporter
+                             "Imenu indexing buffers..." 1 (length lst))))
+    (prog1
+        (cl-loop with cur-buf = (current-buffer)
+                 for b in lst
+                 for count from 1
+                 when (and (with-current-buffer b
```
xiongtx: We can use `with-current-buffer` outside the `and` to avoid repetition.

helm-imenu.el
```diff
@@ -270,22 +298,29 @@ only '((foo . bar)) is needed."
...
+              :candidates 'helm-imenu--in-all-buffers-cache
+              :fuzzy-match helm-imenu-fuzzy-match))))
   (let ((imenu-auto-rescan t)
         (str (thing-at-point 'symbol))
         (helm-execute-action-at-once-if-one
-         helm-imenu-execute-action-at-once-if-one))
-    (helm :sources 'helm-source-imenu-all
+         helm-imenu-execute-action-at-once-if-one)
+        (helm--maybe-use-default-as-input
+         (not (null (memq 'helm-source-imenu-all
```
xiongtx: Why is `(not (null...` necessary? Isn't a truthy value from `memq` sufficient?

helm-imenu.el
```diff
@@ -191,6 +197,28 @@ only '((foo . bar)) is needed."
                           (helm-imenu-candidates b)))
       (progress-reporter-done progress-reporter))))
 
+(defun helm-imenu-collect-sources-from-all-buffers ()
+  (let* ((lst (buffer-list))
+         (progress-reporter (make-progress-reporter
+                             "Imenu indexing buffers..." 1 (length lst))))
+    (prog1
+        (cl-loop with cur-buf = (current-buffer)
+                 for b in lst
```
xiongtx: Doesn't seem like `lst` is used anywhere else, so a call to `(buffer-list)` should be sufficient.

helm-imenu.el
thierryvolpiatto: `memq` is not returning a boolean so it is cleaner to return a boolean, though yes it would work also of course with a list instead of `t`.

thierryvolpiatto:

> We can use with-current-buffer outside the and to avoid repetition.

Yes ;-) done.

Thanks.

-- 
Thierry

PR #1710, opened 2017-03-14
Fix imenu transformer use customize variable handle types.

LiuPai:
* helm-imenu.el (helm-imenu-type-faces): New user var.
(helm-imenu-transformer): Use it.

### DIFF
--- a/helm-imenu.el
+++ b/helm-imenu.el
@@ -61,6 +61,20 @@ When nil all candidates are displayed in a single source."
   :type 'boolean
   :group 'helm-imenu)
 
+(defcustom helm-imenu-type-faces
+  '(("^Variables$" . font-lock-variable-name-face)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
刘湃 <notifications@github.com> writes:

>   * helm-imenu.el (helm-imenu-type-faces): New user var.
>     (helm-imenu-transformer): Use it.

Thanks, but I suggest you define instead three new faces, each one
inheriting from one of the font-lock* faces.
This avoid a new loop inside the inner one and simplify customization
also.

-- 
Thierry

LiuPai:
This patch allow people use different faces for different types, more than three fixed face.

thierryvolpiatto:
Oh, ok I see now, didn't fully read patch. Merging for now, but probably we should find something to not recompute face at each turn.
Thanks.

PR #1721, opened 2017-03-23
Don't widen indirect Org buffers when getting candidates

alphapapa:
* helm-org.el: (helm-org--get-candidates-in-buffer): Do it.

This way, if a buffer is created with org-tree-to-indirect-buffer, this
function will only return headings in that buffer's subtree.  In a
direct, narrowed buffer, all headings in the buffer will still be
returned.  This may not be a perfect solution, because if a user runs
`clone-indirect-buffer' on an entire Org buffer and then narrows it, he
might prefer this function to return all headings, but the former case
seems much more common and useful, so this behavior is probably best.

Thanks for your continued work on Helm, Thierry.

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -173,7 +173,10 @@ Note this have no effect in `helm-org-in-buffer-headings'."
                                   (apply old-fn args)))))
       (save-excursion
         (save-restriction
-          (widen)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks!

Issue #1442, opened 2016-03-23
Add `helm-version` command


xiongtx:
## Problem

It is not obvious, particularly for users who are not very familiar with Helm (or Emacs in general), what version of Helm is being used. This information is useful when reporting bugs, etc.

The answer can be found by looking at the `helm-pkg` file, but this many not be obvious to everyone.
## Suggested Solution:

Many packages have a `<package-name>-version` command. See `org-version` for an example.

Consider adding a similar command for Helm.

PR #1730, opened 2017-03-29
[#1442] Add `helm-version` var and command

xiongtx:
Address #1442.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -35,6 +35,23 @@
 (require 'helm-source)
 
 
+;;; Version
+
+(defconst helm-version "2.5.4")
+
+(defun helm-version (&optional interactivep)
+  "Return `helm-version' string.
+
+Optional parameter INTERACTIVEP is always t when called
+interactively.
+
+When called interactively, show version in echo area."
+  (interactive (list t))
+  (if interactivep
+      (message "Helm %s" helm-version)
+    helm-version))
+
+
 ;;; Multi keys
 ;;
 ;;


thierryvolpiatto:

>   * [#1442] Add `helm-version` var and command

I would prefer a command that extract the version number, see
https://github.com/thierryvolpiatto/emacs-tv-config/blob/master/init-helm-thierry.el#L40

-- 
Thierry

xiongtx:
When installing from MELPA, `helm-core-pkg.el`'s version number gets overwritten by the MELPA version, i.e. `"20170207.42"` instead of `"2.5.4"`.

I don't know which is more useful for bug reporting, but it's strange that `"2.5.4"` is nowhere to be found in  `helm-core` when installing from MELPA. In fact, you'll need to look in `helm-pkg.el` to find it.

thierryvolpiatto:

> When installing from MELPA, helm-core-pkg.el's version number gets
> overwritten by the MELPA version, i.e. "20170207.42" instead of
> "2.5.4".

This is the reason why I have not already installed helm/version.

> I don't know which is more useful for bug reporting,

The only useful thing is to have the revision used, but most users are
installing from Melpa and have not this information, and the Melpa
version is giving no clues of what have been exactly pushed to Melpa at
this time.

> but it's strange that "2.5.4" is nowhere to be found in helm-core when
> installing from MELPA. In fact, you'll need to look in helm-pkg.el to
> find it.

I don't want to maintain a helm-version variable in helm if this is what
you are asking.

--
Thierry

xiongtx:
Maybe just a `VERSION` file with a version string then?

xiongtx:
What about putting the version number in the package header with the `Version` keyword? 

That seems to be the standard practice, and there's even an [`lm-version`](https://github.com/bbatsov/emacs-lisp-style-guide/issues/35) function to support it.

thierryvolpiatto:

> What about putting the version number in the package header with the
> Version keyword?

We have two packages in one...
I don't like also the way package.el fetch information in source files
header, this is a practice of another age, the standard place should be
the *pkg.el files and they should be used instead of building another
*pkg.el file, I already submitted the idea to emacs but they still want
to use headers of source files.

> That seems to be the standard practice, and there's even an lm-version
> function to support it.

There is no standard practice actually to provide a version number in
packages.

--
Thierry

PR #1731, opened 2017-03-29
Ask user to install `lacarte` if not installed

xiongtx:
This makes for a better user experience, especially for new users.


thierryvolpiatto:
Thanks.

PR #1732, opened 2017-03-29
Preserve order when fuzzy matching in helm-recentf

alphapapa:
Currently, when helm-recentf-fuzzy-match is enabled, matches with equal fuzzy score (e.g. same filename, different path length) are sorted by path length, which partially defeats the purpose of recentf as they are no longer in order of recency.  With this commit, tied fuzzy matches in helm-recentf will preserve their original ordering by recency, making helm-recentf more useful.

* helm.el: helm-fuzzy-matching-sort-ties-by-length: Add var 
                (helm-fuzzy-matching-default-sort-fn-1): Do it.
* helm-files.el (helm-recentf): Do it.

Thanks, Thierry.

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -3845,9 +3845,10 @@ found."
 (defun helm-recentf ()
   "Preconfigured `helm' for `recentf'."
   (interactive)
-  (helm :sources 'helm-source-recentf
-        :ff-transformer-show-only-basename nil
-        :buffer "*helm recentf*"))
+  (let ((helm-fuzzy-matching-sort-ties-by-length nil))
+    (helm :sources 'helm-source-recentf
+          :ff-transformer-show-only-basename nil
+          :buffer "*helm recentf*")))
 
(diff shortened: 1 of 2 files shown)


alphapapa:
Forgive me, but I reconsidered my approach and pushed another branch which does the same thing but is organized slightly differently: https://github.com/emacs-helm/helm/compare/master...alphapapa:recentf-fuzzy-preserve-order-2?expand=1

I think the second branch is better, because it keeps the setting with the source rather than the command, so any command that uses the recentf source will benefit from it.  Since I've already pushed both and started this PR, I'll let you decide if you want one or the other (or neither, but I hope we can work something out, haha ;).

Thanks.

thierryvolpiatto:
Seems my previous mail didn't end up here, so yes I prefer the second branch.

alphapapa:
Ok, I'll send a PR for that one.  Thanks.

PR #1733, opened 2017-03-29
Preserve order when fuzzy matching in helm-recentf

alphapapa:
Currently, when helm-recentf-fuzzy-match is enabled, matches with equal
fuzzy score (e.g. same filename, different path length) are sorted by
path length, which partially defeats the purpose of recentf as they are
no longer in order of recency.  With this commit, tied fuzzy matches in
helm-recentf will preserve their original ordering by recency, making
helm-recentf more useful.

* helm.el: (helm-fuzzy-matching-default-sort-fn-1): Do it.
           (helm-fuzzy-matching-sort-fn-preserve-ties-order): Add function.
* helm-files.el (helm-recentf-fuzzy-match): Do it.

Thanks, Thierry.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -3131,15 +3131,16 @@ CANDIDATE. Contiguous matches get a coefficient of 2."
                              pat-lookup str-lookup :test 'equal))
                     2)))))
 
-(defun helm-fuzzy-matching-default-sort-fn-1 (candidates &optional use-real basename)
+(defun helm-fuzzy-matching-default-sort-fn-1 (candidates &optional use-real basename preserve-tie-order)
   "The transformer for sorting candidates in fuzzy matching.
 It sorts on the display part by default.
 
 Sorts CANDIDATES by their scores as calculated by
-`helm-score-candidate-for-pattern'. Ties in scores are sorted by
-length of the candidates. Set USE-REAL to non-`nil' to sort on the
-real part.  If BASENAME is non-nil assume we are completing filenames
-and sort on basename of candidates."
+`helm-score-candidate-for-pattern'.  Set USE-REAL to non-`nil' to
+sort on the real part.  If BASENAME is non-nil assume we are
+completing filenames and sort on basename of candidates.  If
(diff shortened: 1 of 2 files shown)


helm.el
```diff
@@ -3175,13 +3176,22 @@ and sort on basename of candidates."
...
+                       (unless preserve-tie-order
+                         (< len1 len2)))
                       ((> scr1 scr2)))))))))
 
 (defun helm-fuzzy-matching-default-sort-fn (candidates _source &optional use-real)
   "Default `filtered-candidate-transformer' to sort candidates in fuzzy matching."
   (helm-fuzzy-matching-default-sort-fn-1 candidates use-real))
 
+(defun helm-fuzzy-matching-sort-fn-preserve-ties-order (candidates _source &optional use-real)
+  "`filtered-candidate-transformer' to sort candidates in fuzzy matching, preserving order of ties.
```
xiongtx: Does this need to be a public fn? Can it be made private?

helm.el
```diff
@@ -3131,15 +3131,16 @@ CANDIDATE. Contiguous matches get a coefficient of 2."
                              pat-lookup str-lookup :test 'equal))
                     2)))))
 
-(defun helm-fuzzy-matching-default-sort-fn-1 (candidates &optional use-real basename)
+(defun helm-fuzzy-matching-default-sort-fn-1 (candidates &optional use-real basename preserve-tie-order)
```
xiongtx: Can be made private fn?

thierryvolpiatto: It could be reused in many other places to define new sort command (like in locate).

helm.el
thierryvolpiatto: why ?

thierryvolpiatto:
Thanks.

helm.el
xiongtx: :man_shrugging: I like to keep interfaces (i.e. public fns, vars, etc.) as small as possible. Makes code easier to refactor.

PR #1734, opened 2017-03-30
Add `refresh` option to `helm-info`

xiongtx:
Address [Emacs StackExchange issue](http://emacs.stackexchange.com/q/30040/10269).


thierryvolpiatto:
Thanks.

PR #1741, opened 2017-04-10
Fix marking last candidate in a source

alphapapa:
* helm.el (helm-make-visible-mark): Do it.

If the user marked the last candidate in a source, with another source listed afterward, too much would be marked.  Now we stop the mark where the current source ends.

Thanks.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -5384,11 +5384,13 @@ Meaning of prefix ARG is the same as in `reposition-window'."
 (defun helm-make-visible-mark (&optional src selection)
   (let* ((source (or src  (helm-get-current-source)))
          (sel    (or selection (helm-get-selection nil nil src)))
-         (o (make-overlay (point-at-bol)
-                          (if (helm-pos-multiline-p)
-                              (or (helm-get-next-candidate-separator-pos)
-                                  (point-max))
-                              (1+ (point-at-eol))))))
+         (selection-end (if (helm-pos-multiline-p)
+                            (or (helm-get-next-candidate-separator-pos)  ; Stays within source
+                                (helm-get-next-header-pos)
+                                (point-max))
+                          ;; Not multiline
+                          (1+ (point-at-eol))))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:

>   * helm.el (helm-make-visible-mark): Do it.
>
> If the user marked the last candidate in a source, with another source
> listed afterward, too much would be marked. Now we stop the mark where
> the current source ends.

Sorry but I don't understand what you want to do here, can you develop ?

Thanks.

--
Thierry

alphapapa:
There seems to be a bug in which, if multiple, multi-line sources are displayed, and the user marks the last candidate in a source, and there is another source containing a single candidate after that source, the candidate in the next source will be marked also.  For example:

```
SOURCE A HEADER
Candidate 1 first line
candidate 1 second line
----
Candidate 2 first line
candidate 2 second line
SOURCE B HEADER
Candidate 3 first line
candidate 3 second line
```

If the user marks candidate 2, it will also mark the source B header and candidate 3.

In fact, it seems to happen across multiple single-candidate sources, e.g.

```
SOURCE A HEADER
Candidate 1 first line
candidate 1 second line
----
Candidate 2 first line
candidate 2 second line
SOURCE B HEADER
Candidate 3 first line
candidate 3 second line
SOURCE C HEADER
Candidate 4 first line
candidate 4 second line
```

If the user marks candidate 2, it will also mark the source B header, candidate 3, the source C header, and candidate 4.

This PR fixes this bug.

thierryvolpiatto:
Ok, I get it, here a recipe:

```elisp
(helm :sources `(,(helm-build-sync-source "test1"
                                          :candidates '("foo1" "bar1" "baz1")
                                          :multiline t)
                 ,(helm-build-sync-source "test"
                                          :candidates '("foo" "bar" "baz")))
      :buffer "*helm test*")
```
Thanks to catch this.

PR #1743, opened 2017-04-13
Add Org file bookmarks to helm-bookmarks

alphapapa:
This adds support for bookmarked Org files to `helm-bookmark`.  They no longer appear as uncategorized bookmarks.

Thanks.


thierryvolpiatto:
Thanks!

PR #1751, opened 2017-04-20
Add helm-wikipedia-show-summary

alphapapa:
This adds a new action based on helm-wikipedia-persistent-action that exits Helm and shows an article summary in a permanent buffer.

Thanks.

### DIFF
--- a/helm-net.el
+++ b/helm-net.el
@@ -277,37 +277,52 @@ Can be \"-new-tab\" (default) or \"-new-window\"."
                                           helm-pattern)))))))
 
 (defvar helm-wikipedia--summary-cache (make-hash-table :test 'equal))
+(defun helm-wikipedia-show-summary (input)
+  "Show Wikipedia summary for INPUT in new buffer."
+  (interactive)
+  (let ((buffer (get-buffer-create "*helm wikipedia summary*"))
+        (summary (helm-wikipedia--get-summary input)))
+    (with-current-buffer buffer
+      (visual-line-mode)
+      (erase-buffer)
+      (insert summary)
+      (pop-to-buffer (current-buffer)))))
+
 (defun helm-wikipedia-persistent-action (candidate)
   (unless (string= (format "Search for '%s' on wikipedia"
                            helm-pattern)
                    (helm-get-selection nil t))
     (message "Fetching summary from Wikipedia...")
     (let ((buf (get-buffer-create "*helm wikipedia summary*"))
-          result mess)
-      (while (progn
-               (setq result (or (gethash candidate helm-wikipedia--summary-cache)
-                                (puthash candidate
-                                         (prog1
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks.

thierryvolpiatto:
I have reverted, it is not compiling, please fixit and submit again, thanks.

xiongtx:
Looks nice, but should Wikipedia-related stuff be part of Helm instead of a separate package? 

I view `helm` at 3 levels:

- `helm-core`, the library for Helm features
- `helm`, the package that improves *existing* Emacs functionality (e.g. completion, find files, etc.)
- Other Helm packages, which use `helm-core` and `helm` to add *new* functionality

alphapapa:
> I have reverted, it is not compiling, please fixit and submit again, thanks.

Sorry, I had the keymap defined after the source that used it.  Fixed and resubmitting.

BTW, maybe there should be a note on the wiki about being sure to run `make` before submitting PRs.  I tried byte-compiling the file and the whole directory from within Emacs, but I only found the error by running `make`.  :)

> Looks nice, but should Wikipedia-related stuff be part of Helm instead of a separate package?

I think it should be, yes.  I view Helm at 3 levels, but slightly differently:

* `helm-core`: the core library
* `helm`: a kind of metapackage that provides commonly useful commands built on the library
* Third-party packages that use Helm (e.g. my own `helm-org-rifle`)

One of the benefits of Helm, despite the fact that some people don't like it, is that it provides so many tools in one box.  Just today I found a new tool in Helm that solved a problem I had in a matter of seconds: `helm-timers`.  For some reason, a `bibtex`-related timer has been set in my Emacs config, and it is failing every time Emacs goes idle for a few seconds.  I haven't been able to find out what is setting the timer.  I was getting annoyed with it, so I ran `helm-M-x`, typed `timer` to see what timer-related commands I could find, and up popped `helm-timers`!  I was able to immediately see the bad timer and delete it with a keystroke!  Problem solved!

Now, if I had had to bring up the package list and look for some `helm-timers` package and read about it and install it and load it and run it...that would have taken a lot more time and effort.

Really, `helm-wikipedia` is a few functions, and it's very useful and quick.  It doesn't need to be moved to its own package.  It's fine.  :)

thierryvolpiatto:

>     I have reverted, it is not compiling, please fixit and submit again, thanks.
>
> Sorry, I had the keymap defined after the source that used it.

That's why it is important to compile helm OUTSIDE of your current
emacs, see below.

> BTW, maybe there should be a note on the wiki about being sure to run
> make before submitting PRs. I tried byte-compiling the file and the
> whole directory from within Emacs,

If you want to do this, use
(async-byte-recompile-directory "/your/helm/directory")
but don't compile your directory from your running emacs, never.

> but I only found the error by running make. :)

Yes, using make is the easiest way to compile helm properly.

--
Thierry

PR #1763, opened 2017-05-03
Change helm-net HTTP URLs to HTTPS

notetiene:
* helm-net.el (helm-home-url): Change the TLD to `com` since Google
  automatically handles region version.
(helm-google-suggest-search-url): Use the encrypted version of Google
Search for better privacy.
(helm-wikipedia-suggest-url): Add string format.
(helm-wikipedia-summary-url): Change URL scheme to HTTPS and add
string format.
(helm-search-suggest-action-youtube-url):
(helm-search-suggest-action-google-maps-url):
(helm-search-suggest-action-google-news-url): Change URL scheme to HTTPS.
(helm-google-suggest-fetch):
(helm-wikipedia-suggest-fetch):
(helm-wikipedia--get-summary): Use format instead of concat.

### DIFF
--- a/helm-net.el
+++ b/helm-net.el
@@ -36,7 +36,7 @@ See `helm-browse-url-default-browser-alist'."
   :group 'helm-net
   :type 'symbol)
 
-(defcustom helm-home-url "http://www.google.fr"
+(defcustom helm-home-url "https://www.google.com"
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks!

PR #1774, opened 2017-05-22
* helm-command.el (helm-M-x): Re-throw error signal.

xuchunyang:
Otherwise, when an error occurs during command execution, the user will not get any feedback. For example,

```el
(defun foo ()
  (interactive)
  (+ 1 "string"))
```


thierryvolpiatto:
Thanks!

Issue #1775, opened 2017-05-22
Hide leading stars in helm-source-org-headings-for-files


jagrg:
If org-hide-leading-stars is t, or org-indent-mode is on, I think it would make sense to also hide the first N-1 stars in the helm-org buffer. This is what I tried:

```
(defun helm-source-org-headings-for-files (filenames &optional parents)
  (helm-make-source "Org Headings" 'helm-org-headings-class
    :filtered-candidate-transformer 'helm-org-hide-leading-stars
    :parents parents
    :candidates filenames))

(defun helm-org-hide-leading-stars (candidates _source)
  (cl-loop for i in candidates
	   collect
	   (cons
	    (when (string-match "^\\(\\**\\)\\(\\* \\)\\(.*\n?\\)" (car i))
	      (replace-match "\\1\\2\\3" nil nil (car i)))
	    (cdr i))))
```

Now:

1. Open an Org buffer with a few subtrees.
2. M-x org-indent-mode
3. Set helm-org-headings-fontify to t
3. M-x helm-org-in-buffer-headings

The problem is it only works with helm-org-headings-fontify set to t. How can we do the same with helm-org-headings-fontify set to nil?

jagrg:
I think I found a solution. Would you accept a PR?

```
(defun helm-org-hide-leading-stars (candidates _source)
  (cl-loop for i in candidates
	   collect
	   (cons
	    (if helm-org-headings-fontify
		(when (string-match "^\\(\\**\\)\\(\\* \\)\\(.*\n?\\)" (car i))
		  (replace-match "\\1\\2\\3" nil nil (car i)))
	      (when (string-match "^\\(\\**\\)\\(\\* \\)\\(.*\n?\\)" (car i))
		(let ((foreground (org-find-invisible-foreground)))
		  (if foreground
		      (concat
		       (propertize (replace-match "\\1" nil nil (car i))
				   'font-lock-face `(:foreground ,foreground))
		       (replace-match "\\2\\3" nil nil (car i)))
		    (replace-match "\\1\\2\\3" nil nil (car i))))))
	    (cdr i))))
```

thierryvolpiatto:
Jonathan Gregory <notifications@github.com> writes:

> I think I found a solution. Would you accept a PR?

I have not a strong opinion on this as my org usage is quite basic (I
never heard about org-indent-mode before), so if you think this is a
good addition and it is not breaking what we had before, yes I accept a
PR.

Thanks.

-- 
Thierry
[...]

PR #1777, opened 2017-05-24
Use Org's startup visibility in Helm

jagrg:
Here's an updated version of the code discussed in https://github.com/emacs-helm/helm/issues/1775. It should indent org headings and hide leading stars according to the user's settings. To test:

```
#+STARTUP: indent hidestars
# #+STARTUP: noindent hidestars
# #+STARTUP: noindent showstars

* Heading A
** Subheading A.1
*** Subsubheading A.1.1
**** Subsubsubheading A.1.1.1
*** Subsubheading A.1.2
** Subheading A.2
* Heading B
```

1. Press C-c C-c twice on the #+STARTUP line to refresh
2. M-x helm-org-in-buffer-headings RET C-g
3. Comment that line and uncomment the next line
4. Repeat steps 1, 2 and 3
5. Change the value of helm-org-headings-fontify and repeat steps 1-4

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -142,8 +142,49 @@ Note this have no effect in `helm-org-in-buffer-headings'."
 
 (defun helm-source-org-headings-for-files (filenames &optional parents)
   (helm-make-source "Org Headings" 'helm-org-headings-class
+    :filtered-candidate-transformer 'helm-org-startup-visibility
     :parents parents
     :candidates filenames))
+
+(defun helm-org-startup-visibility (candidates _source)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks.

Just a question, if one of the `(when (string-match [...]` fails, what happen ?
We may end with `(nil . real)` candidates isn't it ?
If so perhaps this should be protected by calling it earlier and provide an alternative ?

jagrg:
I don't think there's any reason to fail. The candidates are all headings anyway.

thierryvolpiatto:
Ok, so merging, thanks!

PR #1778, opened 2017-05-25
Check variables before making transformation

jagrg:
To reproduce the problem:

1. Eval: (setq helm-org-show-filename t helm-org-format-outline-path t)
2. M-x helm-org-agenda-files-headings


thierryvolpiatto:
Thanks.

PR #1792, opened 2017-06-16
regexp quote default string used by imenu

lewang:
- A symbon can contain valid regexp characters e.g. "*", which won't match
  because of the "\\_<" and "\\_>" anchors used.

### DIFF
--- a/helm-imenu.el
+++ b/helm-imenu.el
@@ -305,7 +305,7 @@ Each car is a regexp match pattern of the imenu type string."
         (helm-execute-action-at-once-if-one
          helm-imenu-execute-action-at-once-if-one))
     (helm :sources 'helm-source-imenu
-          :default (list (concat "\\_<" str "\\_>") str)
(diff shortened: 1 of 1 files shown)


lewang:
As a side question I tried out what format `:default` is supposed to take but couldn't i.e. https://github.com/emacs-helm/helm/blob/master/helm-imenu.el#L341

thierryvolpiatto:
Thanks!

thierryvolpiatto:
Le Wang <notifications@github.com> writes:

> As a side question I tried out what format :default is supposed to
> take but couldn't

See `helm` docstring.

If you don't specify :default in your `helm` call, `(thing-at-point
'symbol)` is used.

If your source is member of `helm-sources-using-default-as-input`
:default is used immediately and helm starts searching with this
:default (The minibuffer stays empty letting user starts another search
if needed), otherwise you have to enter :default with M-n in minibuffer.

--
Thierry

PR #1793, opened 2017-06-16
follow up to #1792

lewang:
that one broke the `nil` case.


thierryvolpiatto:
Fixed conflict and merged, thanks!

Issue #1806, opened 2017-07-03
Display "no results" message when no candidates found in multiple-source command


alphapapa:
Maybe I'm missing something, but I can't seem to find a way to display a "Results not found" message when there are no candidates found in a Helm command with multiple sources.  With a single source, the source itself can return a single-candidate list containing such a message, but with multiple sources, that results in a "Results not found" displayed for every source, rather than a single "Results not found".  I tried using `:redisplay` for this, but my sources are not async, so it doesn't work (I guess that is why it doesn't work).

The reason I'd like to have this feature is that, without it, there is no feedback for the user, so he doesn't know if there are no results, or if the sources haven't finished searching yet.  For example, using `helm-org-rifle-org-directory` (with multiple files in that directory), if you type in a string of gibberish, the only way to know that it's finished searching for results is to look at Emacs' CPU usage: when it stops running at 100%, you know it's done, and if there are no candidates displayed, you know it found none.  This isn't very user-friendly, so I'd like to be able to display a single "No results found" candidate (not one for each Helm source in the command; if there are results in some sources but not others, the sources without results are not displayed at all).

Thanks, Thierry.  :)

thierryvolpiatto:
```elisp
(helm :sources `(,(helm-build-sync-source "test"
                    :candidates '("foo" "bar" "baz")
                    :filtered-candidate-transformer (lambda (candidates _source)
                                                      (or candidates (list "No candidates"))))
                 ,(helm-build-sync-source "test"
                    :candidates '("foo" "bar" "baz")
                    :filtered-candidate-transformer (lambda (candidates _source)
                                                      (or candidates (list "No candidates"))))
                 ,(helm-build-sync-source "test"
                    :candidates '("foo" "bar" "baz")
                    :filtered-candidate-transformer (lambda (candidates _source)
                                                      (or candidates (list "No candidates")))))
      :buffer "*helm test*")
```

![capture d ecran_2017-07-05_08-38-02](https://user-images.githubusercontent.com/1533939/27851961-63b530fa-615d-11e7-99b7-70dd54de366c.png)

![capture d ecran_2017-07-05_08-38-21](https://user-images.githubusercontent.com/1533939/27851971-6e625d7a-615d-11e7-9106-270bf7df8e27.png)

thierryvolpiatto:
Starting from the recipe above, what would you want to have ?

- helm exit with a message "nothing found"
- don't display sources, but just an empty helm buffer with a message "nothing found"
[...]

PR #1816, opened 2017-07-12
No results found proof-of-concept

alphapapa:
As mentioned in #1806.  In case it's easier for you, here's a temporary PR.  Thanks.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -3669,22 +3669,29 @@ without recomputing them, it should be a list of lists."
                      (setq matches (or candidates (helm--collect-matches sources))))
            ;; If computing matches finished and is not interrupted
            ;; erase the helm-buffer and render results (Fix #1157).
-           (when matches
-             (erase-buffer)             ; [1]
-             (cl-loop for src in sources
-                      for mtc in matches
-                      do (helm-render-source src mtc))
-             ;; Move to first line only when there is matches
-             ;; to avoid cursor moving upside down (issue #1703).
-             (helm--update-move-first-line)
-             (helm--reset-update-flag)))
-      (let ((src (or source (helm-get-current-source))))
-        (unless (assq 'candidates-process src)
-          (helm-display-mode-line src)
-          (helm-log-run-hook 'helm-after-update-hook)))
-      (when preselect
-        (helm-log "Update preselect candidate %s" preselect)
-        (helm-preselect preselect source))
+           (erase-buffer)             ; [1]
+           (if (cl-loop for source in matches
+                       thereis source)
+              (progn
+                (helm-log "Matches found: %S" matches)
+                (cl-loop for src in sources
+                         for mtc in matches
+                         do (helm-render-source src mtc))
+                ;; Move to first line only when there is matches
+                ;; to avoid cursor moving upside down (issue #1703).
+                (helm--update-move-first-line)
+                ;; Not sure if this should be outside the progn
+                (let ((src (or source (helm-get-current-source))))
+                  (unless (assq 'candidates-process src)
+                    (helm-display-mode-line src)
+                    (helm-log-run-hook 'helm-after-update-hook)))
+                (when preselect
+                  (helm-log "Update preselect candidate %s" preselect)
+                  (helm-preselect preselect source)))
+            ;; No matches; show results-not-found if input given
+             (when (not (string-empty-p helm-input))
+              (helm-log "No matches for input: %S" helm-input)
+              (helm-insert-match "No results found." 'insert))))
       (setq helm--force-updating-p nil))
     (helm-log "end update")))
 
--- a/helm.el
+++ b/helm.el
@@ -3667,31 +3667,32 @@ without recomputing them, it should be a list of lists."
            ;; updating.
            (helm-log "Matches: %S"
                      (setq matches (or candidates (helm--collect-matches sources))))
-           ;; If computing matches finished and is not interrupted
-           ;; erase the helm-buffer and render results (Fix #1157).
-           (erase-buffer)             ; [1]
-           (if (cl-loop for source in matches
-                       thereis source)
-              (progn
-                (helm-log "Matches found: %S" matches)
-                (cl-loop for src in sources
-                         for mtc in matches
-                         do (helm-render-source src mtc))
-                ;; Move to first line only when there is matches
-                ;; to avoid cursor moving upside down (issue #1703).
-                (helm--update-move-first-line)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks, I will have a look ASAP.

helm.el
```diff
@@ -3667,24 +3667,32 @@ without recomputing them, it should be a list of lists."
...
+                   ;; Not sure if this should be outside the progn
+                   (let ((src (or source (helm-get-current-source))))
+                     (unless (assq 'candidates-process src)
+                       (helm-display-mode-line src)
+                       (helm-log-run-hook 'helm-after-update-hook)))
+                   (when preselect
+                     (helm-log "Update preselect candidate %s" preselect)
+                     (helm-preselect preselect source)))
+               ;; No matches; show results-not-found if input given
+               (when (not (string-empty-p helm-input))
```
xiongtx: `(when (not ...` => `unless`

helm.el
```diff
@@ -3667,24 +3667,32 @@ without recomputing them, it should be a list of lists."
...
-             (helm--update-move-first-line)
-             (helm--reset-update-flag)))
-      (let ((src (or source (helm-get-current-source))))
-        (unless (assq 'candidates-process src)
-          (helm-display-mode-line src)
-          (helm-log-run-hook 'helm-after-update-hook)))
-      (when preselect
-        (helm-log "Update preselect candidate %s" preselect)
-        (helm-preselect preselect source))
+             (if (cl-loop for source in matches
```
xiongtx: This `cl-loop` seems like it can be replaced by `(cl-notany null ...)`.

helm.el
```diff
@@ -3667,24 +3667,32 @@ without recomputing them, it should be a list of lists."
...
+                          thereis source)
+                 (progn
```
xiongtx: This looks like it'd be better off as the `else` part of the `if`.

helm.el
```diff
@@ -3667,31 +3667,32 @@ without recomputing them, it should be a list of lists."
            ;; updating.
            (helm-log "Matches: %S"
                      (setq matches (or candidates (helm--collect-matches sources))))
-           ;; If computing matches finished and is not interrupted
-           ;; erase the helm-buffer and render results (Fix #1157).
-           (erase-buffer)             ; [1]
-           (if (cl-loop for source in matches
-                       thereis source)
```
thierryvolpiatto: This loop is redundant with `(when matches`, it does the same, also it use the name source which is confusing because there is no source in matches.

helm.el
```diff
@@ -3667,31 +3667,32 @@ without recomputing them, it should be a list of lists."
...
+                            for mtc in matches
+                            do (helm-render-source src mtc))
+                   ;; Move to first line only when there is matches
+                   ;; to avoid cursor moving upside down (issue #1703).
+                   (helm--update-move-first-line)
+                   ;; Not sure if this should be outside the progn
+                   (let ((src (or source (helm-get-current-source))))
+                     (unless (assq 'candidates-process src)
+                       (helm-display-mode-line src)
+                       (helm-log-run-hook 'helm-after-update-hook)))
```
thierryvolpiatto: All this block runs out of context, it is why the `helm/noresult` example is running too early when you tried it.

helm.el
thierryvolpiatto: I try generally to avoid cl functions in helm prefering cl-loop which is faster and easier to read.
That said, as I commented above whatever you use here is redundant with `(when matches` i.e have the same effect and should be removed.

helm.el
```diff
@@ -3667,31 +3667,32 @@ without recomputing them, it should be a list of lists."
...
+                          thereis source)
+                 (progn
+                   (helm-log "Matches found: %S" matches)
+                   (cl-loop for src in sources
+                            for mtc in matches
+                            do (helm-render-source src mtc))
+                   ;; Move to first line only when there is matches
+                   ;; to avoid cursor moving upside down (issue #1703).
+                   (helm--update-move-first-line)
+                   ;; Not sure if this should be outside the progn
```
thierryvolpiatto: Yes that should go in the unwind-protect undwindform. See comment below.

thierryvolpiatto:
I have removed stylistic comments to makes clear the important comments.

So as I said in some previous post, the example I sent with helm/noresult and helm/test, is not fully working because I guess you tried it with your helm-update version loaded which doesn't run the after-update-hook in the right place.
So please try it again with the master branch loaded.
You can perhaps use the mode-line to notify "no results", it avoid the side effects of adding a false candidate "no results" in helm-buffer and doesn't have the inconvenients of message.
```elisp
(defun helm/noresult ()
  (when (helm-empty-buffer-p)
    (with-helm-buffer
      (setq mode-line-format
            '(" " mode-line-buffer-identification " "
              (:eval (propertize "[No results]" 'face 'helm-grep-finish)))))))

(defun helm/test ()
  (interactive)
  (add-hook 'helm-after-update-hook 'helm/noresult)
  (unwind-protect
      (helm :sources `(,(helm-build-sync-source "test1"
                          :candidates '("foo" "bar" "baz"))
                       ,(helm-build-sync-source "test2"
                          :candidates '("foo" "bar" "baz"))
                       ,(helm-build-sync-source "test3"
                          :candidates '("foo" "bar" "baz")))
            :buffer "*helm test*")
    (remove-hook 'helm-after-update-hook 'helm/noresult)))
```

helm.el
alphapapa: I had to add this loop because `matches` was returning `'(nil nil nil ...)` and evaluating to non-nil.

alphapapa:
Thanks, Thierry, I'll give this a try soon.

thierryvolpiatto:

> I had to add this loop because matches was returning '(nil nil nil ...) and evaluating to non-nil.

Thanks, I will look into this.

-- 
Thierry

thierryvolpiatto:

> I had to add this loop because matches was returning '(nil nil nil
> ...) and evaluating to non-nil.

Thanks to catch this, I have removed the `(when matches` which obviously
returns always non-nil unless sources are nil.

--
Thierry

thierryvolpiatto:
I could finally remember why I was using `(when matches` which returns as you said a list of results which are maybe nil, however it returns a plain nil when interrupted by helm-while-no-input.
Thanks to point me on this, it allowed me to clarify this point.
For your "no results" notification, I think it should work in the helm-after-update-hook as described in #1806 so I am closing this now.

Issue #1641, opened 2016-11-16
helm-M-x kill-emacs interaction with i3 window manager


plantarum:
I have come across a very unusual interaction between helm and the i3 window manager. Using `helm-M-x` to invoke `kill-emacs` causes the keyboard input to be ignored until the mouse is used to switch windows. It's probably a very local issue, but since I can reproduce it without fail, I'm reporting it in case anyone else encounters this. 

## Expected behavior

`helm-M-x` -> `kill-emacs`, emacs closes and focus returns to the window manager.

## Actual behavior from emacs-helm.sh if possible

1. Start emacs from a terminal (but a GUI version, not emacs -nw) via `emacs-helm.sh`
2. `helm-M-x` `kill-emacs`
3. Emacs closes, but keyboard is non-responsive
4. If another window is open, or another workspace is in use, after selecting the other window or workspace with the mouse the keyboard will work again.
5. If no other window or workspace is available to be selected by the mouse, the only way to regain the functioning of the keyboard is to switch to another virtual terminal (i.e., Ctrl-Alt-F1) and restart the window manager. 

## Steps to reproduce (recipe)

Sorry, not sure how 'actual behaviour' and 'steps to reproduce' are not the same thing?

## Backtraces if some (M-x toggle-debug-on-error)

NA - no actual error within Emacs, but rather a problem when returning focus to the OS

## Describe versions of helm, emacs, operating system etc.
Emacs 25.1.50.2
Helm 20161113.908
Debian Linux
i3 version 4.13

Things I've tested:
- no problem when using other window managers
- no problem when starting helm via emacs-helm.sh, but calling `kill-emacs` via regular `M-x` instead of `helm-M-x`

Now that I've narrowed down the problem to a reproducible recipe for my system, this is easy enough to avoid. However, if anyone has any ideas *why* this happens, it might help me with some other weird behaviours I'm experiencing with my window manager freezing up.

xuchunyang:
I can't reproduce on Ubuntu with the following steps:

1. Login with i3 and its default config
2. Open terminal and cd to helm directory
3. `./emacs-helm.sh` (it opens a graphic Emacs and input focus goes to that window)
4. `helm-M-x kill-emacs` (it closes Emacs window and input focus goes to the terminal window)

---

- Ubuntu 16.04
- i3 4.11-1 from Ubuntu's official repo
- helm git repo and 20161109.1317
- Emacs 26.0.50.3

xuchunyang:
[...]

PR #1818, opened 2017-07-16
Do not call x-focus-frame that confuses the buggy i3 wm (#1641)

errge:
Sorry, but the i3 maintainers decided to play politics and to not even review my changelist from a technical point of view, only from standard committee point of view.  They also told me to get rid of XSetInputFocus from emacs, which is obviously ridiculous, if the rule would be that only WMs are calling that, then the X server would refuse those calls or have an option at least to refuse those calls.

I also understand the issue in #1641 well enough now, to say that it shouldn't happen with most of the other tiling window managers, as i3 contains a bug that is specific to i3.

Fortunately for us, we actually didn't want to call XSetInputFocus, we only called it by accident, so my changelist simply gets rid of that accidental call and fixes the issue that way.

I was able to reproduce the issue with 100% certainty and I can confirm that this change fixes it 100% on my setup.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -2446,9 +2446,29 @@ frame configuration as per `helm-save-configuration-functions'."
                ;; This is needed for minibuffer own-frame config
                ;; when recursive minibuffers are in use.
                ;; e.g M-: + helm-minibuffer-history.
-               (let ((frame (if (minibufferp helm-current-buffer)
+               (cl-letf ((frame (if (minibufferp helm-current-buffer)
                                 (selected-frame)
-                              (last-nonminibuffer-frame))))
+                                (last-nonminibuffer-frame)))
+                         ;; This is a workaround, because the i3 window
+                         ;; manager developers are refusing to fix their
+                         ;; broken timestamp and event handling.
+                         ;;
+                         ;; We basically just disable the part of
+                         ;; select-frame-set-input-focus that would call
+                         ;; XSetInputFocus in Xlib (x-focus-frame), that
+                         ;; resets a timestamp in the xserver which the i3
+                         ;; developers fail to notice.
+                         ;;
+                         ;; Since they don't know about the new timestamp,
+                         ;; their keyboard handling can break after a helm
+                         ;; user quits emacs, as reported in #1641.
+                         ;;
+                         ;; Fortunately for us, we really don't need this
+                         ;; XSetInputFocus call, since we already have focus
+                         ;; for Emacs, the user is just using helm!  We call
+                         ;; select-frame-set-input-focus for the other
+                         ;; side-effects, not for x-focus-frame.
+                         ((symbol-function 'x-focus-frame) #'ignore))
                  (select-frame-set-input-focus frame))))))
 
 (defun helm-split-window-default-fn (window)


errge:
Sorry for the noise, but my fix is not correct yet, I made a mistake, working on it right now...

errge:
Okay, my change is ready to review.

Please note that I had to add a dependency on cl for the flet.  Is this a problem?  If yes, then we can come up with other solutions (e.g. just writing our own select-frame-input-focus without the offending x-focus-frame call).

If flet and cl is not allowed, can you give me some input on why this select-frame-input-focus call is necessary, so I can understand the situation better and come up with a way to fix this?

errge:
Sorry for the noise...

I got rid of the cl dependency, now my changeset only depends on pure emacs lisp and should work in every major emacs version without additional dependencies.

helm.el
```diff
@@ -2449,7 +2449,31 @@ frame configuration as per `helm-save-configuration-functions'."
...
+                 ;; XSetInputFocus call, since we already have focus
+                 ;; for Emacs, the user is just using helm!  We call
+                 ;; select-frame-set-input-focus for the other
+                 ;; side-effects, not for x-focus-frame.
+                 (let ((saved-x-focus-frame (symbol-function 'x-focus-frame)))
+                   (unwind-protect
+                       (progn
+                         (fset 'x-focus-frame #'(lambda (_)))
+                         (select-frame-set-input-focus frame))
+                     (fset 'x-focus-frame saved-x-focus-frame))))))))
```
thierryvolpiatto: IIUC, you want to disable `x-focus-frame` when calling `select-frame-set-input-focus` ? Right?
If so you can use this instead:

```elisp
(cl-letf (((symbol-function 'x-focus-frame) #'ignore))
  (Do_what_you_have_to_do_here))
```

thierryvolpiatto: Something like this:

```diff
@@ -2449,3 +2449,4 @@ frame configuration as per `helm-save-configuration-functions'."
-               (let ((frame (if (minibufferp helm-current-buffer)
-                                (selected-frame)
-                              (last-nonminibuffer-frame))))
+               (cl-letf ((frame (if (minibufferp helm-current-buffer)
+                                    (selected-frame)
+                                  (last-nonminibuffer-frame)))
+                         ((symbol-function 'x-focus-frame) #'ignore))
```

thierryvolpiatto:
Thanks to work on this, my knowledge on X server is too limited to comment on `XSetInputFocus` though.
It looks like `x-focus-frame` is not supported by i3 IIUC, probably it would be better to fix this directly in `select-frame-set-input-focus`, but it would take ages and interminable discussion on emacs list, if you are patient enough you can report there.
In the meaning time, let's apply your workaround here in helm.
The new `cl-flet` of cl-lib is much better than `flet` from cl but works only lexically, in our case that's mean that if `x-focus-frame` doesn't appear physically in BODY it will have no effect, in this case the best is to use `cl-letf` (see my comments in diff).

errge:
Hey,

Yes, understanding the XSetInputFocus interaction with all the XCB async usage and X timestamping logic takes a big discussion, which I think should save you, your short summary of "not supported by i3" I think is a good tldr.

You are also right about Emacs, maybe we should propose an extra argument to `select-frame-set-input-focus`, but since it's only 1 wm in the world, I'm not sure how long and time consuming discussion would that be.

Regarding cl-letf:

    (defun x () 42)
    (defun y () (x))
    (flet ((x () 4242)) (y)) ;; 4242, but in 'cl
    (cl-flet ((x () 4242)) (y)) ;; 42, since it's lexical
    (letf (((symbol-function 'x) (lambda () 4242))) (y)) ;; 4242, but in 'cl
    (cl-letf (((symbol-function 'x) (lambda () 4242))) (y)) ;; 4242

Which shows that you are 100% correct, I don't know how I didn't think of it alone.  I had `cl-letf` in my sight, but I failed to combine it with `symbol-function`.  The generated macro code is the same as mine, so thanks for pointing this out, I will fix the pull request and ping you.

Cheers!

errge:
I have updated the pull request.

thierryvolpiatto:
Great!
I have no access to i3 wm, if you think your changes are fixing #1641 feel free to close it as well.
Thanks.

errge:
@thierryvolpiatto Yes, it's fixed, but I don't have a close button on the github ui, maybe only you can close it.

Please close it, i'm pretty sure it's fixed now.

thierryvolpiatto:
Gergely Risko <notifications@github.com> writes:

> @thierryvolpiatto Yes, it's fixed, but I don't have a close button on the github ui, maybe only you can close it.
>
> Please close it, i'm pretty sure it's fixed now.

Ok I am going to close now.

Many thanks for your work!

-- 
Thierry

Issue #1820, opened 2017-07-30
helm-read-file-name and default-directory


Ambrevar:
## Expected behavior

`helm-read-file-name` should start in `default-directory`.

## Actual behavior from `emacs-helm.sh` if possible (See note at bottom)

It does not and I'm always puzzled at how it decides which starting directory to pick.
It might be the one of the last visited file.

## Steps to reproduce (recipe)

- Browse a file at location `/foo`.
- Start `eshell` and go to `/bar`. See that `default-directory` is now `/bar`.
- Run `helm-find-files` and go to `/quux`, copy `quux.txt` (default: `M-C`).

The copy will spawn a `helm-read-file-name` started somewhere that might not be `default-directory` (here `/bar`).
I am not sure but it might well be `/foo`, the folder of the last visited file.

## Describe versions of helm, emacs, operating system etc.

- Latest Helm from MELPA.
- Emacs 25.2
- Gentoo

thierryvolpiatto:
Pierre Neidhardt <notifications@github.com> writes:

> Expected behavior
>
> helm-read-file-name should start in default-directory.
>
> Actual behavior from emacs-helm.sh if possible (See note at bottom)
>
> It does not and I'm always puzzled at how it decides which starting directory to pick.
> It might be the one of the last visited file.
>
> Steps to reproduce (recipe)
>
>   * Browse a file at location /foo.
>   * Start eshell and go to /bar. See that default-directory is now /bar.
>   * Run helm-find-files and go to /quux, copy quux.txt (default: M-C).
>
> The copy will spawn a helm-read-file-name started somewhere that might
> not be default-directory (here /bar).

This is due to `helm-dwim-target-directory`.

> I am not sure but it might well be /foo, the folder of the last
> visited file.

I guess you have to windows in 3) step:

- One showing the eshell buffer at /bar
- One showing some file at /foo

When you start helm-find-files from the eshell buffer (at /bar) the
default directory to copy will be /foo (dwim-target-directory).
But if you start with only the eshell buffer it will be /bar.
See `helm-dwim-target-directory` docstring.

-- 
Thierry
Gpg Key fingerprint = 6CEC 7081 AB33 E251 4AB8  5FC2 28D1 7F53 59F2 9997

Ambrevar:
That's exactly that.  Now I understand the mechanics it feels much more
usable! :)

Question: when there is more than 2 windows, which one is the other
window?  `other-window`?

If so, with many windows and when the last window switch was done a
[...]

PR #1821, opened 2017-07-31
Provide completion on dwim-directories (#1820)

thierryvolpiatto:
* helm-files.el (helm-dwim-target): New.
(helm-dwim-target-directory): Use it.

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -629,9 +629,10 @@ Should not be used among other sources.")
 
 (defcustom helm-dwim-target 'completion
   "Default target directory for file actions.
+
 Define the directory where you want to start navigating for the target
 directory when copying, renaming etc... You can use the
-`default-directory' of `next-window' or the current
+`default-directory' of `next-window', the current
 `default-directory' or have completion on all the directories
 belonging to each window."
   :group 'helm-files
@@ -641,10 +642,7 @@ belonging to each window."
           (const :tag "Use initial directory or `default-directory'" nil)))
 
 (defun helm-dwim-target-directory ()
-  "Return value of `default-directory' of buffer in other window.
-If there is only one window return the value of currently visited directory
-if found in `helm-ff-history' or fallback to `default-directory'
-of current buffer."
+  "Try to return a suitable directory according to `helm-dwim-target'."
   (with-helm-current-buffer
     (let* ((wins (remove (get-buffer-window helm-marked-buffer-name)
                          (window-list)))
--- a/helm-files.el
+++ b/helm-files.el
@@ -627,7 +627,7 @@ Should not be used among other sources.")
     (message "Helm find files session bookmarked! ")))
 (put 'helm-ff-bookmark-set 'helm-only t)
 
-(defcustom helm-dwim-target 'completion
+(defcustom helm-dwim-target nil
   "Default target directory for file actions.
 
 Define the directory where you want to start navigating for the target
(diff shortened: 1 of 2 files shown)


helm-files.el
```diff
@@ -627,22 +627,48 @@ Should not be used among other sources.")
...
+  :type '(radio :tag "Define default target directory for file actions."
+          (const :tag "Directory belonging to next window" next-window)
+          (const :tag "Completion on directories belonging to each window" completion)
+          (const :tag "Use initial directory or `default-directory'" nil)))
+
 (defun helm-dwim-target-directory ()
   "Return value of `default-directory' of buffer in other window.
 If there is only one window return the value of currently visited directory
 if found in `helm-ff-history' or fallback to `default-directory'
 of current buffer."
```
Ambrevar: You should mention `helm-dwim-target` here.

Ambrevar:
Looks good to me:
- `'completion` is awesome.
- The `nil` behaviour is possibly the most intuitive. Nice to have that available too.

Some documentation in `C-h m` and this is perfect.

thierryvolpiatto:

> Looks good to me:
>
>   * 'completion is awesome.
>   * The nil behaviour is possibly the most intuitive. Nice to have that available too.

I hesitate making this the default WDYT ?

> Some documentation in C-h m and this is perfect.

Done, would be better though, let me know.
Also I had a doubt for prompt name in completion which is "Target
directory", if you think at a more meaningful name, let me know too.
 
Thanks.

-- 
Thierry

Ambrevar:
> I hesitate making this the default WDYT ?

`'completion' is awesome but spawns an additional Helm buffer, so that's a total of 3 Helm buffers for the copy/move actions.  It can be confusing to beginners.
I think `nil` is more straightforward.  I would make it the default.

> Also I had a doubt for prompt name in completion which is "Target
> directory", if you think at a more meaningful name, let me know too.

Indeed, I don't like the name either, a newcomer will probably wonder what's the difference between the target and the file name.
I don't think we should come up with a new _definition_, we could rather provide an explicit sentence, such as:
- "Browse target starting from:", or "Select target starting from:"
- "Switch to folder before selecting target"
- "Base folder for target selection"

Ambrevar:
Wait, looks like there is some unexpected side effect: with 2 windows open, if I ediff from helm-find-files (`C-c =`) I get prompted for a target directory.  I think this should not happen.

thierryvolpiatto:

>     I hesitate making this the default WDYT ?
>
> 'completion' is awesome but spawns an additional Helm buffer, so
> that's a total of 3 Helm buffers for the copy/move actions. It can be
> confusing to beginners.

Agreed.

> I think `nil` is more straightforward. I would make it the default.

Done.

>     Also I had a doubt for prompt name in completion which is "Target
>     directory", if you think at a more meaningful name, let me know too.
>
> Indeed, I don't like the name either, a newcomer will probably wonder
> what's the difference between the target and the file name.  I don't
> think we should come up with a new definition, we could rather provide
> an explicit sentence, such as:
>
>   * "Browse target starting from:",

I took this one.

Thanks for the feedback, I will merge now.

--
Thierry

thierryvolpiatto:

> Wait, looks like there is some unexpected side effect: with 2 windows
> open, if I ediff from helm-find-files (C-c =) I get prompted for a
> target directory. I think this should not happen.

Indeed, I forget to fix it with new behavior, should work same as before
now.

Thanks.

--
Thierry

PR #1824, opened 2017-08-02
helm-org.el (helm-org-refile-within-buffer): Add function

alphapapa:
This function refiles the entry at point to a heading selected with Helm within BUFFER.  Interactively, BUFFER defaults to the current buffer.

When you want to refile within the current buffer, this is much faster than org-refile, because it only looks at headings in the current buffer.  If you have a lot of headings in a lot of open files, org-refile can take several seconds to bring up the list of headings in all files.

This was partially inspired by John Kitchin's post at <http://kitchingroup.cheme.cmu.edu/blog/2016/03/18/Org-teleport-headlines/>

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -21,6 +21,10 @@
 (require 'helm-utils)
 (require 'org)
 
+;; Load org-with-point-at macro when compiling
+(eval-when-compile
+  (require 'org-macs))
+
 (declare-function org-agenda-switch-to "org-agenda.el")
 
 (defgroup helm-org nil
@@ -58,6 +62,7 @@ Note this have no effect in `helm-org-in-buffer-headings'."
 (defcustom helm-org-headings-actions
   '(("Go to heading" . helm-org-goto-marker)
     ("Open in indirect buffer `C-c i'" . helm-org--open-heading-in-indirect-buffer)
+    ("Refile this heading within this buffer `C-c C-w'" . helm-org-refile-within-buffer-action)
     ("Refile current or marked heading to selection `C-c w`" . helm-org-heading-refile)
     ("Insert link to this heading `C-c l`" . helm-org-insert-link-to-heading-at-marker))
   "Default actions alist for
@@ -111,6 +116,7 @@ Note this have no effect in `helm-org-in-buffer-headings'."
   (let ((map (make-sparse-keymap)))
     (set-keymap-parent map helm-map)
     (define-key map (kbd "C-c i") 'helm-org-run-open-heading-in-indirect-buffer)
+    (define-key map (kbd "C-c C-w") 'helm-org-run-refile-within-buffer-action)
     (define-key map (kbd "C-c w") 'helm-org-run-heading-refile)
     (define-key map (kbd "C-c l") 'helm-org-run-insert-link-to-heading-at-marker)
     map)
@@ -274,6 +280,35 @@ nothing to CANDIDATES."
   (with-helm-alive-p
     (helm-exit-and-execute-action
      'helm-org-insert-link-to-heading-at-marker)))
+
+(defun helm-org-refile-within-buffer (buffer)
+  "Refile current heading to another heading in BUFFER.
+Interactively, BUFFER is the current buffer."
+  (interactive (list (current-buffer)))
+  (let* ((org-reverse-note-order t)
+         ;; Checking buffer-base-buffer ensures that we handle
+         ;; indirect buffers, e.g. created with
+         ;; org-tree-to-indirect-buffer
+         (filename (buffer-file-name (or (buffer-base-buffer (current-buffer))
+                                         (current-buffer))))
+         (target (helm-comp-read "Refile to: " (helm-org--get-candidates-in-file filename)))
+         (rfloc (list nil filename nil target)))
+    (org-refile nil nil rfloc)))
+
+(defun helm-org-refile-within-buffer-action (pos)
+  "Action to refile heading at POS to another heading in current buffer.
+POS is an integer or marker, suitable for use with
+`org-with-point-at'."
+  (org-with-point-at pos
+    (helm-org-refile-within-buffer (current-buffer))))
+
+(defun helm-org-run-refile-within-buffer-action ()
+  "Run `helm-org-refile-within-buffer-action'.
+Call this with a keybinding in the Helm buffer."
+  (interactive)
+  (with-helm-alive-p
+    (helm-exit-and-execute-action #'helm-org-refile-within-buffer-action)))
+(put 'helm-org-run-refile-within-buffer-action 'helm-only t)
 
 (defun helm-org-heading-refile (marker)
   "Refile current heading or marked to MARKER.


thierryvolpiatto:
Is it a new action ?
If so shouldn't it be added to actions ?

alphapapa:
Thanks for reminding me.  I added it to the actions and chose a keybinding similar to the other refile function.

By the way, I noticed that `helm-org-heading-refile` does manual manipulation of the trees, cutting and pasting with manual level selection.  Do you think it would be a good idea to rewrite that to use the `org-refile` function, similar to this function I'm adding?

helm-org.el
```diff
@@ -275,6 +277,36 @@ nothing to CANDIDATES."
...
+                                         (current-buffer))))
+         (target (helm-comp-read "Refile to: " (helm-org--get-candidates-in-file filename)))
+         (rfloc (list nil filename nil target)))
+    (org-refile nil nil rfloc)))
+
+(defun helm-org-refile-within-buffer-action (pos)
+  "Action to refile heading at POS to another heading in current buffer.
+POS is an integer or marker, suitable for use with
+`org-with-point-at'."
+  (interactive)
```
thierryvolpiatto: No need to be interactive for this one.

helm-org.el
```diff
@@ -275,6 +277,36 @@ nothing to CANDIDATES."
     (helm-exit-and-execute-action
      'helm-org-insert-link-to-heading-at-marker)))
 
+(defun helm-org-refile-within-buffer (buffer)
+  "Refile current heading to another heading in BUFFER.
+Interactively, BUFFER is the current buffer."
+  (interactive (list (current-buffer)))
```
thierryvolpiatto: Same why is this one interactive ?

thierryvolpiatto:

> Thanks for reminding me. I added it to the actions and chose a
> keybinding similar to the other refile function.

Thanks.  As soon as you have removed the unused (interactive) I will merge.

> By the way, I noticed that helm-org-heading-refile does manual
> manipulation of the trees, cutting and pasting with manual level
> selection. Do you think it would be a good idea to rewrite that to use
> the org-refile function, similar to this function I'm adding?

If what we have actually is working correctly, no need to rewrite it
IMO, but if you think it's a real improvement, then yes go ahead.

Thanks.

--
Thierry

helm-org.el
alphapapa: Thanks, removed.

helm-org.el
alphapapa: This one is useful as a standalone command, not just as an action.  The user can refile the entry at point to any other heading within the buffer, using Helm to prompt for the target heading.  This was originally a function in my personal config, but I realized it belongs in Helm directly.  :)

Of course it also works as an action by calling it with point at a different heading, which is what we do in `helm-org-refile-within-buffer-action` using the `org-with-point-at` macro.

alphapapa:
> Thanks.  As soon as you have removed the unused (interactive) I will merge.

Done.  One of them should remain, though, because it's useful as a command activated outside of Helm.

> If what we have actually is working correctly, no need to rewrite it IMO, but if you think it's a real improvement, then yes go ahead.

Of course, don't fix what isn't broken.  :)  I like using the more abstracted, built-in Org functions when possible, as that way we benefit from bug fixes upstream and reduce the chance for breakage in the future.  But since it's working, I'll leave it alone.  We can always rewrite it to use `org-refile` if we need to.

Thanks.

thierryvolpiatto:
Thanks, merged.

thierryvolpiatto:
Hi,
I just tried this and I have to say I am confused, i couldn't make it working even with a simple org buffer.
Could you explain how to use this and how it is intended to work.
Also, what is exactly the difference with the actual refile action (the one we add since a while).

Thanks.

alphapapa:
> Hi, I just tried this and I have to say I am confused, i couldn't make it working even with a simple org buffer. Could you explain how to use this and how it is intended to work.

Ok, I just pulled from master so this definitely works:

1.  Open an Org file with multiple headings.
2.  Put point inside an entry.
3.  Run `helm-org-refile-within-buffer`.
4.  Choose a heading and hit `RET`.
5.  The heading you started with will be refiled to the one you selected.

You're right that this can be done with the existing refile action.  I did not realize this because I have been confused by the description, "Refile current or marked heading to selection".  I couldn't figure out if "current" meant "where point was when you ran Helm" or "the heading you selected in Helm", whether "marked" meant "marked in Helm" or "marked in the source buffer when you called Helm," and whether "selection" meant "selected in Helm" or "in the selected region in the source buffer when you called Helm."  So I ended up making a function to do what I needed, which ended up having some overlap with this existing action.

However, this also works, which the existing action cannot do:

1.  Open Org file.
2.  Run `helm-org-in-buffer-headings`.
3.  Choose a heading and open the actions menu.
4.  Choose `Refile this heading within this buffer \`C-c C-w\``.
5.  Choose a heading from the new list that appears and press `RET`.
6.  The heading you selected in `helm-org-in-buffer-headings` will be refiled to the heading you selected after selecting the refile action.

Also, the new action function is useful when called as a command outside of Helm, which the existing action function can't do.

I understand the confusion; it's partly why I made this PR in the first place, because I was confused.  :)  If you would like, I will send a new PR which consolidates these actions and clarifies their descriptions.

Thanks.

thierryvolpiatto:

I am sorry to haven't understood that your action does the same than the
original one.

> However, this also works, which the existing action cannot do:

Don't think so, I think you can, see below.

>  1. Open Org file.
>  2. Run helm-org-in-buffer-headings.
>  3. Choose a heading and open the actions menu.
>  4. Choose Refile this heading within this buffer \C-c C-w``.
>  5. Choose a heading from the new list that appears and press RET.
>  6. The heading you selected in helm-org-in-buffer-headings will be refiled to the heading you selected after selecting the refile action.

You can do this with `helm-org-heading-refile` isn't it ?

1) Open org file
2) Run `helm-org-in-buffer-headings`
3) Choose a heading and mark it
4) Move selection where you want to refile
5) Run `helm-org-heading-refile` action from action menu or key

The heading you MARKED will be refiled to the SELECTED candidate.

I agree this is confusing and probably bad described, but it was the
only way I found at the time to avoid running again a new helm-buffer
displaying the same information than the original one.

This is described in the help buffer accessed from C-h m, probably it
could be improved ?

BTW a user recently was confused and asked about refiling:

https://groups.google.com/forum/?hl=en#!topic/emacs-helm/RD7e0QlTJbk

> Also, the new action function is useful when called as a command
> outside of Helm, which the existing action function can't do.

Agree for this one, but perhaps we should remove the new action if we agree
both actions (original and yours) do the same thing, this to avoid more
confusion.

Thanks.

--
Thierry

alphapapa:
Thanks, I submitted a new PR that removes these redundant functions I added and rewrites the old one to handle multiple selection, use `org-refile` for safety and robustness, and updates the help and docstrings.

> BTW a user recently was confused and asked about refiling:

> https://groups.google.com/forum/?hl=en#!topic/emacs-helm/RD7e0QlTJbk

Yeah, he asked on the org-mode list, thinking that helm-org-* was part of Org, so I asked him to open an issue about it.  :)

PR #1828, opened 2017-08-05
Update eshell commentary

bb2020:
`helm-eshell-history` acts more like `eshell-previous-matching-input` (M-r) instead of `eshell-previous-matching-input-from-input` (M-p).


thierryvolpiatto:
Thanks, merged.

PR #1829, opened 2017-08-05
helm-org.el: Remove new refile functions, replace old one

alphapapa:
As discussed at https://github.com/emacs-helm/helm/pull/1824.  This removes the redundant new functions and rewrites the old one to be simpler, more robust, and allow refiling of multiple headings at once.  The description in the help file is also improved.

(helm-org-refile-within-buffer): Remove function
(helm-org-refile-within-buffer-action): Remove function
(helm-org-run-refile-within-buffer-action): Remove function

These were recently added but are redundant, so now removed.

(helm-org-heading-refile): Remove function

This function is replaced by the following new one.

(helm-org--refile-heading-to): Add function

This function replaces the old one, and adds support for refiling
multiple headings at once.  It also uses the org-refile function to do
the work, making the code simpler and more robust (org-refile does
some sanity checking before refiling headings).

(helm-org-run-heading-refile): Rename function

Renamed to helm-org-run-refile-heading-to to match new function.

Thanks.

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -62,8 +62,7 @@ Note this have no effect in `helm-org-in-buffer-headings'."
 (defcustom helm-org-headings-actions
   '(("Go to heading" . helm-org-goto-marker)
     ("Open in indirect buffer `C-c i'" . helm-org--open-heading-in-indirect-buffer)
-    ("Refile this heading within this buffer `C-c C-w'" . helm-org-refile-within-buffer-action)
-    ("Refile current or marked heading to selection `C-c w`" . helm-org-heading-refile)
+    ("Refile heading(s) (multiple-marked-to-selected, or current-to-selected) `C-c w`" . helm-org--refile-heading-to)
     ("Insert link to this heading `C-c l`" . helm-org-insert-link-to-heading-at-marker))
   "Default actions alist for
   `helm-source-org-headings-for-files'."
@@ -116,8 +115,7 @@ Note this have no effect in `helm-org-in-buffer-headings'."
   (let ((map (make-sparse-keymap)))
     (set-keymap-parent map helm-map)
     (define-key map (kbd "C-c i") 'helm-org-run-open-heading-in-indirect-buffer)
-    (define-key map (kbd "C-c C-w") 'helm-org-run-refile-within-buffer-action)
-    (define-key map (kbd "C-c w") 'helm-org-run-heading-refile)
+    (define-key map (kbd "C-c w") 'helm-org-run-refile-heading-to)
     (define-key map (kbd "C-c l") 'helm-org-run-insert-link-to-heading-at-marker)
     map)
   "Keymap for `helm-source-org-headings-for-files'.")
@@ -281,57 +279,25 @@ nothing to CANDIDATES."
     (helm-exit-and-execute-action
      'helm-org-insert-link-to-heading-at-marker)))
 
-(defun helm-org-refile-within-buffer (buffer)
-  "Refile current heading to another heading in BUFFER.
-Interactively, BUFFER is the current buffer."
-  (interactive (list (current-buffer)))
-  (let* ((org-reverse-note-order t)
-         ;; Checking buffer-base-buffer ensures that we handle
-         ;; indirect buffers, e.g. created with
-         ;; org-tree-to-indirect-buffer
-         (filename (buffer-file-name (or (buffer-base-buffer buffer) buffer)))
-         (target (helm-comp-read "Refile to: "
-                                 (helm-org--get-candidates-in-file filename)))
-         (rfloc (list nil filename nil target)))
-    (org-refile nil nil rfloc)))
-
-(defun helm-org-refile-within-buffer-action (pos)
-  "Action to refile heading at POS to another heading in current buffer.
-POS is an integer or marker, suitable for use with
-`org-with-point-at'."
-  (org-with-point-at pos
-    (helm-org-refile-within-buffer (current-buffer))))
-
-(defun helm-org-run-refile-within-buffer-action ()
-  "Run `helm-org-refile-within-buffer-action'.
-Call this with a keybinding in the Helm buffer."
-  (interactive)
-  (with-helm-alive-p
-    (helm-exit-and-execute-action #'helm-org-refile-within-buffer-action)))
-(put 'helm-org-run-refile-within-buffer-action 'helm-only t)
-
-(defun helm-org-heading-refile (marker)
-  "Refile current heading or marked to MARKER.
-The current heading is the heading where cursor was
-before entering helm session, it will be used unless
-you mark a candidate, in this case helm will go to this marked
-candidate in org buffer and refile this candidate to MARKER.
-NOTE that of course if you have marked more than one candidate,
-all the subsequent candidates will be ignored."
-  (let ((mkd (with-helm-buffer
-               (and helm-marked-candidates
-                    (car (helm-marked-candidates))))))
-    (save-selected-window
-      (when (eq major-mode 'org-agenda-mode)
-        (org-agenda-switch-to))
-      (when mkd (helm-org-goto-marker mkd))
-      (org-cut-subtree)
-      (let ((target-level (with-current-buffer (marker-buffer marker)
-                            (goto-char (marker-position marker))
-                            (org-current-level))))
-        (helm-org-goto-marker marker)
-        (org-end-of-subtree t t)
-        (org-paste-subtree (+ target-level 1))))))
+(defun helm-org--refile-heading-to (marker)
+  "Refile headings to heading at MARKER.
+If multiple candidates are marked in the Helm session, they will
+all be refiled.  If no headings are marked, the selected heading
+will be refiled."
+  (let* ((victims (with-helm-buffer (helm-marked-candidates)))
+         (buffer (marker-buffer marker))
+         (filename (buffer-file-name buffer))
+         (rfloc (list nil filename nil marker)))
+    (when (and (= 1 (length victims))
+               (equal (helm-get-selection) (car victims)))
+      ;; No candidates are marked; we are refiling the entry at point
+      ;; to the selected heading
+      (setq victims (list (point))))
+    ;; Probably best to check that everything returned a value
+    (when (and victims buffer filename rfloc)
+      (cl-loop for victim in victims
+               do (org-with-point-at victim
+                    (org-refile nil nil rfloc))))))
 
 (defun helm-org-in-buffer-preselect ()
   (if (org-on-heading-p)
@@ -340,11 +306,11 @@ all the subsequent candidates will be ignored."
         (outline-previous-visible-heading 1)
         (buffer-substring-no-properties (point-at-bol) (point-at-eol)))))
 
-(defun helm-org-run-heading-refile ()
+(defun helm-org-run-refile-heading-to ()
   (interactive)
   (with-helm-alive-p
-    (helm-exit-and-execute-action 'helm-org-heading-refile)))
-(put 'helm-org-run-heading-refile 'helm-only t)
+    (helm-exit-and-execute-action 'helm-org--refile-heading-to)))
+(put 'helm-org-run-refile-heading-to 'helm-only t)
 
 ;;;###autoload
 (defun helm-org-agenda-files-headings ()
--- a/helm-help.el
+++ b/helm-help.el
@@ -1381,18 +1381,22 @@ but this will not push a new entry with concatenated candidates in kill-ring.
 
 *** Refiling
 
-The heading to refile will be the one you were at when starting helm
-session, and the place to refile this heading will be the selected
-candidate (i.e the candidate at point in helm buffer). If you want to
-refile another one, move to it in helm buffer, mark it, then move to
-the candidate of your choice to refile at this place.
-NOTE that of course if you have marked more than one candidate,
-all the subsequent candidates will be ignored.
+You can refile one heading or multiple headings at a time.
+
+To refile one heading, move the point to the entry you want to
+refile, and run \\[helm-org-in-buffer-headings].  Then select the
+heading you want to refile to, and press \\[C-c w], or select the
+refile action from the actions menu.
+
+To refile multiple headings, run \\[helm-org-in-buffer-headings]
+and mark the headings you want to refile.  Then select the
+heading you want to refile to (without marking it), and press
+\\[C-c w], or select the refile action from the actions menu.
 
 ** Commands
 \\<helm-org-headings-map>
 \\[helm-org-run-open-heading-in-indirect-buffer]\t\tOpen heading in indirect buffer.
-\\[helm-org-run-heading-refile]\t\tRefile current heading to selection.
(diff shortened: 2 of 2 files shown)


helm-org.el
```diff
@@ -281,57 +279,20 @@ nothing to CANDIDATES."
...
-        (helm-org-goto-marker marker)
-        (org-end-of-subtree t t)
-        (org-paste-subtree (+ target-level 1))))))
+(defun helm-org--refile-heading-to (marker)
+  "Refile headings to heading at MARKER.
+If multiple candidates are marked in the Helm session, they will
+all be refiled.  If no headings are marked, the selected heading
+will be refiled."
+  (when-let ((victims (with-helm-buffer
+                        (or (when helm-marked-candidates
```
thierryvolpiatto: You can use directly `helm-marked-candidates` (the function) instead, anyway if `helm-get-selelection` is used it will trigger an error because it is not a list.

helm-org.el
```diff
@@ -62,8 +62,7 @@ Note this have no effect in `helm-org-in-buffer-headings'."
 (defcustom helm-org-headings-actions
   '(("Go to heading" . helm-org-goto-marker)
     ("Open in indirect buffer `C-c i'" . helm-org--open-heading-in-indirect-buffer)
-    ("Refile this heading within this buffer `C-c C-w'" . helm-org-refile-within-buffer-action)
-    ("Refile current or marked heading to selection `C-c w`" . helm-org-heading-refile)
+    ("Refile heading(s) (multiple-marked-to-selected, or current-to-selected) `C-c w`" . helm-org--refile-heading-to)
```
thierryvolpiatto: Isn't "Refile heading(s) to selected `C-c w'" enough ?

thierryvolpiatto:
I have added some comments, otherwise looks good, thanks.

thierryvolpiatto:
Also, just a question, isn't it possible that refiling several headings (marked) at once cause problems depending on which headers user have marked ?

helm-org.el
alphapapa: Maybe, but I think having the extra explanation may help the discoverability of marking and refiling multiple headings at once.  All I know for sure is that, after several years, it wasn't until you accepted my previous PR and you reviewed it that I understood what `Refile current or marked heading to selection` meant.  ;)  (That's not a criticism; it's very hard to get the idea across clearly in a few words.)

helm-org.el
alphapapa: Fixed, thanks!

alphapapa:
> Also, just a question, isn't it possible that refiling several headings (marked) at once cause problems depending on which headers user have marked ?

I'm not sure what you mean.  Could you give an example?  I think it works fine because:

1.  The positions of the headings are markers, so each one should stay with its heading as others are refiled.
2.  `org-refile` does sanity and safety checking before refiling.
3.  I tested it and it seems to work fine.  :)

But if you can think of anything I missed, please let me know.

alphapapa:
> Also, just a question, isn't it possible that refiling several headings (marked) at once cause problems depending on which headers user have marked ?

Clarification though: by "marked" do you mean "marked in the Helm buffer" or "marked in the source buffer before activating Helm"?

thierryvolpiatto:

>     Also, just a question, isn't it possible that refiling several
>     headings (marked) at once cause problems depending on which
>     headers user have marked ?
>
> Clarification though: by "marked" do you mean "marked in the Helm
> buffer" or "marked in the source buffer before activating Helm"?

Marked in the helm buffer.


--
Thierry

helm-org.el
```diff
@@ -281,57 +279,19 @@ nothing to CANDIDATES."
...
-        (org-end-of-subtree t t)
-        (org-paste-subtree (+ target-level 1))))))
+(defun helm-org--refile-heading-to (marker)
+  "Refile headings to heading at MARKER.
+If multiple candidates are marked in the Helm session, they will
+all be refiled.  If no headings are marked, the selected heading
+will be refiled."
+  (when-let ((victims (with-helm-buffer
+                        (or (helm-marked-candidates)
+                            (list (helm-get-selection)))))
```
thierryvolpiatto: This is not needed, helm-marked-candidates will always return a list with only current candidate so instead of 

```elisp
(or (helm-marked-candidates)
+                            (list (helm-get-selection)))
```
You need only
```elisp
(helm-marked-candidates
```

helm-org.el
```diff
@@ -281,57 +279,19 @@ nothing to CANDIDATES."
...
-                            (org-current-level))))
-        (helm-org-goto-marker marker)
-        (org-end-of-subtree t t)
-        (org-paste-subtree (+ target-level 1))))))
+(defun helm-org--refile-heading-to (marker)
+  "Refile headings to heading at MARKER.
+If multiple candidates are marked in the Helm session, they will
+all be refiled.  If no headings are marked, the selected heading
+will be refiled."
+  (when-let ((victims (with-helm-buffer
```
thierryvolpiatto: Also please don't use `when-let` which is not available in emacs-24.5, use `helm-aif` instead.

thierryvolpiatto:

>     Also, just a question, isn't it possible that refiling several headings (marked) at once cause problems depending on which headers user have marked ?
>    
> I'm not sure what you mean. Could you give an example? I think it works fine because:
>
>  1. The positions of the headings are markers, so each one should stay with its heading as others are refiled.
>  2. org-refile does sanity and safety checking before refiling.
>  3. I tested it and it seems to work fine. :)

Ok thanks for clarification.

-- 
Thierry

helm-org.el
alphapapa: Thanks.

helm-org.el
alphapapa: Understood.

alphapapa:
This latest comment addresses your latest feedback.  Thanks.

helm-org.el
```diff
@@ -281,57 +279,20 @@ nothing to CANDIDATES."
...
-        (helm-org-goto-marker marker)
-        (org-end-of-subtree t t)
-        (org-paste-subtree (+ target-level 1))))))
+(defun helm-org--refile-heading-to (marker)
+  "Refile headings to heading at MARKER.
+If multiple candidates are marked in the Helm session, they will
+all be refiled.  If no headings are marked, the selected heading
+will be refiled."
+  (let* ((victims (with-helm-buffer (helm-marked-candidates)))
+         (buffer (marker-buffer marker))
```
thierryvolpiatto: Something seems wrong here, if you are now operating on marked candidates, why are you using MARKER, which is a marker refering to the current candidate, and BTW on its buffer, so I guess you have to compute marker-buffer on each candidates while looping, MARKER will be unused (use _marker).

thierryvolpiatto:
Added some comments, here how I would write the function, assuming the usage of org-refile is good (I know nothing about org-refile)

```elisp
(defun helm-org--refile-heading-to (marker)
  "Refile marked headings to selected heading.
If multiple candidates are marked in the Helm session, they will
all be refiled.  If no headings are marked, the selected heading
will be refiled."
  (cl-loop with sel = marker
           for marker in (helm-marked-candidates)
           for filename = (buffer-file-name (marker-buffer marker))
           ;; Selection is on one of the marked candidate
           unless (eql sel marker)
           do (org-with-point-at marker
                (org-refile nil nil (list nil filename nil marker)))))))
```
Fully not tested, I have also added an extra checking in case user have helm selection on one of its marked candidates.

thierryvolpiatto:
This one looks more correct, still not fully tested

```elisp
(defun helm-org-heading-refile (_marker)
  "Refile marked headings to selected heading.
If multiple candidates are marked in the Helm session, they will
all be refiled.  If no headings are marked, the selected heading
will be refiled."
  ;; Move (helm-marked-candidates) to (helm-get-selection)
  (cl-loop with target = (helm-get-selection)
           for m in (helm-marked-candidates)
           for filename = (buffer-file-name (marker-buffer m))
           ;; Selection is on one of the marked candidate
           unless (eql target m)
           do (org-with-point-at m
                  (org-refile nil nil (list nil filename nil target)))))
```

thierryvolpiatto:
I used the old name to quickly test it.  Sorry no more time now for this until tonight.

helm-org.el
alphapapa: > Something seems wrong here, if you are now operating on marked candidates, why are you using MARKER, which is a marker refering to the current candidate,

`marker` is the marker of the target heading, i.e. the heading that candidates will be refiled *to*.  It is the `REAL` of the `(DISPLAY . REAL)` pair.  It is only computed once because there is only one target.  It is only used to get the underlying filename of the target heading's buffer.

> I guess you have to compute marker-buffer on each candidates while looping, MARKER will be unused (use _marker).

This is not necessary.  `org-with-point-at` takes a marker as its first argument.  Each `victim` is a marker.  `org-refile` operates with point at an entry, which is what `org-with-point-at` provides.

alphapapa:
I think this confusion we're having points to an underlying problem: this function is trying to operate in two different modes:

1.  Without marked candidates, the "victim" heading is the entry where the point was when `helm-org-in-buffer-headings` was called, and the target heading is the Helm "selection" when `helm-org--refile-heading-to` is called.
2.  With marked candidates, the "victim" headings are the marked candidates, and the target heading is the Helm "selection" when `helm-org--refile-heading-to` is called.

To me, the second mode seems very natural to use: mark the headings you want to refile, select the target, execute the refile action.  It's like saying, "Put these here."

But the first mode is very confusing to me, probably because the "victim" heading is implicit.  It's like saying, "Remember that heading I was looking at when I called this Helm command that isn't dedicated to refiling?  Move it to here."  Even now I am confused by it so that, apparently, this latest commit I made doesn't work in this mode.  

In the other PR, and in my personal config before I submitted it, I essentially obsoleted the first mode with a separate command that is called directly, rather from within Helm.  The user runs the "refile this heading" command, then selects the target heading with Helm, and the entry at point is refiled.  This seems natural to me.

I think the confusion is made worse by the implicit nature of the first mode.  That's why, in my first PR, I added a new action that worked in two stages: first the user ran `helm-org-in-buffer-headings`, then selected the "victim" heading, then selected the "refile within this buffer" action, then the user was presented with a new list of target headings, then the user selected one and hit `RET`, and then the heading was refiled.  There was no implicit victim heading, it was all explicit.

And while it's nice that `helm-marked-candidates` returns the selection if nothing is marked, this makes supporting the dual modes more confusing.  If the user is using the first mode, `helm-marked-candidates` will return one candidate; but if the user is using the second mode and only marks one candidate, it will also return one candidate.  I think this is why the first mode is broken in my latest commit.

So I can fix that, but I wonder if we should abandon the dual-mode refiling action and go with:

1.  The second mode, in which the user explicitly marks victim headings and then selects the target heading.
2.  A separate command called outside of a Helm session to refile the current entry, which prompts the user to select only the target heading.

Another possibility would be to describe the action in the actions list as "Refile TO this heading", emphasizing that the heading the action is called on is the *target* of the refiling action.  But that would still make the victim heading implicit, which I think is confusing.  And if it's confusing to me, having written this code, I'm guessing it will be confusing to users as well.

Let me know what you think.

alphapapa:
I just pushed a commit that fixed the "first mode", in which there are no marked candidates and we refile the entry at point to the Helm selection  As far as I know, everything now works correctly, including both "modes."  So it should be safe to merge.

If you want, I will submit another PR that moves the "first mode" into a separate command, so that the "victim" heading is always explicit.  I think this would be less confusing and also much more discoverable, as users could find a `helm-org-refile-this-heading` command and bind it to a key.  Multiple headings can be refiled with the `helm-org-in-buffer-headings` command.  Although it would help discoverability to make an explicit command for that as well, one that requires "victim" headings to be explicitly marked.

So my suggestion is this:

1.  Make the refile action in `helm-org-in-buffer-headings` only work in the marked-candidates mode.  If there are no marked candidates, it should give an error.
2.  Make a separate command to refile the heading at point, e.g. `helm-org-refile-this-heading` (the "this" meaning that it explicitly refiles the heading the user is at, not any other heading).
3.  Make a third command to explicitly refile multiple headings.  This would essentially call `helm-org-in-buffer-headings` with only a single action.  This command could also be enhanced to pre-mark headings that are in the region, when the region is active.  So the user could 1) mark headings in a region, e.g. with `M-h` in an Org buffer, 2) call `helm-org-refile-these-headings`, 3) the Helm session would open with all headings in the region already marked, 4) the user would select the target heading and press `RET`, 5) the headings would be refiled.

I will send a PR for that if you like.  Let me know what you want to do.  Thanks.

thierryvolpiatto:
Ok, I have finally merged, if there is problems we will fix them along the way, I am too buzzy actually to review carefully all this.
Thanks for your work.

thierryvolpiatto:

> I just pushed a commit that fixed the "first mode", in which there are no marked candidates and we refile the entry at point to the Helm selection As far as I know,
> everything now works correctly, including both "modes." So it should be safe to merge.
>
> If you want, I will submit another PR that moves the "first mode" into a separate command, so that the "victim" heading is always explicit. I think this would be less
> confusing and also much more discoverable, as users could find a helm-org-refile-this-heading command and bind it to a key. Multiple headings can be refiled with the
> helm-org-in-buffer-headings command. Although it would help discoverability to make an explicit command for that as well, one that requires "victim" headings to be
> explicitly marked.
>
> So my suggestion is this:
>
>  1. Make the refile action in helm-org-in-buffer-headings only work in the marked-candidates mode. If there are no marked candidates, it should give an error.
>  2. Make a separate command to refile the heading at point, e.g. helm-org-refile-this-heading (the "this" meaning that it explicitly refiles the heading the user is
>     at, not any other heading).
>  3. Make a third command to explicitly refile multiple headings. This would essentially call helm-org-in-buffer-headings with only a single action. This command could
>     also be enhanced to pre-mark headings that are in the region, when the region is active. So the user could 1) mark headings in a region, e.g. with M-h in an Org
>     buffer, 2) call helm-org-refile-these-headings, 3) the Helm session would open with all headings in the region already marked, 4) the user would select the target
>     heading and press RET, 5) the headings would be refiled.
>
> I will send a PR for that if you like. Let me know what you want to do. Thanks.

Looks great, isn't 2) and 3) be made in a single command ?
i.e if region is active refile headings in region, otherwise the heading
at point. Also Julien said it is possible to mark things in org agenda,
I am not aware of this feature, but I guess it could be implemented.

-- 
Thierry

alphapapa:
> Looks great, isn't 2) and 3) be made in a single command ? i.e if region is active refile headings in region, otherwise the heading at point.

Yes, of course, you're right.  :)  I'll work on this.

> Also Julien said it is possible to mark things in org agenda, I am not aware of this feature, but I guess it could be implemented.

It could, but I think it's not necessary, because if `helm-mode` is active, and Helm is the default completion method, it will already work with Helm with the existing bulk `org-agenda-refile` command.  I guess the benefit of having a Helm-specific command would be that we could fontify the headings, but I don't think it would be worth reimplementing the command just for that.

However, an idea that might be interesting would be a `helm-org-agenda-entries` command, or something like that, that would present all the entries in the agenda, allowing the user to filter, navigate, mark, refile, etc.  I think that would be fairly easy to do, and would be generally useful.  If you like the idea, I'll see what I can do.

PR #1833, opened 2017-08-07
helm-org.el (helm-org-in-buffer-headings): Don't nullify helm-org-format-outline-path

alphapapa:
Since this command shows headings in the current buffer, it makes sense to hide the filename, but showing the full outline path can be very useful in narrowing down the list.  For example, in a tree like:

* Languages
** English
*** Alphabet
** Greek
*** Alphabet

If the user wanted to go to the "Greek/Alphabet" heading, without displaying the outline path, he'll see this after typing "alphabet":

*** Alphabet
*** Alphabet

Which wouldn't be useful.  But with full paths, he'd see:

** Languages/English/Alphabet
** Languages/Greek/Alphabet

By leaving helm-org-format-outline-path alone, we respect the user's setting, and allow it to be overridden by binding it around this command.

Thanks.

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -325,7 +325,7 @@ will be refiled."
 (defun helm-org-in-buffer-headings ()
   "Preconfigured helm for org buffer headings."
   (interactive)
-  (let (helm-org-show-filename helm-org-format-outline-path)
+  (let (helm-org-show-filename)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks.

PR #1839, opened 2017-08-15
declare init-hooks for the grep and moccur sources

sameerds:
These hooks are useful for external packages. For example, the new after-init-hooks introduced here will be used by the package `[helm-modeless](https://github.com/sameerds/helm-modeless)' that I am working on.


thierryvolpiatto:
Thanks!

PR #1855, opened 2017-09-12
Overhaul emacs-helm.sh

Ambrevar:
Please review:
- There might be a good reason for the `2` exit code instead of `1`.
- Why was `helm-M-x` set differently?

### DIFF
--- a/emacs-helm.sh
+++ b/emacs-helm.sh
@@ -110,5 +110,4 @@ cat > $CONF_FILE <<EOF
 (add-hook 'kill-emacs-hook #'(lambda () (and (file-exists-p "$CONF_FILE") (delete-file "$CONF_FILE"))))
 EOF
 
-$EMACS -Q -l $CONF_FILE "$@"
-
+$EMACS -Q -l "$CONF_FILE" "$@"
--- a/emacs-helm.sh
+++ b/emacs-helm.sh
@@ -2,24 +2,23 @@
 
 
 ## Copyright (C) 2012 ~ 2017 Thierry Volpiatto <thierry.volpiatto@gmail.com>
-## 
+##
 ## This program is free software; you can redistribute it and/or modify
 ## it under the terms of the GNU General Public License as published by
 ## the Free Software Foundation, either version 3 of the License, or
 ## (at your option) any later version.
-## 
+##
 ## This program is distributed in the hope that it will be useful,
 ## but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:

> Please review:
>
>   * There might be a good reason for the 2 exit code instead of 1.

No particualr reason AFAIR

>   * Why was helm-M-x set differently?

No problem to remap instead of global binding M-x.

Thanks.


-- 
Thierry

thierryvolpiatto:

> Please review:
>
>   * There might be a good reason for the 2 exit code instead of 1.

No particualr reason AFAIR

>   * Why was helm-M-x set differently?

No problem to remap instead of global binding M-x.

Thanks.


-- 
Thierry

thierryvolpiatto:

> Please review:
>
>   * There might be a good reason for the 2 exit code instead of 1.

I finally had time to read the bash documentation of exit codes:

   All of the Bash builtins return an exit status of zero if they
succeed and a non-zero status on failure, so they may be used by the
conditional and list constructs.  All builtins return an exit status of
2 to indicate incorrect usage.

This is why I used 2, which is correct IMO.

-- 
Thierry

Ambrevar:
1. Your script is a script, not a Bash built-in.
2. It's not Bash anymore :)

thierryvolpiatto:

>  1. Your script is a script, not a Bash built-in.
>  2. It's not Bash anymore :)

Perhaps, but I guess the conventions (posix?) for exit are not much different
between sh and bash, isn't it?

-- 
Thierry

Ambrevar:
Regardless. Built-ins (`test`, etc.) return some error code:  There is no reason `emacs-helm.sh` should return the same error code as built-ins.  We are talking orange and bananas here.  `emacs-helm.sh` is a program, and like any Unix program it should return 0 on success, or _its own convention of exit codes_ on error.  If the exit codes don't matter, leave them to 1.  There is no good reason to puzzle the contributors by skipping 1.

Ambrevar:
For future reference: A quick look at the dash(1) man page seems to show no mention of the `2` exit code.
A quick practical test (`test -ne 1; echo $?`) return 2, so it could be POSIX.

thierryvolpiatto:

> Regardless. Built-ins (test, etc.) return some error code: There is no
> reason emacs-helm.sh should return the same error code as
> built-ins. We are talking orange and bananas here. emacs-helm.sh is a
> program, and like any Unix program it should return 0 on success, or
> its own convention of exit codes on error. If the exit codes don't
> matter, leave them to 1. There is no good reason to puzzle the
> contributors by skipping 1.

Thanks for explanations.


--
Thierry

PR #1862, opened 2017-09-13
Fix helm-ucs for upcoming Emacs release

notetiene:
A new incompatible change is that ucs-names will be a hash-table
instead of a list.  See Emacs Bug#28302.

* helm-font.el (helm-calculate-ucs-alist-max-len): Add function to
calculate ucs-names candicates when wrapped inside a list.
(helm-calculate-ucs-hash-table-max-len): Add function to calculate
ucs-names candicates when wrapped inside a hash-table.
(helm-calculate-ucs-max-len): Refactor to support the hash-table.
(helm-ucs-collect-symbols-alist): Add function to collect the
ucs-names list.
(helm-ucs-collect-symbols-hash-table): Add function to collect the
ucs-names hash-table.
(helm-ucs-collect-symbols): Refactor to support the hash-table.

### DIFF
--- a/helm-font.el
+++ b/helm-font.el
@@ -76,12 +76,77 @@
 (defvar helm-ucs--names nil)
 (defvar helm-ucs-history nil)
 
-(defun helm-calculate-ucs-max-len ()
-  "Calculate the length of longest `ucs-names' candidate."
-  (cl-loop for (_n . v) in (ucs-names)
+(defun helm-calculate-ucs-alist-max-len (names)
+  "Calculate the length of the longest NAMES list candidate."
+  (cl-loop for (_n . v) in names
            maximize (length (format "#x%x:" v)) into code
            maximize (max 1 (string-width (format "%c" v))) into char
            finally return (cons code char)))
+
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks, merged.

thierryvolpiatto:
Just tried with emacs-26, will need some more filtering with new behavior.

notetiene:
Do you mean with `helm-ucs` or in general?

I know for sure I could refactor the code merged to a simpler loop probably. However, I'm not really used to cl-loops.

thierryvolpiatto:
Etienne <notifications@github.com> writes:

> Do you mean with helm-ucs or in general?

It is general, but it is not due to your recent changes, it was here but
I didn't see it, I found (with emacs-26 it popup in first position) that
"LINE FEED (LF)" is not printable (it represent "\n"), I removed it for
now as I don't think people needs helm-ucs to print "\n".

> I know for sure I could refactor the code merged to a simpler loop
> probably. However, I'm not really used to cl-loops.

Probably, but don't worrie of this I will do it if needed.

Thanks again for your work.

--
Thierry

PR #1863, opened 2017-09-14
Improve phrasing in most of helm-help.el

Ambrevar:
This is a first run.  It probably needs more editing:
- Some parts are redundant or not very clear.
- Some bindings might not be correct.
- There could be more typos...

I can't understand the documentation of helm-kill-ring properly.  Could you rewrite it for me?
You can e-mail me if you like.  I understand your mother tongue is French: you can send me a documentation in French, I'll translate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 </p>
 
 
-<h1 align="center">Emacs-helm</h1>
+<h1 align="center">Emacs-Helm</h1>
 
 <p align="center">
   <img src="https://avatars3.githubusercontent.com/u/1541688?v=3&amp;s=200" alt="Emacs-helm" title="" />
--- a/helm-help.el
+++ b/helm-help.el
@@ -1071,17 +1071,17 @@ use a prefix argument.
 \\[helm-grep-run-default-action]\t\tRun default action (same as `RET').
 \\[helm-grep-run-save-buffer]\t\tSave to a `helm-grep-mode' enabled buffer.")
 
-;;; Pdf grep help
+;;; PDF grep help
 ;;
 ;;
 (defvar helm-pdfgrep-help-message
-  "* Helm PdfGrep Map
+  "* Helm PDFgrep Map
 
 ** Commands
 \\<helm-pdfgrep-map>
-\\[helm-goto-next-file]\t\tNext File.
-\\[helm-goto-precedent-file]\t\tPrecedent File.
-\\[helm-yank-text-at-point]\t\tYank Text at point in minibuffer.")
+\\[helm-goto-next-file]\t\tNext file.
+\\[helm-goto-precedent-file]\t\tPrevious file.
+\\[helm-yank-text-at-point]\t\tYank text at point in minibuffer.")
 
 ;;; Etags help
 ;;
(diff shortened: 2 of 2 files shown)


thierryvolpiatto:

> This is a first run. It probably needs more editing:

Many thanks for this.

>   * Some parts are redundant or not very clear.

Let me know the parts that are not clear, if you feel some parts are
redundants feel free to remove them where needed but keeping in mind
that each help string is local to a source, so some importants things
may be repeated in differents sources.

>   * Some bindings might not be correct.

As some default bindings have changed over the years, for sure I forget
to update some of them.

>   * There could be more typos...

Oh yes!

> I can't understand the documentation of helm-kill-ring properly.

Can you tell me which parts you don't understand so that I make the
rewrite clearer ?

Thanks.

-- 
Thierry

Ambrevar:
> You can bring any candidate on top of kill-ring by using `\\<helm-map>\\[helm-kill-selection-and-quit]'
> on it, this allow for example to paste this candidate using x-selection in an external application like
> firefox.

You that any candidate can be copied to the clipboard/primary-selection, right?

> Contrarily to the regular helm command `helm-kill-selection-and-quit'
> which is bound in helm to same key,

That's the same command!?  What do you mean "bound to the same key"?

> it is here killing the real value (i.e the whole candidate not truncated)
> without the need of a prefix argument.

We must explain what "truncating" means.

> Note: Of course numeric prefix args like `yank' are not available due to the nature
> of helm which allow you to see and move to the desired candidate

 "yank" is not a numeric prefix arg... What do you mean?

I think you mean something like this:

"As opposed to `yank', numeric prefix arguments are ignored with `helm-show-kill-ring' as there is no need for them since selection happens within Helm."

> and also because
> it is already available in helm with [[Shortcuts for executing Default Action on the nth candidate][Shortcuts for executing Default Action on the nth candidate]].

What is available in helm?

thierryvolpiatto:

>     You can bring any candidate on top of kill-ring by using `\\[helm-kill-selection-and-quit]'
>     on it, this allow for example to paste this candidate using x-selection in an external application like
>     firefox.
>    
> You that any candidate can be copied to the clipboard/primary-selection, right?

Yes

>     Contrarily to the regular helm command `helm-kill-selection-and-quit'
>     which is bound in helm to same key,
>    
> That's the same command!? What do you mean "bound to the same key"?

No in the helm-kill-ring-map C-c C-k is bound to
helm-kill-ring-kill-selection whereas in helm-map it is bound to
helm-kill-selection-and-quit, so the same key is C-c C-k which is bound
to two different commands.


>     it is here killing the real value (i.e the whole candidate not truncated)
>     without the need of a prefix argument.
>    
> We must explain what "truncating" means.

As you have maybe noticed, candidates in helm-kill-ring are truncated
according to helm-kill-ring-max-offset.

>     Note: Of course numeric prefix args like `yank' are not available due to the nature
>     of helm which allow you to see and move to the desired candidate
>    
> "yank" is not a numeric prefix arg... What do you mean?

Read numeric prefix args like in `yank' command.

> I think you mean something like this:
>
> "As opposed to yank', numeric prefix arguments are ignored with `helm-show-kill-ring' as there is no need for them since selection happens within Helm."

Yes, thanks you ;-)

>     and also because
>     it is already available in helm with [[Shortcuts for executing Default Action on the nth candidate][Shortcuts for executing Default Action on the nth candidate]].
>    
> What is available in helm?

Selection of candidates by numeric shortcuts, i.e to select and execute
default action on the 3rd candidate from your current selection you can
use C-c 3, to execute the default action on the 3rd candidate before
your selection you can hit C-x 3.
So implementing C-u 3 RET here would be useless.

Thanks to look into this.

-- 
Thierry

PR #1864, opened 2017-09-15
Improve phrasing in helm-kill-ring documentation

Ambrevar:


helm-help.el
```diff
@@ -1411,42 +1411,42 @@ condition that you don't choose a command using Helm completion.
...
 (defvar helm-kill-ring-help-message
-  "* Helm kill-ring
+  "* Helm kill ring
 
 ** Tips
 
-You can bring any candidate on top of kill-ring by using `\\<helm-map>\\[helm-kill-selection-and-quit]'
-on it, this allow for example to paste this candidate using x-selection in an external application like
-firefox.
+Every Helm session lets you save a candidate to the kill-ring / clipboard /
```
thierryvolpiatto: Note that candidate is already saved to kill-ring, C-c C-k is just pushing it on top of stack so that it appear in clipboard/primary selection

Ambrevar: - I am talking about all Helm sessions, not necessarily Helm-kill-ring sessions, so the candidate might not be on the stack.

- I could be wrong, but to be rigorous, it's _killing the text_ that places saves it to the clipboard/primary + pushes it on top of the kill ring.  If you modify the kill-ring manually to change the top, it will no alter the clipboard/primary.

thierryvolpiatto: Ok, right.

thierryvolpiatto:
Merged, thanks.

Issue #755, opened 2014-12-13
Helm buffers not using the "not saved" face properly


pronobis:
Currently, helm-buffers using `helm-ff-symlink` face to mark unsaved buffers, while `helm-buffer-not-saved` exists. This is the relevant piece of code:

```
      ((and file-name (buffer-modified-p buf))
       (helm-buffer--show-details
        name name-prefix file-name size mode dir
        'helm-ff-symlink 'helm-buffer-process nil details 'mod))
```

Shouldn't that be changed to `helm-buffer-not-saved`?

PR #1866, opened 2017-09-17
Use separate face for modified buffers

rcj:
In a buffer overview, modified buffers were shown using the
helm-ff-symlink face. This is slightly irritating if this face is
customised and the customisation isn't a good fit for modified
buffers.

Using helm-buffer-modified instead, inheriting from helm-ff-symlink,
makes it possible to style modified buffers independently from
symlinks.

Issue: #755


thierryvolpiatto:
Thanks, merged.

PR #1871, opened 2017-09-18
Add helm-switch-to-buffers-function and perform alternate splits

Ambrevar:


Ambrevar:
Wait, I think the function composition is not right with respect to the ELSE part of `helm-aif`.  I'm working on it.

thierryvolpiatto:

> Wait, I think the function composition is not right with respect to
> the ELSE part of helm-aif. I'm working on it.

Yes, the function composition should handle only the case with more than
1 marked candidates i.e it should contain only the "then" clause of helm-aif.
>
> —


--
Thierry

Ambrevar:
I've updated the code following the changes made upstream.
There are still some bugs I'm trying to fix.  Stay tuned.

In the mean time, could you tell me how to hook the new defcustom function into `helm-ff-RET`?

thierryvolpiatto:

> I've updated the code following the changes made upstream.
> There are still some bugs I'm trying to fix. Stay tuned.
>
> In the mean time, could you tell me how to hook the new defcustom
> function into helm-ff-RET?

I think the function you want to modify to apply your changes to HFF is
`helm-simultaneous-find-file`.

--
Thierry

thierryvolpiatto:
And also `helm-find-files-other-window`.

thierryvolpiatto:
This remind me I didn't make the error handling in `helm-find-files-other-window` and I see that I could arrange to reuse the same code than in `helm-simultaneous-find-file` for this.
I will let you do your changes first and will apply changes after to not complicate the merge with another conflict.

thierryvolpiatto:
I have pushed some changes to a new branch, you can perhaps cherry-pick changes there before working on HFF.

Ambrevar:
I think the function composition with regard to helm-utils and `helm-simultaneous-find-file` is not ideal: lots of code duplication.

In `helm-simultaneous-find-file`, wouldn't it be simpler to open all files in the background, then hand the list of buffers to `helm-switch-to-buffers`?

thierryvolpiatto:

> I think the function composition with regard to helm-utils and
> helm-simultaneous-find-file is not ideal: lots of code duplication.
>
> In helm-simultaneous-find-file, wouldn't it be simpler to open all
> files in the background, then hand the list of buffers to
> helm-switch-to-buffers?

Yes, of course, if it helps do like this.

--
Thierry

Ambrevar:
`helm-switch-to-buffers` seems to only be called by `helm-display-buffers-other-windows`.

- Shall we merge the two functions?
- Is the `other-window` parameter always true?

`helm-simultaneous-find-file` cannot call `helm-switch-to-buffers` because then the marked candidates are files and not buffers.  I suggest we refactor the code as follows:

- `helm-switch-to-buffers` takes a list of buffers as argument.
- Every call to it has to be refactored to pass a list of buffers.  In practice, this only happens at two spots: `helm-simultaneous-find-file` and `helm-buffer-switch-other-window`.

thierryvolpiatto:

> helm-switch-to-buffers seems to only be called by helm-display-buffers-other-windows.
>
>   * Shall we merge the two functions?

No.

>   * Is the other-window parameter always true?

No.

> helm-simultaneous-find-file cannot call helm-switch-to-buffers because
> then the marked candidates are files and not buffers.

You can generate your buffers in background with something like 
`(mapcar 'find-file-noselect marked)` and then pass this list to
`helm-switch-to-buffers`, no?.


> I suggest we refactor the code as follows:
>
>   * helm-switch-to-buffers takes a list of buffers as argument.

Yes, see above.

>   * Every call to it has to be refactored to pass a list of buffers.

Or make it optional and provide a default when unspecified?

>   In practice, this only happens at two spots:
>   helm-simultaneous-find-file and helm-buffer-switch-other-window.
>
> —


-- 
Thierry

Ambrevar:
> >  * Shall we merge the two functions?
> No.
> >   * Is the other-window parameter always true?
> No.

Could you be more explicit?  I don't see where else `helm-switch-to-buffers` is called, and I don't see the point of having a function just calling another one.

Where is `helm-switch-to-buffers` called with `other-window=t`?

thierryvolpiatto:

> I don't see where else helm-switch-to-buffers is called, and I don't
> see the point of having a function just calling another one.

Do what arrange you the best for what you are actually writing, we will
see once done, if you feel there is no need for a function and you can
replace it in another way, go ahead.

> Where is helm-switch-to-buffers called with other-window=t?

AFAICT in only one place.


-- 
Thierry

thierryvolpiatto:
I have now merged helm-find-files-other-window in master (I will tag helm soon and I want this merged in master).

Ambrevar:
Will do as soon as possible.

Ambrevar:
I've commited some working changes:
1. Typos and syntax while I was at it.
2. Removed helm-simultaneous: it duplicates most of the code found in helm-utils.
3. I've moved to core mechanism of `helm-switch-to-buffers` to `helm-window-show-buffers`.  Now the following functions rely on it:
    - `helm-switch-to-buffers`
    - `helm-display-buffers-other-window`
    - `helm-find-files-or-marked`
    - `helm-find-files-other-window`
4. I think the naming difference between `helm-switch-to-buffers` and `helm-display-buffers-other-window` is misleading.  Besides they are only used by the helm buffer menu now, so I think we should move them to _helm-buffers.el_.  That would be more consistent with _helm-ff_.
5. Included 2 example functions: alternate-split and mosaic.  The latter is still unfinished.

Let me know what you think, in particular about point 4.  Renaming functions breaks backward compatibility, but you recently did this in a80aeccb7ad61c4eb57248d423265713e335070d.

Ambrevar:
I've polished `helm-window-mosaic-fn', it should work well now.

thierryvolpiatto:

> I've commited some working changes:
>
>  1. Typos and syntax while I was at it.
>  2. Removed helm-simultaneous: it duplicates most of the code found in helm-utils.
>  3. I've moved to core mechanism of helm-switch-to-buffers to helm-window-show-buffers. Now the following functions rely on it:
>       + helm-switch-to-buffers
>       + helm-display-buffers-other-window
>       + helm-find-files-or-marked
>       + helm-find-files-other-window
>  4. I think the naming difference between helm-switch-to-buffers and
>     helm-display-buffers-other-window is misleading. Besides they are
>     only used by the helm buffer menu now, so I think we should move
>     them to helm-buffers.el. That would be more consistent with
>     helm-ff.

Agree.

>  5. Included 2 example functions: alternate-split and mosaic. The
>  latter is still unfinished.

I think you fixed the mosaic fn, seems to work fine here.
The default fn seems to work fine as well.
The alternate-split fn however seems to duplicate the first window.

> Let me know what you think, in particular about point 4. Renaming
> functions breaks backward compatibility, but you recently did this in
> a80aecc.

This is not a problem, at least for this, the previous changes are
recent enough to be changed.

Thanks for your work.

--
Thierry

thierryvolpiatto:
Perhaps the defcustom should propose the functions to user with something like this:

``` elisp
(defcustom helm-window-show-buffers-function #'helm-window-default-split-fn
  "The default function to use when opening several buffers at once.
It is typically used to rearrange windows."
  :group 'helm-utils
  :type '(choice
          (function
           (const :tag "Split windows vertically or horizontally"
                  #'helm-window-default-split-fn)
           (const :tag "Split in alternate windows"
                  #'helm-window-alternate-split-fn)
           (const :tag "Split windows in mosaic"
                  #'helm-window-mosaic-fn))))
```

Ambrevar:
I've refactored the function names, I've fixed the issue with the alternate splits and I've added the 3 choices to the defcustom.

thierryvolpiatto:
Many thanks for your work, merged.

PR #1883, opened 2017-09-25
Pull out the logic for  pipe commands into a separate function.

zhenya1007:
Having a separate function allows the user to advise and/or re-define
just that function if they let-bind `helm-grep-default-command' around
a call to `helm-do-grep-1' and friends, but still want to have the
"space-separated patterns pipe to grep" functionality.  (In my case, I
let-bound it to command that runs codesearch to make a quick-and-dirty
Emacs interface for that command).

### DIFF
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -420,8 +420,8 @@ It is intended to use as a let-bound variable, DON'T set this globaly.")
                       (or (and norm-com norm-com-ack-p)
                           (and rec-com rec-com-ack-p)))))))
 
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks, but you have free variables in your new function `smart` and `pipe-switches`.
Compile helm with make should show you warnings for this.

zhenya1007:
Thank you for pointing this out.   The patch I have just pushed should address that.

thierryvolpiatto:
Thanks, merged.

PR #1888, opened 2017-10-06
Multiline colouration

Ambrevar:
Don't merge yet!

I've implemented multiline colouration by re-using the work you did.
I'll keep experimenting and foolproofing the code.

In the meantime, could you have a look at my changes?  Please question anything you find dubious.
I am discovering new grounds here, so any guidance is more than welcome :)

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -4765,17 +4765,18 @@ candidates."
       (goto-char helm-selection-point))
     (move-overlay
      helm-selection-overlay (point-at-bol)
-     (let ((header-pos (helm-get-next-header-pos))
-           (separator-pos (helm-get-next-candidate-separator-pos)))
-       (if (helm-pos-multiline-p)
+     (if (helm-pos-multiline-p)
+         (let ((header-pos (helm-get-next-header-pos))
+               (separator-pos (helm-get-next-candidate-separator-pos)))
            (or (and (null header-pos) separator-pos)
                (and header-pos separator-pos
                     (< separator-pos header-pos)
                     separator-pos)
                header-pos
-               (point-max))
-         (helm-aif (next-single-property-change (point) 'helm-candidate)
-             it
+               (point-max)))
+       (helm-aif (next-single-property-change (point) 'helm-candidate)
+           it
+         (let ((header-pos (helm-get-next-header-pos)))
            (or header-pos (point-max))))))
     (setq helm-selection-point (overlay-start helm-selection-overlay))
     (when (and helm-allow-mouse (null nomouse))
--- a/helm.el
+++ b/helm.el
@@ -3625,11 +3625,11 @@ It is used for narrowing list of candidates to the
                         (helm-insert-candidate-separator)
                       (setq separate t)))
                   (helm-insert-match m 'insert count source)
-                  (put-text-property beg (point) 'helm-candidate
-                                     (if (cl-oddp count) 'odd 'even))
-                  (and (not helm-use-multiline-separator)
-                       (cl-evenp count)
-                       (add-face-text-property beg (point) 'helm-selection-even t)))
+                  (unless helm-use-multiline-separator
+                    (put-text-property beg (point) 'helm-candidate
+                                       (if (cl-oddp count) 'odd 'even))
+                    (when (cl-evenp count)
+                      (add-face-text-property beg (point) 'helm-selection-even t))))
              finally (and (null singleline)
                           (put-text-property start (point)
                                              'helm-multiline t)))))
@@ -4493,6 +4493,8 @@ Key arg DIRECTION can be one of:
         (funcall move-func)
         (and (memq direction '(next previous))
              (helm-skip-noncandidate-line direction))
+        (and helm-use-multiline-separator (helm-pos-multiline-p)
+             (helm-move--beginning-of-multiline-candidate))
         (helm-display-source-at-screen-top-maybe where)
         (helm-mark-current-line)
         (when follow
@@ -4512,7 +4514,8 @@ Key arg DIRECTION can be one of:
         (forward-line 1)))))
 
 (defun helm-move--previous-multi-line-fn ()
-  (forward-line -1)
+  (when helm-use-multiline-separator
+    (forward-line -1))
   (unless (helm-pos-header-line-p)
     (helm-skip-header-and-separator-line 'previous)
     (helm-move--beginning-of-multiline-candidate)))
--- a/helm.el
+++ b/helm.el
@@ -758,6 +758,22 @@ so it have only effect when `helm-always-two-windows' is non-nil."
   "Face for currently selected item in the helm buffer."
   :group 'helm-faces)
 
+(defface helm-selection-even
+  '((((background dark)) :background "DeepSkyBlue1"
+     :distant-foreground "black")
+    (((background light)) :background "turquoise1"
+     :distant-foreground "black"))
+  "Face for even items in the Helm multiline buffer."
+  :group 'helm-faces)
+
+(defface helm-selection-odd
+  '((((background dark)) :background "DeepSkyBlue3"
+     :distant-foreground "black")
+    (((background light)) :background "turquoise3"
+     :distant-foreground "black"))
+  "Face for odd items in the Helm multiline buffer."
+  :group 'helm-faces)
+
 (defface helm-separator
   '((((background dark)) :foreground "red")
     (((background light)) :foreground "#ffbfb5"))
@@ -1903,7 +1919,9 @@ only."
                                              (forward-line 2)
                                              (or (helm-pos-header-line-p) (eobp)))
                                            (eobp)))
-                                  (search-forward helm-candidate-separator nil t))
+                                  (if helm-use-multiline-separator
+                                      (search-forward helm-candidate-separator nil t)
+                                    (helm--forward-candidate)))
                        do (cl-incf count-multi)
                        finally return count-multi)
               (cl-loop with ln = 0
@@ -3590,6 +3608,8 @@ It is used for narrowing list of candidates to the
 (defun helm--candidates-in-buffer-p (matchfns)
   (equal matchfns '(identity)))
 
+(defvar helm-use-multiline-separator t)
+
 (defun helm-render-source (source matches)
   "Display MATCHES from SOURCE according to its settings."
   (helm-log "Source name = %S" (assoc-default 'name source))
@@ -3607,11 +3627,16 @@ It is used for narrowing list of candidates to the
                   (put-text-property beg (point) 'helm-candidate
                                      (if (cl-oddp count) 'odd 'even)))
              else
-             do (progn
-                  (if separate
-                      (helm-insert-candidate-separator)
-                    (setq separate t))
-                  (helm-insert-match m 'insert count source))
+             do (let ((beg (point)))
+                  (when helm-use-multiline-separator
+                    (if separate
+                        (helm-insert-candidate-separator)
+                      (setq separate t)))
+                  (helm-insert-match m 'insert count source)
+                  (put-text-property beg (point) 'helm-candidate
+                                     (if (cl-oddp count) 'odd 'even))
+                  (unless helm-use-multiline-separator
+                    (add-face-text-property beg (point) (if (cl-oddp count) 'helm-selection-odd 'helm-selection-even) t)))
              finally (and (null singleline)
                           (put-text-property start (point)
                                              'helm-multiline t)))))
@@ -4475,8 +4500,6 @@ Key arg DIRECTION can be one of:
         (funcall move-func)
         (and (memq direction '(next previous))
              (helm-skip-noncandidate-line direction))
-        (when (helm-pos-multiline-p)
-          (helm-move--beginning-of-multiline-candidate))
         (helm-display-source-at-screen-top-maybe where)
         (helm-mark-current-line)
         (when follow
@@ -4492,7 +4515,8 @@ Key arg DIRECTION can be one of:
                          (< separator-pos header-pos))
                      header-pos
                      separator-pos))
-      (forward-line 1))))
+      (when helm-use-multiline-separator
+        (forward-line 1)))))
 
 (defun helm-move--previous-multi-line-fn ()
   (forward-line -1)
@@ -4877,7 +4901,7 @@ If action buffer is displayed, kill it."
 (defun helm-get-next-candidate-separator-pos ()
   "Return the position of the next candidate separator from point."
   (let ((hp (helm-get-next-header-pos)))
-    (helm-aif (next-single-property-change (point) 'helm-candidate-separator)
+    (helm-aif (next-single-property-change (point) (if helm-use-multiline-separator 'helm-candidate-separator 'helm-candidate))
         (or
          ;; Be sure we don't catch
          ;; the separator of next source.
@@ -4890,7 +4914,7 @@ If action buffer is displayed, kill it."
 
 (defun helm-get-previous-candidate-separator-pos ()
   "Return the position of the previous candidate separator from point."
-  (previous-single-property-change (point) 'helm-candidate-separator))
+  (previous-single-property-change (point) (if helm-use-multiline-separator 'helm-candidate-separator 'helm-candidate)))
 
 (defun helm-pos-header-line-p ()
   "Return t if the current line is a header line."
--- a/helm.el
+++ b/helm.el
@@ -759,19 +759,11 @@ so it have only effect when `helm-always-two-windows' is non-nil."
   :group 'helm-faces)
 
 (defface helm-selection-even
-  '((((background dark)) :background "DeepSkyBlue1"
+  '((((background dark)) :background "DimGrey"
      :distant-foreground "black")
-    (((background light)) :background "turquoise1"
+    (((background light)) :background "LightGrey"
      :distant-foreground "black"))
   "Face for even items in the Helm multiline buffer."
-  :group 'helm-faces)
-
-(defface helm-selection-odd
-  '((((background dark)) :background "DeepSkyBlue3"
-     :distant-foreground "black")
-    (((background light)) :background "turquoise3"
-     :distant-foreground "black"))
-  "Face for odd items in the Helm multiline buffer."
   :group 'helm-faces)
 
 (defface helm-separator
@@ -3635,8 +3627,9 @@ It is used for narrowing list of candidates to the
                   (helm-insert-match m 'insert count source)
                   (put-text-property beg (point) 'helm-candidate
                                      (if (cl-oddp count) 'odd 'even))
-                  (unless helm-use-multiline-separator
-                    (add-face-text-property beg (point) (if (cl-oddp count) 'helm-selection-odd 'helm-selection-even) t)))
+                  (and (not helm-use-multiline-separator)
+                       (cl-evenp count)
+                       (add-face-text-property beg (point) 'helm-selection-even t)))
              finally (and (null singleline)
                           (put-text-property start (point)
                                              'helm-multiline t)))))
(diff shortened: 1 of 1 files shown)


alphapapa:
@Ambrevar Please test this with `helm-org-rifle`, as it depends on the separator and multi-line candidates.  Thanks.

thierryvolpiatto:

> Don't merge!
>
> I've implemented multiline colouration by re-using the work you did.
> I'll keep experimenting and foolproofing the code.

Thanks for working on this.
The changes I made was working quite well but had some problems in some
places, don't remember what though.

> In the meantime, could you have a look at my changes?

Yes, ASAP.

-- 
Thierry

Ambrevar:
@alphapapa: I've tested against `helm-show-kill-ring`.  Why `helm-org-rifle` in particular?

Ambrevar:
I've updated the changes:
- The colours have now their own faces.
- Movements are working properly it seems.

I think the existing code is overly complicated and I'm afraid my patch is not helping.  A few suggestions of what we could do:

- Remove the separator code if you agree that colour alternation is superior.
- Untangled the nested function calls.

Ambrevar:
@thierryvolpiatto
In your original patch, you added this:
```
             do (let ((beg (point)))
                  (helm-insert-match m 'insert count source)
                  (put-text-property beg (point) 'helm-candidate
                                     (if (cl-oddp count) 'odd 'even)))
```
Why putting odd/even properties on singlelines?

thierryvolpiatto:

> I've updated the changes:
>
>   * The colours have now their own faces.

Why two faces, only one face seems enough

>   * Movements are working properly it seems.
>
> I think the existing code is overly complicated and I'm afraid my
> patch is not helping. A few suggestions of what we could do:
>
>   * Remove the separator code if you agree that colour alternation is
>   superior.

No please keep the original code as it is, also I don't think using a
var to toggle face and separator for multiline is the right thing to do,
using the :multiline slot seems to be a better alternative allowing
sources with separator and others with alternate faces.
E.g. ":multiline 'face" or ":multiline 'separator", ":multiline t"
should be equivalent of ":multiline separator" for backward
compatibility.

--
Thierry

thierryvolpiatto:

> @thierryvolpiatto
> In your original patch, you added this:
>
>              do (let ((beg (point)))
>                   (helm-insert-match m 'insert count source)
>                   (put-text-property beg (point) 'helm-candidate
>                                      (if (cl-oddp count) 'odd 'even)))
>
> Why putting odd/even properties on singlelines?

Don't remember the details, but I guess we need the property on single
line candidates as well because they should be present among
other multiline candidates.

What we have to do on this issue is first merge the original code
(fix_multiline branch) again into master and see what's wrong with and
without multiline, when this code will be fixed let's apply your patch
on top of this.

Perhaps next week I will be more quiet to look into this more carefully.

Thanks.

-- 
Thierry

Ambrevar:
> Why two faces, only one face seems enough

Well, one for odd lines, one for even lines... Did you mean something else?

I think we are misunderstood in what we want to achieve: you want sources to define how they display (separator or not) through the slot.
I don't see the point in enforcing _appearance_ to the user.  I want to be able to use `helm-show-kill-ring` with colors and not separators.

> What we have to do on this issue is first merge the original code (fix_multiline branch) again into master and see what's wrong with and without multiline, when this code will be fixed let's apply your patch on top of this.

Well, that's what I did: first commit re-applied your changes, second one fixes them (to the extend I've tested) and adds the variables/faces.

alphapapa:
> I've tested against helm-show-kill-ring. Why helm-org-rifle in particular?

@Ambrevar Because it uses the separators and multi-line candidates in a less-common way.  I would like for it to not be broken by this patch.

Ambrevar:
> Don't remember the details, but I guess we need the property on single
> line candidates as well because they should be present among
> other multiline candidates.

The `singleline` is not a property of candidates but of the source.  A singleline source will never need colors I believe.
As a matter of fact, the code does not deal with separators, so why should it deal with colors?

Ambrevar:
@alphapapa: Just tested with helm-org-rifle and it seems to work without an itch.

Ambrevar:
> > Why two faces, only one face seems enough
> Well, one for odd lines, one for even lines... Did you mean something else?

Duh, I guess you meant we can leave odd entries to the normal background color and only color even entries.  Right?  I think it's much better indeed.
I suggest dark gray and light gray for dark background and light background respectively.

thierryvolpiatto:

>         Why two faces, only one face seems enough
>         Well, one for odd lines, one for even lines... Did you mean something else?
>
> Duh, I guess you meant we can leave odd entries to the normal
> background color and only color even entries. Right?

Yes.

> I think it's much better indeed.  I suggest dark gray and light gray
> for dark background and light background respectively.

As you want, take care however to not use something similar to
helm-selection.


--
Thierry

thierryvolpiatto:

>     Don't remember the details, but I guess we need the property on single
>     line candidates as well because they should be present among
>     other multiline candidates.
>
> The singleline is not a property of candidates but of the source. A
> singleline source will never need colors I believe.

Forget colors for now.

You may have candidates with multilines without having specified the
`:multiline` keyword, e.g.

    (helm :sources (helm-build-sync-source "test"
                     :candidates '("foo\nbar" "baz\nfoo" "bar\nfoo"))
          :buffer "*helm test*")

So here a list of 3 candidates which is displayed as a list of 6
candidates in current master branch.

The work I did previously fixed this.

> As a matter of fact, the code does not deal with separators, so why
> should it deal with colors?

I didn't have the time to investigate much, but I think we should decide
of the semantic for the :multiline keyword and where the colorisation
should happen (not sure yet if helm-render-source is the good place).

I hope I will have more time next week to look into all this more
seriously.

Thanks.
 
--
Thierry

alphapapa:
@Ambrevar Thank you for testing it with helm-org-rifle.  :)

Forgive me if I missed this in the conversation, but please take care to not require applying the alternate colors.  For example, helm-org-rifle would not need that, and it would cause problems if it were enabled.

Ambrevar:
It does not, in fact it works with colors too, albeit useless since helm-org-rifle is using its own separator.

As the maintainer of helm-org-rifle, why don't you want to follow upstream?  I understand you like empty separators better, but ultimately I believe that choice should be left to the user.

Ambrevar:
I've removed the colouration of odd multilines.

alphapapa:
> As the maintainer of helm-org-rifle, why don't you want to follow upstream? 

I'm not sure why you would think that I don't want to follow upstream.  I've worked with Thierry and contributed patches to Helm myself.

> I understand you like empty separators better, but ultimately I believe that choice should be left to the user.

When I first created helm-org-rifle, the default Helm multi-line-candidate separator was a plain string of red `----------`.  It didn't look good in helm-org-rifle results, especially because each result is a propertized Org entry with fontified heading, which provides its own visual separation, just like in an Org buffer.  So I set the separator to an empty string.

In the `helm-org-rifle-occur` commands, there is a separator, which is also a blank string, but is fontified with a customizable face.  This is important since results in that command are persistent, can be manipulated, collapsed, deleted, etc.

This sounds like a nice change for most Helm commands, but I don't think it would be good to use it in helm-org-rifle, because its results are fontified the same way they appear in Org buffers, and this would interfere with that.  So I hope that I won't have to work around this in my own code.

thierryvolpiatto:

> This sounds like a nice change for most Helm commands, but I don't
> think it would be good to use it in helm-org-rifle, because its
> results are fontified the same way they appear in Org buffers, and
> this would interfere with that. So I hope that I won't have to work
> around this in my own code.

Nothing will change, I want the default to stay the same as before.

Also there is one new variable `helm-use-multiline-separator` that could be
removed by using only `helm-multiline-separator` with a different value,
like a symbol (face?) instead of a string.

Originally I wanted to use setting by source only using the
:multiline slot, but it looks complicated as we have to handle the
truncated width as well in this slot. 


--
Thierry

Ambrevar:
@alphapapa:  Sorry for the misunderstanding, I did not mean to undermine your valuable contribution to Helm, far from that! :)
My question was specifically about the separator of helm-org-rifle which does not follow upstream.  But you've answered that, so all good.

> because its results are fontified the same way they appear in Org buffers, and this would interfere with that.

There is no interference: 1. The original code is still here. 2. My patch changes the _background_ colours, thus this won't interfere with any of the usual org fontification.

@thierryvolpiatto: What is `helm-multiline-separator`?

Ambrevar:
@alphapapa: 
> It didn't look good in helm-org-rifle results, especially because each result is a propertized Org entry ? with fontified heading, which provides its own visual separation, just like in an Org buffer. So I set the separator to an empty string.

If headers are always starting entries, why not removing the separator altogether?

alphapapa:
>  My patch changes the background colours

That might be helpful in helm-org-rifle, so I look forward to trying that.  The hard part may be getting the color right, since it depends on the user's theme.  Is there a standard Emacs color for that?  I don't think the standard `highlight` face would be a good choice, because, of course, that is how the currently selected item is shown.

> If headers are always starting entries, why not removing the separator altogether?

Because it doesn't look good with them displayed so close together.  This is especially so because headers are independent, so they may vary wildly in size, boldness, color, etc, and it looks confusing to have a small, deep heading displayed immediately before a large, shallow heading.  Also, entries have entry text, not just heading text, so without a blank line, a line of entry text might be closer to the next heading than to its own heading.

Anyway, you can experiment with the code and remove it if you don't like it.

thierryvolpiatto:

> @thierryvolpiatto: What is helm-multiline-separator?

Sorry I meant `helm-candidate-separator`.

But I am not really sure on how to enable this for now.

I saw that your patch is working with the alternate faces but breaks
multiline with separator, selection becomes quickly erratic after moving
up and down, wrong candidate is selected when using
helm-move-to-line-cycle-in-source, only first line of some candidates is
selected etc...

-- 
Thierry

Ambrevar:
I can reproduce two issues:

- cycle on, separator off: selection jumps back to beginning on some entries.

- cycle on, seperator on: cycling backwards from first candidate to last, the last entry is not properly selected.

Ambrevar:
I've fixed both issues (first description was inacurate, see the commit).

thierryvolpiatto:

> I've fixed both issues (first description was inacurate, see the commit).

Thanks, looks fixed indeed.

However, next/previous-line is now really slow, this is not your additions
that make helm slow but the initial patch (1/4) which need to be
reworked to make helm fast again.

To feel the difference, use master and your branch with M-x
helm-list-elisp-packages-no-fetch and try to move around with C-n/p.

-- 
Thierry

Ambrevar:
I can't reproduce the performance issue.
I've tried with `emacs-helm.sh` and I get the same speed both on master and my branch.

thierryvolpiatto:

> I can't reproduce the performance issue.
> I've tried with emacs-helm.sh and I get the same speed both on master and my branch.

You need more than 5000 candidates and keep pressing <down> or <up> to
really see the difference.

The problem is `next-single-property-change` which is slow.

-- 
Thierry

Ambrevar:
OK, I can reproduce with 4000 candidates.  Let's investigate.

Ambrevar:
Here is the profile I get when I fire up `helm-list-elisp-packages-no-fetch` and keep pressing `C-n` for a few seconds:

```
- command-execute                                                4088  93%
 - call-interactively                                            4088  93%
  - funcall-interactively                                        3967  90%
   - eval-last-sexp                                              3959  90%
    - elisp--eval-last-sexp                                      3959  90%
     - eval                                                      3959  90%
      - helm-list-elisp-packages-no-fetch                        3959  90%
       - helm-list-elisp-packages                                3959  90%
        - helm                                                   3959  90%
         - apply                                                 3959  90%
          - helm                                                 3959  90%
           - apply                                               3959  90%
            - helm-internal                                      3959  90%
             - helm-read-pattern-maybe                           3629  83%
              - read-from-minibuffer                             3433  78%
               - command-execute                                 3325  76%
                - call-interactively                             3325  76%
                 - funcall-interactively                         3247  74%
                  - helm-next-line                               3247  74%
                   - helm--next-or-previous-line                 3247  74%
                    - helm-move-selection-common                 3247  74%
                     - helm-mark-current-line                    1944  44%
                      - helm-get-next-candidate-separator-pos               1276  29%
                         helm-get-next-header-pos                 660  15%
                        helm-get-next-header-pos                  660  15%
                      - helm-window                                 4   0%
                       - helm-buffer-get                            4   0%
                          helm-action-window                        4   0%
                     - helm-move--next-line-fn                   1291  29%
                      - helm-get-next-candidate-separator-pos               1291  29%
                         helm-get-next-header-pos                 655  14%
                       helm-display-mode-line                       4   0%
                     + #<compiled 0x1d2bfab>                        4   0%
                 + byte-code                                       78   1%
               + redisplay_internal (C function)                   68   1%
              + helm-update                                       184   4%
                helm-get-candidate-number                           4   0%
             + helm-initialize                                    326   7%
             + #<compiled 0xc5c105>                                 4   0%
   + helm-M-x                                                       8   0%
  + byte-code                                                     121   2%
+ ...                                                             275   6%
+ redisplay_internal (C function)                                   4   0%
```

Ambrevar:
Same profiling procedure on master:
```
- command-execute                                                2916  91%
 - call-interactively                                            2916  91%
  - funcall-interactively                                        2761  86%
   - eval-last-sexp                                              2761  86%
    - elisp--eval-last-sexp                                      2761  86%
     - eval                                                      2761  86%
      - helm-list-elisp-packages-no-fetch                        2761  86%
       - helm-list-elisp-packages                                2761  86%
        - helm                                                   2761  86%
         - apply                                                 2761  86%
          - helm                                                 2761  86%
           - apply                                               2761  86%
            - helm-internal                                      2761  86%
             - helm-read-pattern-maybe                           2442  76%
              - read-from-minibuffer                             2272  70%
               + redisplay_internal (C function)                 1172  36%
               - command-execute                                    4   0%
                - call-interactively                                4   0%
                 - funcall-interactively                            4   0%
                  - helm-next-line                                  3   0%
                   - helm--next-or-previous-line                    3   0%
                    - helm-move-selection-common                    3   0%
                       helm-follow-execute-persistent-action-maybe                  3   0%
               + timer-event-handler                                2   0%
              + helm-update                                       142   4%
                helm-get-candidate-number                           8   0%
             + helm-initialize                                    313   9%
             + #<compiled 0x9efaf7>                                 6   0%
  + byte-code                                                     155   4%
+ ...                                                             288   8%
```

Ambrevar:
I've fixed the issue.]I've rebased my changes over your master branch since there were some conflicts.  The fix commit is on my master branch, but somehow GitHub does not see the new commit on this PR.  Am I missing something or should I open a new PR?

thierryvolpiatto:

> I've fixed the issue.]I've rebased my changes over your master branch
> since there were some conflicts. The fix commit is on my master
> branch, but somehow GitHub does not see the new commit on this PR.

I can see it from here, I will try and look at your new patch ASAP.
Thanks.

> Am I missing something or should I open a new PR?

Perhaps reload the github page ?

--
Thierry

thierryvolpiatto:
Thierry Volpiatto <thievol05@zoho.eu> writes:

> Pierre Neidhardt <notifications@github.com> writes:
>
>> I've fixed the issue.

It is not fixed here it still unusable on large set of candidates.
The problem is coming from `helm--forward-candidate` and
`helm--backward-candidate` that are calling
`next-single-char-property-change` and `previous-single-property-change`
which are really slow.

Sorry to not help more, but I have not much time to read/write code
actually apart for trivial changes, perhaps next week I will have some
time to work on this.

-- 
Thierry

Ambrevar:
> Perhaps reload the github page ?

It took ages to refresh.

> It is not fixed here it still unusable on large set of candidates.

I'll see if I can reproduce.

Ambrevar:
I cannot reproduce: it's really fast with my patch.  Here is the profile:

```
- command-execute                                                3950  92%
 - call-interactively                                            3950  92%
  - funcall-interactively                                        3806  89%
   - eval-last-sexp                                              3802  89%
    - elisp--eval-last-sexp                                      3802  89%
     - eval                                                      3802  89%
      - helm-test-init                                           3802  89%
       - helm-list-elisp-packages-no-fetch                       3653  85%
        - helm-list-elisp-packages                               3653  85%
         - helm                                                  3653  85%
          - apply                                                3653  85%
           - helm                                                3653  85%
            - apply                                              3653  85%
             - helm-internal                                     3653  85%
              - helm-read-pattern-maybe                          3337  78%
               - read-from-minibuffer                            3146  74%
                - command-execute                                1590  37%
                 - call-interactively                            1590  37%
                  - funcall-interactively                        1590  37%
                   - helm-next-line                              1590  37%
                    - helm--next-or-previous-line                1590  37%
                     - helm-move-selection-common                1590  37%
                      - helm-move--next-line-fn                  1578  37%
                       - helm-get-next-candidate-separator-pos               1574  37%
                          helm-get-next-header-pos                810  19%
                        helm-display-mode-line                     12   0%
                + redisplay_internal (C function)                 973  22%
               + helm-update                                      156   3%
                 helm-get-candidate-number                         12   0%
              + helm-initialize                                   310   7%
              + #<compiled 0x90a613>                                6   0%
       + package-refresh-contents                                 147   3%
       + package-initialize                                         2   0%
   + helm-M-x                                                       4   0%
  + byte-code                                                     144   3%
+ ...                                                             300   7%

```

Ambrevar:
On master, `redisplay_internal` is the bottleneck at 36% while `command-execute` takes 0%.
With my patch, `redisplay_internal` takes 22% while `command-execute` takes 37%.

The slow part is seemingly around the new `helm-get-next-candidate-separator-pos`. I'll see how we can optimize this.

Ambrevar:
So the offending part is this:
```
(defun helm-move--next-line-fn ()
  (cond ((and (not (helm-pos-multiline-p))
              (helm-get-next-candidate-separator-pos))
        ((not (helm-pos-multiline-p))
         (helm--forward-candidate))
        (t (helm-move--next-multi-line-fn)))
```

And more specifically:
```
  (cond ((and (not (helm-pos-multiline-p))
              (helm-get-next-candidate-separator-pos))
```

While I understand the last 2 conditions, why the first one?
If this is not multiline, then there should be no separator, or am I mistaken?

thierryvolpiatto:

> While I understand the last 2 conditions, why the first one?
> If this is not multiline, then there should be no separator, or am I mistaken?

helm-pos-multiline-p tells you if you are on a separator line, not you
are using multiline or not.

-- 
Thierry

Ambrevar:
That's a confusing name then :p

Can you detail the meaning of the algorithm?  Here is what I understand:

Cond:

- If not on a separator line (no cost) and there is a separator somewhere down below (scans the whole buffer, super costly `O(n)`), forward line.  Why would we do that?  Why the check and why forwarding line?

- If not on a separator, go to next candidate.  Makes sense.

- If on a separator, move to next multiline.  Mes sense.

Assuming first condition is really needed, a possible fix would be to check if it's a multiline source and we are last candidate.  But I still don't get the point.

Ambrevar:
Any insight?

thierryvolpiatto:

> Any insight?

I have no idea on how to implement it for now (faster than what we
have).

-- 
Thierry

Ambrevar:
Can you explain the intention of the first condition mentioned above?

Also here is an optimization idea, no clue if that'd be faster though. When looking for the next separator,
```
(defun helm-get-next-candidate-separator-pos ()
  "Return the position of the next candidate separator from point."
  (let ((hp (helm-get-next-header-pos)))
    (helm-aif (next-single-property-change (point) 'helm-candidate-separator)
```

Instead of scanning all characters, why not scanning the first character of every line?
(Assuming separators can only be on their own line.)

That would probably involve Lisp code instead of C, so that could actually be slower.

Ambrevar:
Are you dropping this?
Sorry for the long radio-silence, but I was eventually thinking of giving it another shot at some point.  What do you think?

thierryvolpiatto:

> Are you dropping this?

Yes, this implementation is not the good approach anyway.

> Sorry for the long radio-silence, but I was eventually thinking of
> giving it another shot at some point. What do you think?

As you want, make another PR if needed, keep in mind what we need is
performance, previous implementation was too slow to be usable.

--
Thierry

PR #1898, opened 2017-10-15
Make alternate-split and mosaic respect horizontal split preference

Ambrevar:
Fix inaccuracies in function documentation.

Mosaic can now be rectangular which makes for less weird results with, say, 5 buffers.


thierryvolpiatto:
Thanks.

PR #1905, opened 2017-10-30
Add helm-eshell-prompts

Ambrevar:
New Helm source for Eshell: Browse all Eshell prompts.

This aligns well with the logic of helm-imenu, helm-org-in-buffer-headings, helm-semantic, etc.


thierryvolpiatto:
Thanks, merged.

PR #1908, opened 2017-11-02
Fixes to helm-sys.el

ahungry:
- Corrected typo for defcustom name colum to column
- Corrected incorrect logic for setting CPU sort function under GNU/Linux


thierryvolpiatto:
Thanks, merged.

PR #1953, opened 2018-01-19
Give new frame input focus

### LINKED ISSUE #1952: pop-up frame should be given input focus
`helm-display-buffer-pupup-frame` is great, but the old frame still has focus.  `select-frame-set-input-focus` should be used to give the new frame input focus.

hexmode:
fixes #1952


thierryvolpiatto:
Thanks merged.

PR #1977, opened 2018-03-03
eshell: Fix infinite loop on non-empty last prompt

Ambrevar:


thierryvolpiatto:
Can't reproduce the infloop (tried with last prompt running ping).


         (cl-loop while (not (eobp))
                  for promptno from 1
+                 for prev-point = (point)
                  do (eshell-next-prompt 1)
+                 until (= (point) prev-point)
                  collect (list (buffer-substring-no-properties
                                 (point) (line-end-position))
                                (point)


Any way, isn't this simpler ?

(cl-loop while (and (not (eobp)) (eshell-next-prompt 1)) 
                 for promptno from 1
                 collect (list (buffer-substring-no-properties
                                (point) (line-end-position))
                               (point)
                               (buffer-name)
                               (and helm-eshell-prompts-promptidx-p
                                    promptno)))

-- 
Thierry

Ambrevar:
I think you failed to reproduce because you've used Emacs <26.  Starting from version 26, the prompt detection has changed and `eshell-next-prompt` does not simply move to the next paragraph anymore, it tries to go to the next read-only line.

Consider the following Eshell buffer:

```
$ <|>echo
```

with `<|>` being the point.  On Emacs 26, `eshell-next-prompt` won't move the point.  As a consequence, `(eobp)` will always be false, hence the infinite loop.
This also applies to #1978.

thierryvolpiatto:

> I think you failed to reproduce because you've used Emacs
> <26. Starting from version 26, the prompt detection has changed and
> eshell-next-prompt does not simply move to the next paragraph anymore,
> it tries to go to the next read-only line.

I made some more changes, it is now working on 26+ as well.

> Consider the following Eshell buffer:
>
> $ <|>echo
>
> with <|> being the point. On Emacs 26, eshell-next-prompt won't move
> the point.

Well if user modify its eshell-prompt like this perhaps he have to
modify eshell-prompt-regexp as well to fit with it ?
Anyway we will use now another version of eshell-next-prompt that fits
our needs.

--
Thierry

Ambrevar:
> Well if user modify its eshell-prompt like this perhaps he have to
> modify eshell-prompt-regexp as well to fit with it ?

Not sure what you mean: I did not meant that the prompt is modified (`$` is just an example, the issue occurs with any prompt).

Your latest commits don't work for me: `eshell-skip-prompt` returns nil in `helm-eshell-next-prompt` and the prompt list is always empty.

thierryvolpiatto:

> Your latest commits don't work for me: eshell-skip-prompt returns nil
> in helm-eshell-next-prompt and the prompt list is always empty.

Work for me on both emacs-25 and 27, please start from emacs-helm.sh to
see if it works from there.

Thanks.

--
Thierry

Ambrevar:
After further investigation, I think I've found what causes this: you've
switched from eshell-next-prompt to eshell-skip-prompt, but I think
that's a mistake.  From the documentation:

> Skip past the text matching regexp ‘eshell-prompt-regexp’.
> If this takes us past the end of the current line, don’t skip at all.

Thus it won't work if the prompt is on multiple lines.

Another issue: if the banner is deactivated, the prompt then starts from the
first character, which means that the first property change is the end
of the prompt rather then the beginning.  The following
`eshell-skip-prompt` then returns `nil` and no prompt gets listed at all.

My suggestion: revert to using `eshell-next-prompt`, and on Emacs<26
inline the code of Emacs 26.

thierryvolpiatto:

> After further investigation, I think I've found what causes this: you've
> switched from eshell-next-prompt to eshell-skip-prompt, but I think
> that's a mistake. From the documentation:
>
>> Skip past the text matching regexp ‘eshell-prompt-regexp’.
>> If this takes us past the end of the current line, don’t skip at all.
>
> Thus it won't work if the prompt is on multiple lines.

Don't know what you mean by "prompt is on multiple lines".

> Another issue: if the banner is deactivated, the prompt then starts from the
> first character, which means that the first property change is the end
> of the prompt rather then the beginning. The following
> `eshell-skip-prompt` then returns `nil` and no prompt gets listed at all.

So let fix helm-eshell-next-prompt to handle this use case.

> My suggestion: revert to using `eshell-next-prompt`, and on Emacs<26
> inline the code of Emacs 26.

The code of emacs-26 is not working for me either on emacs-25 and 26+,
it is why I made helm-eshell-next-prompt which works for both here.

Can you tell me if helm-eshell-prompts/all works from emacs-helm.sh ?

If yes, then you will have just to handle the corner cases you mentioned above.

-- 
Thierry

Ambrevar:
> Don't know what you mean by "prompt is on multiple lines".

The prompt can be an arbitrary string and can include new lines.
Example:

```
(login@host)[path]
>
```

In the above example, the commandline starts after the `>`.

> The code of emacs-26 is not working for me either on emacs-25 and 26+,
> it is why I made helm-eshell-next-prompt which works for both here.

Not sure what does not work for you: `eshell-next-prompt` in Emacs 26+
properly moves the cursor to the next prompt for me.  The only issue
with Helm-Eshell is that it won't move the point once on the last
prompt.

> Can you tell me if helm-eshell-prompts/all works from emacs-helm.sh ?

No it does not.

Ambrevar:
For `helm-eshell-prompts-all`, I meant that it does not work once the banner is disabled and/or the prompt contains a newline, so it's not exactly vanilla.

thierryvolpiatto:

> For helm-eshell-prompts-all, I meant that it does not work once the
> banner is disabled and/or the prompt contains a newline, so it's not
> exactly vanilla.

Sorry but you lost me, I understand nothing, initially the bug was with
emacs-26, now you suggest to use the code of 26 which initially was not
working??? And I don't understand what you mean above.
I will revert now to a simple re-search-forward that works in most
conditions, if still not working for you feel free to submit a new PR
that works both for 25 and 26.

Thanks.

--
Thierry

Ambrevar:
Sorry if this was unclear.  It's because there are _different_ things not working here:

- On Emacs 25, `eshell-next-prompt` is badly implemented and only goes to the next paragraph.  It moves the point when on the last prompt and thus the prompt listing will always end.

- On Emacs 26, `eshell-next-prompt` works properly for browsing the prompts.  The issue was that when the last prompt's commandline is not empty, Helm-Eshell will loop indefinitely because `eshell-next-prompt` does not move the point on the last prompt.

- The changes you've propose fixed the previous two issues but created a new one: because you switched from `eshell-next-prompt` to `eshell-skip-prompt`, multiline-prompts are not supported.  There is also a but when Eshell starts without banner.  Therefore I think `eshell-skip-prompt` is not the right approach.

My initial PR works for me, if you still want to consider this approach: use `eshell-next-prompt` and check whether points moves to make sure we don't get stuck.
For Emacs 25-, we could backport `eshell-next-prompt` from Emacs 26 (I did not do that in my PR).

thierryvolpiatto:
Thanks for explanations.

> Sorry if this was unclear. It's because there are different things not working here:
>
>   * On Emacs 25, eshell-next-prompt is badly implemented and only goes to the next paragraph. It moves the point when on the last prompt and thus the prompt listing
>     will always end.
>    
>   * On Emacs 26, eshell-next-prompt works properly for browsing the prompts. The issue was that when the last prompt's commandline is not empty, Helm-Eshell will loop
>     indefinitely because eshell-next-prompt does not move the point on the last prompt.
>    
>   * The changes you've propose fixed the previous two issues but
>   created a new one: because you switched from eshell-next-prompt to
>   eshell-skip-prompt,

I didn't, eshell-next-prompt is using eshell-skip-prompt, I did the same
in helm-eshell-next-prompt, but I didn't use a single call to
eshell-skip-prompt, anyway I think this approach is wrong and uselessly
complicated.

>   
>     multiline-prompts are not supported. There is also a but when
>   Eshell starts without banner. Therefore I think eshell-skip-prompt
>   is not the right approach.

Agree, a simple re-search-forward is much better and working fine with
no banner/multiline prompt (following screenshots).

Perhaps it can be enforced with a check for read-only text property if
you fear matching eshell-prompt-regexp between two prompts (improbable
but well).

> My initial PR works for me, if you still want to consider this
> approach: use eshell-next-prompt and check whether points moves to
> make sure we don't get stuck.

No doubt this is working, but it is uselessly complicated as said above.

> 
> For Emacs 25-, we could backport eshell-next-prompt from Emacs 26 (I
> did not do that in my PR).

I don't like emacs-26 eshell-next-prompt, which is wrong IMO.

-- 
Thierry

thierryvolpiatto:
Here a recipe from emacs-helm.sh:

eval

```elisp
(setq eshell-banner-message ""
      eshell-prompt-function
      (lambda ()
        (concat (abbreviate-file-name (eshell/pwd))
                "\n"
                (if (= (user-uid) 0) " # " " $ "))))
```
M-x eshell
M-x helm-eshell-prompts

capture d ecran_2018-03-10_07-34-23

Ambrevar:
Seems to be working for me now.  Thanks!

PR #1993, opened 2018-03-19
Require `format-spec` in `helm-grep.el`

### LINKED ISSUE #1992: `helm-grep` should `(require 'format-spec)`
## Expected behavior

`helm-grep` makes use of `format-spec`, which is in the `format-spec` package. It should therefore `(require 'format-spec)`.

## Actual behavior (from `emacs-helm.sh` if possible, see note at the bottom)

`helm-grep` signals an error, as noted in bbatsov/helm-projectile#86.

```
Error running timer: (error "In Grep' source:helm-grep-collect-candidates' (void-function format-spec)")
```

xiongtx:
Fix #1992

### DIFF
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -17,6 +17,7 @@
 
 ;;; Code:
 (require 'cl-lib)
+(require 'format-spec)
 (require 'helm)
 (require 'helm-help)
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks.

PR #1995, opened 2018-03-20
Make whole prompt customizable

Ambrevar:
`pos' needs to be passed as parameter because its definition is non-trivial.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -756,8 +756,8 @@ This option have no effect with emacs versions lower than 26."
   :group 'helm
   :type 'boolean)
 
-(defcustom helm-default-cursor-face-function
-  #'helm--set-default-cursor-face
+(defcustom helm-default-prompt-display-function
+  #'helm--set-default-prompt-display
   "The function to use to set face of fake cursor in header-line."
   :group 'helm
   :type 'function)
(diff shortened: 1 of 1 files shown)


Ambrevar:
The custom prompt for evil-collection-helm:

```
(defun evil-collection-helm--set-prompt-display (pos)
  (let (beg state region-active m)
    (with-selected-window (minibuffer-window)
      (setq beg (save-excursion (vertical-motion 0 (helm-window)) (point))
            state evil-state
            region-active (region-active-p)
            m (mark t)))
    (when region-active
      (setq m (- m beg))
      ;; Increment pos to handle the space before prompt (i.e `pref').
      (put-text-property (1+ (min m pos)) (+ 2 (max m pos))
                         'face
                         (list :background (face-background 'region))
                         header-line-format))
    (put-text-property
     ;; Increment pos to handle the space before prompt (i.e `pref').
     (+ 1 pos) (+ 2 pos)
     'face
     (if (eq state 'insert)
         'underline
       ;; Don't just use 'cursor, this can hide the current character.
       (list :inverse-video t
             :foreground (face-background 'cursor)
             :background (face-background 'default)))
     header-line-format)))
```

thierryvolpiatto:
Thanks, merged.

PR #1998, opened 2018-03-21
Add recent UCS source to `helm-ucs`

xiongtx:
image

Small remaining issues (no barrier to merge):

- Recent UCS entries don't have `helm-ucs-face` applied (presumably b/c `candidate` doesn't have the face)
- `helm-source-ucs-recent` doesn't update dynamically when a persistent action that modifies `helm-ucs-recent` is taken, even though `volatile` is `t`.
  - Perhaps I'm misunderstanding the purpose of `volatile`? If so, how to achieve the desired dynamic update?

### DIFF
--- a/helm-font.el
+++ b/helm-font.el
@@ -20,10 +20,16 @@
 (require 'cl-lib)
 (require 'helm)
 (require 'helm-help)
+(require 'ring)
 
 (defgroup helm-font nil
   "Related applications to display fonts in helm."
   :group 'helm)
+
+(defcustom helm-ucs-recent-size 10
+  "Number of recent chars to keep."
+  :type 'integer
+  :group 'helm-font)
 
 (defvar helm-ucs-map
   (let ((map (make-sparse-keymap)))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Merged, thanks.

thierryvolpiatto:

> Small remaining issues (no barrier to merge):
>
>   * Recent UCS entries don't have helm-ucs-face applied (presumably
>   b/c candidate doesn't have the face)

Don't know for this, I presume the ring package remove text properties
of its items when adding to ring (I am not familiar with ring.el).

>
>   * helm-source-ucs-recent doesn't update dynamically within a single
>   Helm session when a persistent action that modifies helm-ucs-recent
>   is taken, even though volatile is t.  + Perhaps I'm misunderstanding
>   the purpose of volatile here?

Yes volatile doesn't update source, it just prevent caching candidates,
so it is indeed needed here to see your new candidates, otherwise
initial value is used until next session (or force-update). What you
need here is a persistent action that update candidates, something like
this:

    (lambda (candidate)
      (funcall (cdar helm-ucs-actions) candidate)
      (helm-update))

--
Thierry

xiongtx:
> I presume the `ring` package remove text properties of its items when adding to ring

Not so:

```emacs-lisp
(let ((r (make-ring 10)))
  (ring-insert r (propertize "foo" 'face font-lock-builtin-face))
  (ring-elements r))
;; => (#("foo" 0 3 (face font-lock-builtin-face)))
```

Using a `message` in `helm-ucs-insert-char`, I see that `candidate` is not fontified. So Helm itself is likely stripping out the text properties somewhere, which it probably shouldn't.

Issue #1832, opened 2017-08-07
Eshell completion goes wrong on `../../`


Ambrevar:
## Expected behavior

Path completion on `../../` (and more levels back) should work just like vanilla `eshell-pcomplete`.

## Actual behavior

- `ls ../../<TAB>` expands to `ls ./`.
- `ls ../../../<TAB>` expands to `ls ./` and so on.
- `ls ../../anything-completable-but-double-dots` does not yield anything.

## Steps to reproduce (recipe)

Add this to your init file:
```
(define-key eshell-mode-map [remap eshell-pcomplete] 'helm-esh-pcomplete)
```

Open Eshell and try out the aforementioned `ls...` commands.

## Describe versions of helm, emacs, operating system etc.

- helm: latest MELPA
- Emacs 25.2

thierryvolpiatto:
Pierre Neidhardt <notifications@github.com> writes:

> Expected behavior
>
> Path completion on ../../ (and more levels back) should work just like vanilla eshell-pcomplete.

Thanks, should work now.

-- 
Thierry
Gpg Key fingerprint = 6CEC 7081 AB33 E251 4AB8  5FC2 28D1 7F53 59F2 9997

Ambrevar:
It does fix the `../../` issue but it broke completion of anything but filenames: command, args, etc.

thierryvolpiatto:
Pierre Neidhardt <notifications@github.com> writes:

> It does fix the ../../ issue but it broke completion of anything but filenames: command, args, etc.

I made a quick fix, should work now.

Thanks.

-- 
Thierry
Gpg Key fingerprint = 6CEC 7081 AB33 E251 4AB8  5FC2 28D1 7F53 59F2 9997

Ambrevar:
Still a corner case: it does not complete subcommands, e.g. `git <TAB>` does not do anything.

Ambrevar:
Note that I'm using a custom completion function to generate `git`'s completion.

thierryvolpiatto:
Pierre Neidhardt <notifications@github.com> writes:

> Still a corner case: it does not complete subcommands, e.g. git <TAB>
> does not do anything.

Not only for git but for many other commands, some are implemented but
are not working with sudo etc...

See https://github.com/leoliu/pcmpl-git-el
and https://github.com/thierryvolpiatto/pcomplete-extension

--
Thierry
[...]

PR #2003, opened 2018-03-23
Add a slash instead of a space when point is behind a directory

solaryeti:
Pressing <tab> when point is behind a directory that does not contain
a / will result in a space being added. If the user wanted to continue
tabbing they would need to backspace, type a / and then they can
continue pressing / for completions.

This is different behaviour to what one might encounter in other
shells or in eshell with the default completion. This change
reintroduces that behaviour by adding a / instead of a space.

For example, taking _ to be point, pressing tab in the following situation:
`` cd /directory_ ``
will result in:
`` cd /directory<space>_``

This can interrupt a user's flow when they wanted to continue specifying either files or directories after the directory name. Continuing on the above example, this change results in the following behaviour:
``cd /directory/_``

The user is then free to continue pressing tab to bring up new completions. 

I see that there have been issues in the past with completions and slashes, especially related to . and .. (issue #1832). I've tested and it looks to me like this change doesn't introduce any regressions for that issue. I could not, however, find an automated to test this. If you have suggestions regarding that, or this change does break any other behaviour please let me know.

My elisp is not that strong so I'm open to any suggestions you have regarding this or if there is a better way to do this. Since this also changes behaviour users might be used to, or like, I'm also unsure if this change is better served with a variable that can enable/disable the new behaviour.

### DIFF
--- a/helm-eshell.el
+++ b/helm-eshell.el
@@ -299,7 +299,9 @@ The function that call this should set `helm-ec-target' to thing at point."
                        (and del-space (looking-back "\\s-" (1- (point)))
                             (delete-char -1))
                        (if (and (null helm-eshell--quit-flag)
-                                (looking-back "[.]\\{1,2\\}\\'" (1- (point))))
+                                (and (stringp last) (file-directory-p last))
+                                (looking-back "\\([.]\\{1,2\\}\\|[^/]\\)\\'"
+                                              (1- (point))))
                            (prog1 t (insert "/"))
                          ;; We need another flag for space here, but
                          ;; global to pass it to `helm-quit-hook', this
@@ -308,7 +310,10 @@ The function that call this should set `helm-ec-target' to thing at point."
                          ;; more completion, see issue #1832.
                          (unless (or helm-eshell--quit-flag
                                      (looking-back "/\\'" (1- (point))))
-                           (prog1 t (insert " ")))))
+                           (prog1 t (insert " ")))
+                         (when (and helm-eshell--quit-flag
+                                    (string-match-p "[.]\\{2\\}\\'" last))
+                           (insert "/"))))
                  (remove-hook 'helm-quit-hook 'helm-eshell--quit-hook-fn)
                  (setq helm-eshell--quit-flag nil)))))))
 


thierryvolpiatto:
Steven Meunier <notifications@github.com> writes:

> Pressing when point is behind a directory that does not contain
> a / will result in a space being added. If the user wanted to continue
> tabbing they would need to backspace, type a / and then they can
> continue pressing / for completions.
>
> This is different behaviour to what one might encounter in other
> shells or in eshell with the default completion. This change
> reintroduces that behaviour by adding a / instead of a space.
>
> For example, taking _ to be point, pressing tab in the following situation:
> cd /directory_
> will result in:
> cd /directory<space>_
>
> This can interrupt a user's flow when they wanted to continue
> specifying either files or directories after the directory
> name. Continuing on the above example, this change results in the
> following behaviour: cd /directory/_

This would be a good behavior if "/directory" where unique and you want
to complete to its subdirectories, however what if it is not unique and
you want to complete to a directory prefixed by "directory" e.g.:

cd /foo RET
ls RET
/foo1
/foo12
/foo123
/foo1234

cd /foo1 TAB

With the behavior you propose helm will add a "/" after "/foo1"
preparing to complete to the subdirectories and files of "/foo1" on next
TAB pressed, but what if you wanted to complete to "/foo123"?

--
Thierry

solaryeti:
I've tested your example and I get completions offered for foo1, foo12, etc., when tabbing after /foo1. Please correct me if my understanding is wrong, but because the function checks for possible completions before this condition is reached they are still presented to the user if available.

Just to be sure, are you getting a / after /foo1 when you test the change yourself? 

It is my intention that the slash should only be added if the user has typed the full directory name and there are no other possible completions.

thierryvolpiatto:
Steven Meunier <notifications@github.com> writes:

> I've tested your example and I get completions offered for foo1,
> foo12, etc., when tabbing after /foo1. Please correct me if my
> understanding is wrong, but because the function checks for possible
> completions before this condition is reached they are still presented
> to the user if available.

Yes you are right, I didn't look at the code.
Isn't something like this better (and simpler)?

diff --git a/helm-eshell.el b/helm-eshell.el
index 177f9b45..0f55d16a 100644
--- a/helm-eshell.el
+++ b/helm-eshell.el
@@ -299,7 +299,8 @@ The function that call this should set `helm-ec-target' to thing at point."
                        (and del-space (looking-back "\\s-" (1- (point)))
                             (delete-char -1))
                        (if (and (null helm-eshell--quit-flag)
-                                (looking-back "[.]\\{1,2\\}\\'" (1- (point))))
+                                (looking-back "\\([.]\\{1,2\\}\\|[^/]\\)\\'"
+                                              (1- (point))))
                            (prog1 t (insert "/"))
                          ;; We need another flag for space here, but
                          ;; global to pass it to `helm-quit-hook', this

--
Thierry

solaryeti:
Thanks for the suggestion! Yes, that is better and simpler.

I do find though that I still need to add (file-directory-p last) to the condition or else a / will be added to files as well:
```
diff --git a/helm-eshell.el b/helm-eshell.el
index 177f9b45..985eb591 100644
--- a/helm-eshell.el
+++ b/helm-eshell.el
@@ -299,7 +299,9 @@ The function that call this should set `helm-ec-target' to thing at point."
                        (and del-space (looking-back "\\s-" (1- (point)))
                             (delete-char -1))
                        (if (and (null helm-eshell--quit-flag)
-                                (looking-back "[.]\\{1,2\\}\\'" (1- (point))))
+                                (file-directory-p last)
+                                (looking-back "\\([.]\\{1,2\\}\\|[^/]\\)\\'"
+                                              (1- (point))))
                            (prog1 t (insert "/"))
                          ;; We need another flag for space here, but
                          ;; global to pass it to `helm-quit-hook', this
```

 I've tested a little bit and don't see that creating any other issues. Do you see any problem with that? If not, then I'll update the change with the above.

solaryeti:
While testing I've noticed that a slash does not get added if tabbing after '.' or '..'. For example:
  `` cd .. TAB ``
does nothing but gives the message "No completions of .."

If I understand the existing code correctly then this should get a slash, but it doesn't. I have found that removing (null helm-eshell--quit-flag) from the condition solves this without breaking the behaviour I'm trying to introduce. In other words, removing (null helm-eshell--quit-flag) still gives me the behaviour of adding / to the directory name and now also to adding it to '..' and '.' (provided there are no completions available for files starting with '.').

Nothing seems to break when I remove (null helm-eshell--quit-flag) but I guess I am just not testing widely enough. Do you know what behaviour removing this would break?

thierryvolpiatto:
Steven Meunier <notifications@github.com> writes:

> While testing I've noticed that a slash does not get added if tabbing
> after '.' or '..'. For example: cd .. TAB does nothing but gives the
> message "No completions of .."
>
> If I understand the existing code correctly then this should get a
> slash, but it doesn't. I have found that removing (null
> helm-eshell--quit-flag) from the condition solves this without
> breaking the behaviour I'm trying to introduce. In other words,
> removing (null helm-eshell--quit-flag) still gives me the behaviour of
> adding / to the directory name and now also to adding it to '..' and
> '.' (provided there are no completions available for files starting
> with '.').
>
> Nothing seems to break when I remove (null helm-eshell--quit-flag) but
> I guess I am just not testing widely enough. Do you know what
> behaviour removing this would break?

It would insert a "/" when inserting from completion.
Use something like this instead:

diff --git a/helm-eshell.el b/helm-eshell.el
index 177f9b45..53b2f4c8 100644
--- a/helm-eshell.el
+++ b/helm-eshell.el
@@ -299,7 +299,8 @@ The function that call this should set `helm-ec-target' to thing at point."
                        (and del-space (looking-back "\\s-" (1- (point)))
                             (delete-char -1))
                        (if (and (null helm-eshell--quit-flag)
-                                (looking-back "[.]\\{1,2\\}\\'" (1- (point))))
+                                (looking-back "\\([.]\\{1,2\\}\\|[^/]\\)\\'"
+                                              (1- (point))))
                            (prog1 t (insert "/"))
                          ;; We need another flag for space here, but
                          ;; global to pass it to `helm-quit-hook', this
@@ -308,7 +309,10 @@ The function that call this should set `helm-ec-target' to thing at point."
                          ;; more completion, see issue #1832.
                          (unless (or helm-eshell--quit-flag
                                      (looking-back "/\\'" (1- (point))))
-                           (prog1 t (insert " ")))))
+                           (prog1 t (insert " ")))
+                         (when (and helm-eshell--quit-flag
+                                    (string-match-p "[.]\\{2\\}\\'" last))
+                           (insert "/"))))
                  (remove-hook 'helm-quit-hook 'helm-eshell--quit-hook-fn)
                  (setq helm-eshell--quit-flag nil)))))))


--
Thierry

thierryvolpiatto:
Steven Meunier <notifications@github.com> writes:

> Thanks for the suggestion! Yes, that is better and simpler.
>
> I do find though that I still need to add (file-directory-p last) to
> the condition or else a / will be added to files as well:

I prefer avoiding file-directory-p here if possible, I can't reproduce
insertion of "/" after a file, how do you trigger this ? (perhaps it is
because you have removed `(null helm-eshell--quit-flag)`?)

-- 
Thierry

solaryeti:
> It would insert a "/" when inserting from completion.

Do you have any example of when this would happen? I'm sorry, I'm not trying to argue or second guess you here - it's just that I don't experience this and if you have a use case where I can reproduce that it also gives me an extra use case that I can test my change against to ensure that I don't break any behaviour.

> Use something like this instead:

Thanks for the suggestion. That fixes the problem with a "/" not being added to ".." nicely.

> I prefer avoiding file-directory-p here if possible, I can't reproduce insertion of "/" after a file, how do you trigger this ? (perhaps it is because you have removed `(null helm-eshell--quit-flag)`?)

No, I only removed `(null helm-eshell--quit-flag)` afterwards when I was looking at why '..' wasn't getting a slash when pressing tab. I have tested with and without file-directory-p and I always get the problem if I exclude it.

To be sure that I don't have anything that could be interfering with helm, I have tested using the emacs-helm.sh script. These are the only differences to the helm repo:
```
diff --git a/emacs-helm.sh b/emacs-helm.sh
index 1c4e22f0..37510f15 100755
--- a/emacs-helm.sh
+++ b/emacs-helm.sh
@@ -106,6 +106,7 @@ cat > $CONF_FILE <<EOF
   (define-key lisp-interaction-mode-map [remap completion-at-point] 'helm-lisp-completion-at-point)
   (define-key emacs-lisp-mode-map       [remap completion-at-point] 'helm-lisp-completion-at-point))
 (add-hook 'kill-emacs-hook #'(lambda () (and (file-exists-p "$CONF_FILE") (delete-file "$CONF_FILE"))))
+(add-hook 'eshell-mode-hook (lambda() (define-key eshell-mode-map [remap eshell-pcomplete] 'helm-esh-pcomplete)))
 EOF
 
 $EMACS -Q -l "$CONF_FILE" "$@"
diff --git a/helm-eshell.el b/helm-eshell.el
index 38c418a5..5eb89621 100644
--- a/helm-eshell.el
+++ b/helm-eshell.el
@@ -298,10 +298,10 @@ The function that call this should set `helm-ec-target' to thing at point."
                        ;; it when nothing found.
                        (and del-space (looking-back "\\s-" (1- (point)))
                             (delete-char -1))
                        (if (and (null helm-eshell--quit-flag)
-                                (looking-back "[.]\\{1,2\\}\\'" (1- (point))))
+                                (looking-back "\\([.]\\{1,2\\}\\|[^/]\\)\\'"
+                                              (1- (point))))
                            (prog1 t (insert "/"))
                          ;; We need another flag for space here, but
                          ;; global to pass it to `helm-quit-hook', this
@@ -310,7 +310,10 @@ The function that call this should set `helm-ec-target' to thing at point."
                          ;; more completion, see issue #1832.
                          (unless (or helm-eshell--quit-flag
                                      (looking-back "/\\'" (1- (point))))
-                           (prog1 t (insert " ")))))
+                           (prog1 t (insert " ")))
+                         (when (and helm-eshell--quit-flag
+                                    (string-match-p "[.]\\{2\\}\\'" last))
+                           (insert "/"))))
                  (remove-hook 'helm-quit-hook 'helm-eshell--quit-hook-fn)
                  (setq helm-eshell--quit-flag nil)))))))
```
I then do the following in eshell:
``ls emacs-helm.sh<TAB>``
which results in
``ls emacs-helm.sh/``

I am using Emacs 25.3.1.

thierryvolpiatto:
Steven Meunier <notifications@github.com> writes:

> I then do the following in eshell:
> ls emacs-helm.sh<TAB>
> which results in
> ls emacs-helm.sh/

Ok I see, thanks. So go ahead with the same patch above + your
file-directory-p test (added a test for last to not fail with
(file-directort-p nil)):

diff --git a/helm-eshell.el b/helm-eshell.el
index 177f9b45..66e81d75 100644
--- a/helm-eshell.el
+++ b/helm-eshell.el
@@ -299,7 +299,9 @@ The function that call this should set `helm-ec-target' to thing at point."
                        (and del-space (looking-back "\\s-" (1- (point)))
                             (delete-char -1))
                        (if (and (null helm-eshell--quit-flag)
-                                (looking-back "[.]\\{1,2\\}\\'" (1- (point))))
+                                (and (stringp last) (file-directory-p last))
+                                (looking-back "\\([.]\\{1,2\\}\\|[^/]\\)\\'"
+                                              (1- (point))))
                            (prog1 t (insert "/"))
                          ;; We need another flag for space here, but
                          ;; global to pass it to `helm-quit-hook', this
@@ -308,7 +310,10 @@ The function that call this should set `helm-ec-target' to thing at point."
                          ;; more completion, see issue #1832.
                          (unless (or helm-eshell--quit-flag
                                      (looking-back "/\\'" (1- (point))))
-                           (prog1 t (insert " ")))))
+                           (prog1 t (insert " ")))
+                         (when (and helm-eshell--quit-flag
+                                    (string-match-p "[.]\\{2\\}\\'" last))
+                           (insert "/"))))
                  (remove-hook 'helm-quit-hook 'helm-eshell--quit-hook-fn)
                  (setq helm-eshell--quit-flag nil)))))))
 

> I am using Emacs 25.3.1.
>
> —


-- 
Thierry

solaryeti:
I have updated the commit with your latest suggestions. Thank you so much for taking the time to review this and for the improvements you have made to what I originally suggested.

thierryvolpiatto:
Thanks, merged.

solaryeti:
Thank you, and thank you very much for all the effort you put into helm!

PR #2005, opened 2018-03-27
Fix broken helm-source-tracker

cpbotha:
These are changes to get helm-source-tracker working again on the
tracker version 1.6 which ships with Ubuntu 16.04. Tracker 1.6.2 was
released on March 1, 2016.

### DIFF
--- a/helm-x-files.el
+++ b/helm-x-files.el
@@ -98,16 +98,28 @@
                     (replace-match "" t t cand)) end
            collect built))
 
+(defun helm-source-tracker-1_6-transformer (candidates _source)
+  ;; loop through tracker candidates selecting out file:// lines
+  ;; then select part after file:// and url decode to get straight filenames
+  (cl-loop for cand in candidates
+           when (and (stringp cand)
+                     (string-match "\\`[[:space:]]*file://\\(.*\\)" cand))
+           collect (url-unhex-string(match-string 1 cand))))
+
 (defvar helm-source-tracker-search
   (helm-build-async-source "Tracker Search"
     :candidates-process
      (lambda ()
+       ;; the tracker-search command has been deprecated, now invoke via tracker
+       ;; also, disable the contextual snippets which we don't currently use
        (start-process "tracker-search-process" nil
-                      "tracker-search"
+                      "tracker" "search"
+                      "--disable-snippets"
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Charl P. Botha <notifications@github.com> writes:

> These are changes to get helm-source-tracker working again on the
> tracker version 1.6 which ships with Ubuntu 16.04. Tracker 1.6.2 was
> released on March 1, 2016.

If you think helm-source-tracker-1_6-transformer is enough
and is working well, please rename it helm-source-tracker-transformer
and remove the old function (have a look however if the extra features
in this old function are not needed).

Also perhaps we should provide a variable handling the command line to
use to allow users using an old version of tracker or anticipate further
changes in next versions ?

Thanks.

-- 
Thierry

cpbotha:
The old code has no comments documenting the intention of the author, and I don't have a Linux system old enough to see what exactly else it on that version of the tracker.

However, there was a previous PR where you also suggested adding `--remove-snippets` to NOT show the search context found inside the files: https://github.com/emacs-helm/helm/pull/533/files

However, the contributor removed your `--remove-snippets` and worked on the transformer function to process the two-lines-per-result (filename + context in the file contents).

Anyways, currently the complicated code does not work (and is probably broken on most current Linux systems), and my new much simpler code in this PR does.

I don't have time right now to continue with adding back the old "show contents of file with search results" solution, and lack of documentation by the previous author is complicating this unnecessarily.

Long story short: I'm happy to leave this PR here until more users complain about tracker, and/or I find the time to look at adding search context (i.e. snippets) back.

thierryvolpiatto:
Charl P. Botha <notifications@github.com> writes:

> Long story short: I'm happy to leave this PR here until more users
> complain about tracker, and/or I find the time to look at adding
> search context (i.e. snippets) back.

Hmm there is few chances that somebody look at this soon, so I will
merge your PR and replace old transformer with old, if somebody complain
that some features are missing we will consider adding back the features
of old transformer to new one, for now it looks good enough.

Thanks.

--
Thierry

thierryvolpiatto:
Merged thanks.

cpbotha:
Thanks, I also think that's the best call to make now. :)

PR #2011, opened 2018-04-01
Avoid excessively calling format-mode-line

ryjm:
During the initial buffer list setup, `format-mode-line` and `length`
were called in every buffer regardless of the values of
`helm-buffer-max-len-mode` and `helm-buffer-max-length`. Since the
performance of format-mode-line depends on the mode, this can really
slow down the initialization of the buffer list. For example,
`delight-mode`'s advice around `format-mode-line` was causing
an obscene number of calls to `dired-get-filename`.

format-mode-line         1044        9.8059119999  0.0093926360
helm-buffers-list--init  1           9.804546      9.804546
dired-get-filename       120960      0.0           0.0

Might as well only do this computation once.

### DIFF
--- a/helm-buffers.el
+++ b/helm-buffers.el
@@ -227,24 +227,27 @@ Note that this variable is buffer-local.")
   ;; reused in each source (issue #1907), now 'candidates attr is set
   ;; directly so that each list of candidates is local to source.
   (helm-attrset 'candidates (funcall (helm-attr 'buffer-list)))
-  (let ((result (cl-loop with allbufs = (memq 'helm-shadow-boring-buffers
-                                              (helm-attr
-                                               'filtered-candidate-transformer
-                                               helm-source-buffers-list))
-                         for b in (if allbufs
-                                      (helm-attr 'candidates)
-                                    (helm-skip-boring-buffers
-                                     (helm-attr 'candidates)
-                                     helm-source-buffers-list))
-                         maximize (length b) into len-buf
-                         maximize (length (with-current-buffer b
-                                            (format-mode-line mode-name)))
-                         into len-mode
-                         finally return (cons len-buf len-mode))))
-    (unless (default-value 'helm-buffer-max-length)
-      (helm-set-local-variable 'helm-buffer-max-length (car result)))
-    (unless (default-value 'helm-buffer-max-len-mode)
-      (helm-set-local-variable 'helm-buffer-max-len-mode (cdr result)))))
+  (cl-loop with allbufs = (memq 'helm-shadow-boring-buffers
+                                (helm-attr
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks for your contribution.

Jake Miller <notifications@github.com> writes:

> During the initial buffer list setup, format-mode-line and length
> were called in every buffer regardless of the values of
> helm-buffer-max-len-mode and helm-buffer-max-length.

You should not care of these. As a result with your patch modes are no
more aligned and preselection gets wrong (try C-]).

> Since the performance of format-mode-line depends on the mode, this
> can really slow down the initialization of the buffer list.

We don't need to call format-mode-line all the time here, it may be
needed in only in few cases where mode-name is a mode-line spec, so a
simple check on this suffice e.g.

@@ -246 +246,3 @@ Note that this variable is buffer-local.")
-                                            (format-mode-line mode-name)))
+                                            (if (stringp mode-name)
+                                                mode-name
+                                              (format-mode-line mode-name))))

In 99% of cases format-mode-line will not be called.
I will commit a change along this, let me know if it still slow for you.
Thanks.

> For example, delight-mode's advice around format-mode-line was causing
> an obscene number of calls to dired-get-filename.

Well we have no control on what people do with advices...

-- 
Thierry

ryjm:
> On Apr 1, 2018, at 2:28 AM, Thierry Volpiatto <notifications@github.com> wrote:
> 
> 
> Thanks for your contribution.
> 
> Jake Miller <notifications@github.com> writes:
> 
> > During the initial buffer list setup, format-mode-line and length
> > were called in every buffer regardless of the values of
> > helm-buffer-max-len-mode and helm-buffer-max-length.
> 
> You should not care of these. As a result with your patch modes are no
> more aligned and preselection gets wrong (try C-]).

Ah you’re right, oops.

> 
> > Since the performance of format-mode-line depends on the mode, this
> > can really slow down the initialization of the buffer list.
> 
> We don't need to call format-mode-line all the time here, it may be
> needed in only in few cases where mode-name is a mode-line spec, so a
> simple check on this suffice e.g.
> 
> @@ -246 +246,3 @@ Note that this variable is buffer-local.")
> - (format-mode-line mode-name)))
> + (if (stringp mode-name)
> + mode-name
> + (format-mode-line mode-name))))
> 
> In 99% of cases format-mode-line will not be called.
> I will commit a change along this, let me know if it still slow for you.
> Thanks.

Much better solution!

> > For example, delight-mode's advice around format-mode-line was causing
> > an obscene number of calls to dired-get-filename.
> 
> Well we have no control on what people do with advices…

Yeah, just mentioned it cause it made me notice how many times format-mode-line was getting called :)

PR #2032, opened 2018-05-05
helm-grep.el: Only add tramp prefix to fname when needed

Kypert:
When using helm-projectile-grep over tramp, the tramp prefix is added
twice to the candidate path. See: bbatsov/helm-projectile#76

In the function helm-grep-action, only add the tramp prefix to the fname if
it does not already include it.

### DIFF
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -604,7 +604,7 @@ WHERE can be one of other-window, other-frame."
          (tramp-host   (file-remote-p (or helm-ff-default-directory
                                           default-directory) 'host))
          (tramp-prefix (concat "/" tramp-method ":" tramp-host ":"))
-         (fname        (if tramp-host
+         (fname        (if (and tramp-host (not (file-remote-p loc-fname)))
                            (concat tramp-prefix loc-fname) loc-fname)))
     (cl-case where
       (other-window (helm-window-show-buffers


thierryvolpiatto:
M.Jonsson <notifications@github.com> writes:

> When using helm-projectile-grep over tramp, the tramp prefix is added
> twice to the candidate path. See: bbatsov/helm-projectile#76

If local filename is remote it is you have a problem.
Please report the bug as a helm bug with the tool and functions provided
in helm, not helm-projectile, helm-projectile is not part of helm.

Thanks.

-- 
Thierry

thierryvolpiatto:
I confirm this is not reproductible from helm-find-files, so I guess this is a helm-projectile bug.

Kypert:
Thank you for the quick response. I am a bit confused about what you mean. I realize that I am quite new to the way of workings in helm, for my understanding, could you please clarify:
1) You want me to create an issue in helm so that we can track this change? I am new to github, so I thought it would be enough to refer to the issue in helm-projectile.
2) You do not think that it is a problem in helm? I do not clearly see the connection to `helm-find-files`, but I think that I can reproduce it with `helm-grep-do-git-grep` when I am in a remote repository. As I see it, if loc-fname already comes with the tramp prefix (https://github.com/Kypert/helm/blob/f633966ee4640ba27fc0eaa794e67b0bfe2450a1/helm-grep.el#L596), there is no need to add it again?

thierryvolpiatto:
M.Jonsson <notifications@github.com> writes:

> Thank you for the quick response. I am a bit confused about what you mean. I realize that I am quite new to the way of workings in helm, for my understanding, could
> you please clarify:
>
>  1. You want me to create an issue in helm so that we can track this
>  change?

If you can reproduce the issue with helm only (recipe would be great, perhaps
use sudo method so that everybody can reproduce) report the issue in
helm to see what's wrong, but as I can't reproduce the issue here I
think it is more a helm-projectile issue which is not part of helm.

>  2. You do not think that it is a problem in helm?

Yes, but I may be wrong, let me know how you end up with a loc-fname
which is a remote name, I think you have an issue in helm-projectile.

>  I do not clearly see the connection to helm-find-files,

From helm-find-files start a grep session with C-s.

>  but I think that I can reproduce it with helm-grep-do-git-grep when I
>  am in a remote repository.

You think you can reproduce or you are actually able to reproduce?
Normally git should refuse to grep at all if you are root.

>  As I see it, if loc-fname already comes with the tramp prefix (
>  https://github.com/Kypert/helm/blob/f633966ee4640ba27fc0eaa794e67b0bfe2450a1/helm-grep.el#L596),
>  there is no need to add it again?

Right, but loc-fname should not be a remote name, never, if it is, here
is the problem and indeed needs to be fixed, but not from action IMHO.

-- 
Thierry

thierryvolpiatto:
I made a change to allow logging fnames. Please send here a helm debug log of a helm grep session or a helm git-grep session to see what's going on.
Use C-h C-d from helm to enable debugging.
- Use grep on a small set of files or small directory.
- Use C-! before entering pattern and C-! once done

Thanks.

Kypert:
I made some observations yesterday and reached out to the helm-projectile maintainer. I logged my findings in bbatsov/helm-projectile#76, tldr: it might come down to how the files are presented, fname is not taken from `(car-safe split)` but from ` (expand-file-name (car split) root)` where `(car split)` is a relative path in the helm-projectile case but an absolute path in the workings cases. 

Does that make any sense to you?

thierryvolpiatto:
M.Jonsson <notifications@github.com> writes:

> I made some observations yesterday and reached out to the
> helm-projectile maintainer. I logged my findings in
> bbatsov/helm-projectile#76, tldr: it might come down to how the files
> are presented, fname is not taken from (car-safe split) but from
> (expand-file-name (car split) root) where (car split) is a relative
> path in the helm-projectile case but an absolute path in the workings
> cases.
>
> Does that make any sense to you?

Yes, I guess you have a relative path because the backend is git, right?

Your change is not bad, I think we can merge it but I want to know why
you endup with with a remote path in helm buffer and fix it, 
I guess it is because we have (expand-file-name file_relative_path root) 
somewhere but want to be sure.

Thanks.

--
Thierry

thierryvolpiatto:
Perhaps this patch is enough:
```diff
diff --git a/helm-grep.el b/helm-grep.el
index 34d8bbf..a08f9c6 100644
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -1158,7 +1158,9 @@ in recurse, and ignore EXTS, search being made recursively on files matching
          (line   (if ansi-p (helm--ansi-color-apply candidate) candidate))
          (split  (helm-grep-split-line line))
          (fname  (if (and root split)
-                     (expand-file-name (car split) root)
+                     (expand-file-name (car split)
+                                       (or (file-remote-p root 'localname)
+                                           root))
                    (car-safe split)))
          (lineno (nth 1 split))
          (str    (nth 2 split))
```

thierryvolpiatto:
Have a try and let me know if it works for you.

Thanks.

Kypert:
> Yes, I guess you have a relative path because the backend is git, right?

Yes, I am using git as the backend. But the issue is still present if I am using normal grep in helm-projectile, what matters is how the [git] grep command is executed, if the result is including an absolute or relative path.

I share your view that it is because `(expand-file-name file_relative_path root)`, root (a tramp path) will be used, but in the normal case when an absolute path is expanded, root (default-directory) does not add the tramp path. But I do not know the C-implementation of `expand-file-name`.

> Have a try and let me know if it works for you.

Yes! Your patch works fine :) Your patch makes more sense than mine (maybe rename fname to loc-fname to reinforce the fact?). Will you merge it, or do you need me to adjust my commit?

thierryvolpiatto:
M.Jonsson <notifications@github.com> writes:

> Yes! Your patch works fine :) Your patch makes more sense than mine
> (maybe rename fname to loc-fname to reinforce the fact?).

I commented the code and merged.

> Will you merge it, or do you need me to adjust my commit?

I merged it, I think this change is enough, no need to merge your change
in action function.

Thanks for your time on this.

--
Thierry

Kypert:
Likewise, many thanks for your time and efforts!

ronievron:
The current code in helm-grep-action leaves our the tramp-user component.
The following works for me and avoids having to put together the various remote components.

 (tramp-fname (file-remote-p (or helm-ff-default-directory
                                          default-directory) nil))
         (fname        (if tramp-fname
                           (concat tramp-fname loc-fname) loc-fname)))

PR #2038, opened 2018-05-11
Make helm-esh-pcomplete persistent action iterative

Miciah:
With this change, `helm-esh-pcomplete`'s persistent action inserts the chosen candidate and then starts a new `helm-esh-pcomplete` session (unless there was only candidate to begin with).  This behavior is more similar to traditional shell completion and is especially convenient for using file completion to drill down a path.

* `helm-eshell.el` (`helm-esh-source`): Set `helm-eshell--repeat` to `t` in the persistent action.
(`helm-eshell--repeat`): New variable.
(`helm-esh-pcomplete`): Loop when persistent action sets `helm-eshell--repeat` to `t`.

---

In case it will make things clearer, here is an example.  I want to find a particular Python egg.  I know I have the egg installed under `~/.local/`; I type in `ls ~/.loc`:

```console
λ ls ~/.loc
```

...and press Tab:

```console
λ ls ~/.local/
```

There was only one candidate, so `helm-esh-pcomplete` inserts it and exits.

Now I press `Tab` again.  Under `~/.local/`, there are three candidates: `~/.local/bin/`, `~/.local/lib/`, and `~/.local/share/`.  I select `~/.local/lib/` and invoke the persistent action.  Here is where the new behavior manifests: The persistent action inserts the chosen candidate and then immediately starts a new `helm-esh-pcomplete` session.  Under `~/.local/lib/`, I have `~/.local/lib/python2.7/` and `~/.local/lib/python3.4/`, so Helm shows these two options, and the prompt line reflects that the selection is initially on the first candidate:

```console
λ ls ~/.local/lib/python2.7/
```

I invoke the persistent action again, which inserts the chosen candidate and starts a new `helm-esh-pcomplete` session.  Under `~/.local/lib/python2.7/`, there is only one candidate: `~/.local/lib/python2.7/site-packages/`.  However, since this `helm-esh-pcomplete` session was started using the persistent action, the session displays this candidate rather than immediately inserting it; this is the behavior I want: as long as I am drilling down the hierarchy, I don't want to be dropped out of the Helm session.

```console
λ ls ~/.local/lib/python2.7/site-packages/
```

Since I want to drill down further, I invoke the persistent action again.  Now I am presented with a long list of files.  I type some input to filter candidates for the desired file; if I want to drill down further into some directory, I can invoke the persistent action again; if I just want to select a candidate, I can press Enter; or I can press Escape to end the Helm session, which would leave me with `ls ~/.local/lib/python2.7/site-packages/`.

This behavior is very convenient and feels natural for files.  It works well enough in other contexts too; for example, I can type `chown `, press Tab, and invoke the persistent action upon a candidate whereupon `helm-esh-pcomplete` will immediately offer to complete a path name.  Another example: Type in `mount -`, press Tab, select `-t`, and invoke the persistent action; this inserts `mount -t ` and immediate offers filesystem types for completion.  (I hope these examples are somewhat compelling; the paucity of completions included in Eshell makes it difficult to come up with any wicked cool examples.)

---

I hope this behavior makes sense and that you will be willing to accept this change.  In any case, many thanks for Helm!

### DIFF
--- a/helm-eshell.el
+++ b/helm-eshell.el
@@ -80,7 +80,9 @@
                      (remove-hook 'minibuffer-setup-hook 'eshell-mode)))
    (candidates :initform 'helm-esh-get-candidates)
    ;(nomark :initform t)
-   (persistent-action :initform 'ignore)
+   (persistent-action :initform (lambda (_candidate)
+                                  (setq helm-eshell--repeat t)
+                                  (helm-exit-and-execute-action 'helm-ec-insert)))
    (nohighlight :initform t)
    (filtered-candidate-transformer
     :initform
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Miciah Dashiel Butler Masters <notifications@github.com> writes:

> With this change, helm-esh-pcomplete's persistent action inserts the
> chosen candidate and then starts a new helm-esh-pcomplete session
> (unless there was only candidate to begin with). This behavior is more
> similar to traditional shell completion and is especially convenient
> for using file completion to drill down a path.
>
>   * helm-eshell.el (helm-esh-source): Set helm-eshell--repeat to t in the persistent action.
>     (helm-eshell--repeat): New variable.
>     (helm-esh-pcomplete): Loop when persistent action sets helm-eshell--repeat to t.


Thanks for your time and efforts to make this PR, but I prefer keeping a
traditional completion in eshell for now i.e. hitting TAB when needed.
For file completion you can use helm-find-files:

    $ ls C-x C-f => navigate to file => C-c i


>
> In case it will make things clearer, here is an example. I want to
> find a particular Python egg. I know I have the egg installed under
> ~/.local/; I type in ls ~/.loc:

> λ ls ~/.local/lib/python2.7/site-packages/

You can do instead:

    $ ls C-x C-f /home/you/..site-packages/
 
to have completion on all site-packages under "~/". 
Hit RET on one and then C-c i on the file or
directory you want to ls (you can mark files if you want to ls more than
one).

Thanks.

--
Thierry

Issue #2058, opened 2018-06-14
bug deleting macro from helm-execute-helm


tttuuu888:
- Make 3 macros like below:
  `<f3> 1 <f4> <f3> 2 <f4> <f3> 3 <f4>`
- M-x helm-execute-helm and select macro `2` and `<f3>` to delete it.
- Not only macro `2` but also  `3` has gone.

I tested this in GNU Emacs 26.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.18.9) of 2018-05-29. (Starting Emacs with `emacs-helm.sh` script.)

PR #2059, opened 2018-06-14
Fix delete macro from helm-execute-kmacro.

tttuuu888:
This is fox for issue #2058.
`(kmacro-push-ring)` was missed so we were losing macro on top of the list when we delete the other macro from helm-execute-kmacro. 

I tested this on Emacs 26.1 and  Emacs 25.3.

Thanks a lot.


thierryvolpiatto:
Thanks!

Issue #2069, opened 2018-07-27
`helm-list-emacs-process` leaves tabulated list header in other buffer


jwinder:
## Expected behavior

`helm-list-emacs-process` should not affect original buffer.

## Actual behavior (from `emacs-helm.sh` if possible, see note at the bottom)

After finishing with `helm-list-emacs-process`, and extra border or header is left at the top of the original buffer.

## Steps to reproduce (recipe)

From `emacs-helm.sh`, run `helm-list-emacs-process`. Then finish an action on a process or cancel out of the mini-buffer. An extra border or header is left at the top of the original buffer.

I've found that this behavior is due to `tabulated-list-init-header` which is called by `list-processes--refresh` in the init of `helm-list-emacs-process`. Therefore, calling `tabulated-list-init-header` results in the same issue.

I've found a temporary fix for this is to move the init function (not sure if there's a better solution):

```elisp
(setq helm-source-emacs-process
  (helm-build-sync-source "Emacs Process"
    :candidates (lambda ()
                  (list-processes--refresh)
                  (mapcar #'process-name (process-list)))
    :persistent-action (lambda (elm)
                         (delete-process (get-process elm))
                         (helm-delete-current-selection))
    :persistent-help "Kill Process"
    :action (helm-make-actions "Kill Process"
                               (lambda (_elm)
                                 (cl-loop for p in (helm-marked-candidates)
                                          do (delete-process (get-process p)))))))
```

## Describe versions of Helm, Emacs, operating system, etc.

macOS 10.13.6. Not sure if this behavior happens on linux.
Emacs 26.1
Both helm 2.9.8 and master

## Are you using `emacs-helm.sh` to reproduce this bug (yes/no):

yes

## IMPORTANT NOTE

Helm provides a script named `emacs-helm.sh` which runs Helm in a neutral
environment: no other packages and only minimal settings.

When possible, use it to reproduce your Helm issue to ensure no other package is
interfering.

To run it, simply switch to the directory where Helm is installed and call `./emacs-helm.sh`.
If necessary you can specify emacs executable path on command line with "-P" option.

Thanks.

thierryvolpiatto:
Joe Winder <notifications@github.com> writes:

> ## Expected behavior
>
> `helm-list-emacs-process` should not affect original buffer.
>
> ## Actual behavior (from `emacs-helm.sh` if possible, see note at the bottom)
>
[...]

PR #2071, opened 2018-07-28
helm-list-emacs-process: set local header line to nil in original buffer

jwinder:
fixes https://github.com/emacs-helm/helm/issues/2069


thierryvolpiatto:
Thanks, merged.

PR #2077, opened 2018-08-08
Fix helm-org--get-candidates-in-file for the when it is given a buffe…

alastairdb:
…r instead of a filename

`helm-documentation` builds a buffer in org-mode that is not associated with a file. Currently, when `helm-org-show-filename` is non-nil, this generates an error. The pull request fixes that.


thierryvolpiatto:
Thanks, merged.

PR #2084, opened 2018-08-25
Add face to customize appearance of prompt

bennofs:
This is very useful for themes that use the helm header line and have a different background color for the minibuffer/header line.


thierryvolpiatto:
Thanks.

PR #2086, opened 2018-08-31
fix bug where adaptive history is invalid

CeleritasCelery:
The variables `print-length` and `print-level` control the amount of a sexp that is printed before it is truncated. If `nil`, these variables will let sexp be printed of any length. However if one of them is set to a value and helm-adaptive history is saved, then there is a chance that the output file will have a truncated sexp with invalid `...` symbols. This will cause helm to error when it tries to read `helm-adaptive-history-file`. So we bind these variables to nil before calling `prin1` so that the output will never be truncated.


alphapapa:
This is probably a good idea!

thierryvolpiatto:
Thanks, merged!

PR #2113, opened 2018-11-14
WIP: Set inhibit-quit when creating a process

corngood:
WIP only because I'm not very familiar with helm, so I'm looking for feedback.

This prevents processes from being leaked by ensuring they are added to
helm-async-processes before inhibit-quit is cleared.

Repro steps:
- install cygwin emacs-w32 (tested with 26.1)
- install `helm-ag` and `helm`
- `emacs -q --eval "(progn (package-initialize)(helm-do-ag))"`
- on the 'search in files' prompt select a directory where ag will take a long time to complete
- on the 'pattern' prompt, hold down the 'a' key to continually change the pattern
- C-g
- C-x C-c

Depending on timing, emacs should prompt you to kill a whole bunch of leaked child processes.

This seems to happen because the candidates proc is being called inside `helm-while-no-input`, and emacs is quitting with input events between when the process is created, and when it's added to the async process list.  The way input is signalled is a bit different on win32, so it might not happen on other platforms.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -3484,9 +3484,9 @@ WARNING: Do not use this mode yourself, it is internal to helm."
 
 (defun helm-get-candidates (source)
   "Retrieve and return the list of candidates from SOURCE."
-  (let* (inhibit-quit
-         (candidate-fn (assoc-default 'candidates source))
+  (let* ((candidate-fn (assoc-default 'candidates source))
          (candidate-proc (assoc-default 'candidates-process source))
+         (inhibit-quit candidate-proc)
          cfn-error
          (notify-error
           (lambda (&optional e)
@@ -3547,7 +3547,8 @@ WARNING: Do not use this mode yourself, it is internal to helm."
   "Return the cached value of candidates for SOURCE.
 Cache the candidates if there is no cached value yet."
   (let* ((name (assoc-default 'name source))
-         (candidate-cache (gethash name helm-candidate-cache)))
+         (candidate-cache (gethash name helm-candidate-cache))
+         (inhibit-quit (assoc-default 'candidates-process source)))
     (helm-aif candidate-cache
         (prog1 it (helm-log "Use cached candidates"))
       (helm-log "No cached candidates, calculate candidates")


thierryvolpiatto:
David McFarland <notifications@github.com> writes:

> WIP only because I'm not very familiar with helm, so I'm looking for feedback.
>
> This prevents processes from being leaked by ensuring they are added to
> helm-async-processes before inhibit-quit is cleared.

inhibit-quit is evil an should not be used, especially here.

> Repro steps:
> - install cygwin emacs-w32 (tested with 26.1)
> - install `helm-ag` and `helm`

Note that helm-ag is not part of Helm.

> 
> - `emacs -q --eval "(progn (package-initialize)(helm-do-ag))"` - on
> the 'search in files' prompt select a directory where ag will take a
> long time to complete - on the 'pattern' prompt, hold down the 'a' key
> to continually change the pattern - C-g - C-x C-c

I have no access to a Windows box, but on my Linux box all work as
expected.

> Depending on timing, emacs should prompt you to kill a whole bunch of
> leaked child processes.

Why? Helm is killing its processes when quitting or exiting and it is
the way to do IMO, did I miss something?

> This seems to happen because the candidates proc is being called
> inside `helm-while-no-input`, and emacs is quitting with input events
> between when the process is created, and when it's added to the async
> process list.  The way input is signalled is a bit different on win32,
> so it might not happen on other platforms.

It is not clear to me what is the problem you are encountering, can you
explain what is exactly happening?

Thanks.

--
Thierry

corngood:
> Note that helm-ag is not part of Helm.

Yeah, I understand, but all it's doing it providing 'candidates-process, which returns a new process.  I don't think it's doing anything wrong.

> It is not clear to me what is the problem you are encountering, can you explain what is exactly happening?

1. we're inside `helm-while-no-input`, so any input signals will terminate execution immediately
2. helm-get-cached-candidates calls helm-get-cached-candidates, which calls 'candidates-process from the source
3. a process is created by the source and returned
4. helm-get-cached candidates gets the process, and adds it to the async process list
```
      (let ((candidates (helm-get-candidates source)))
        (cond ((processp candidates)
               (push (cons candidates
                           (append source
                                   (list (cons 'item-count 0)
                                         (cons 'incomplete-line ""))))
                     helm-async-processes)
```

Execution can terminate at any time within `helm-while-no-input` when something sets quit-flag to 'input (and then maybe_quit is called), as long as quit is not inhibited, for example:

https://github.com/emacs-mirror/emacs/blob/master/src/w32fns.c#L3494

If that quit happens between (3) and (4), the process has still been created, but isn't stored in the async process list, so nothing ever deletes it.  In my experience, at least on win32, input can be signalled at any time.

My reasoning for doing the change this way is that it's important that execution from (3) to the end of (4) is atomic, but only when a process is created ('candidates-process).  If we're just executing elisp code ('candidates) we can probably terminate without leaking anything.

thierryvolpiatto:
I don't understand what is motivating this change, please describe what's annoy you at the user level.
I don't really understand what you mean by "leaking anything".
Thanks.

corngood:
It's basically what I described in the repro steps above.  I bring up helm-ag to search my project, and I start typing a pattern.  Each key-press should start a new child process, and kill the previous one, but because of this bug, some of them will be 'leaked'.  What I mean by leaked is that helm is not aware that they are running (they aren't in the async process list), so it will never kill them.  They will still show up in the emacs process list.

So if I bring up helm-ag and type in 'helm-while-no-input', depending on how quickly I type, I might leak 5 ag processes before I finish typing.  If not killed, ag will easily max out my CPU and/or disk IO for 30 seconds in a large project.  So now I have 6 ag processes competing for CPU/IO, and only one of them is running the final query that I care about.

helm-ag is where I usually see this, because it's by far the most common thing I do that uses 'candidates-process, but I was also able to reproduce it with `helm-do-grep-ag` and `helm-find` which are both in helm itself.

Here's an example of `(list-processes)` after I run `helm-do-grep-ag`, type 'a' a bunch of times and then close it with C-g:

https://clbin.com/bgnBg

thierryvolpiatto:
David McFarland <notifications@github.com> writes:

> It's basically what I described in the repro steps above.  I bring up
> helm-ag to search my project, and I start typing a pattern.  Each
> key-press should start a new child process, and kill the previous one,
> but because of this bug, some of them will be 'leaked'.  What I mean
> by leaked is that helm is not aware that they are running (they aren't
> in the async process list), so it will never kill them.  They will
> still show up in the emacs process list.
>
> So if I bring up helm-ag and type in 'helm-while-no-input', depending
> on how quickly I type, I might leak 5 ag processes before I finish
> typing.  If not killed, ag will easily max out my CPU and/or disk IO
> for 30 seconds in a large project.  So now I have 6 ag processes
> competing for CPU/IO, and only one of them is running the final query
> that I care about.

Ok, I understand now, that's bad indeed, but I suspect more an emacs bug
around while-no-input than a helm bug.
What emacs version are you using?

> helm-ag is where I usually see this, because it's by far the most
> common thing I do that uses 'candidates-process, but I was also able
> to reproduce it with `helm-do-grep-ag` and `helm-find` which are both
> in helm itself.
>
> Here's an example of `(list-processes)` after I run `helm-do-grep-ag`,
> type 'a' a bunch of times and then close it with C-g:
>
> https://clbin.com/bgnBg

Indeed, that's bad, I see also your default shell is zsh, not sure if
that is involved.
Here I can't reproduce with your recipe, all processes are killed, I use
emacs-26.


--
Thierry

corngood:
> Indeed, that's bad, I see also your default shell is zsh, not sure if that is involved. Here I can't reproduce with your recipe, all processes are killed, I use emacs-26.

I've reproduced it on a clean cygwin install where it's using bash, so I don't think that's it.

It's possible that it's limited to win32 (mingw/cygwin), because of how it handles input:

```c
static void
signal_user_input (void)
{
  /* Interrupt any lisp that wants to be interrupted by input.  */
  if (!NILP (Vthrow_on_input))
    {
      fprintf(stderr, "%u: %s %i\n", (unsigned)getpid(), __FUNCTION__, __LINE__);
      Vquit_flag = Vthrow_on_input;
      /* Calling maybe_quit from this thread is a bad idea, since this
	 unwinds the stack of the Lisp thread, and the Windows runtime
	 rightfully barfs.  */
    }
}
```

The comment implies that it can be called on another thread, so you'd have to assume it could happen at any time.  That being said, I see nothing in the docs that indicates this is incorrect behaviour.  It's possible that it happens on other platforms, with certain GUI configurations, or will happen in future versions of emacs.

> Ok, I understand now, that's bad indeed, but I suspect more an emacs bug around while-no-input than a helm bug. What emacs version are you using?

I'm using 26.1.  

I'm not sure why you think it's a bug.  From https://www.gnu.org/software/emacs/manual/html_node/elisp/Event-Input-Misc.html

> This construct runs the body forms and returns the value of the last one—but only if no input arrives. If any input arrives during the execution of the body forms, it aborts them (working much like a quit). The while-no-input form returns nil if aborted by a real quit, and returns t if aborted by arrival of other input.

This is exactly what's happening if input arrives while helm is getting candidates.

> If a part of body binds inhibit-quit to non-nil, arrival of input during those parts won't cause an abort until the end of that part. 

This is what I'm suggesting we do to protect the sensitive part between (3) and (4).  It seems to be exactly how they are suggesting `inhibit-quit` be used.

corngood:
Thanks.  Just tag me if you suspect anything has gone wrong.

thierryvolpiatto:
Ok, thanks for you PR.

PR #2119, opened 2018-12-21
Follow redirects since google returns 301 Moved

dvzubarev:
I was trying helm-google-suggest with curl enabled.
And came across an error:
```(error \"XML: (Not Well-Formed) Invalid end tag ‘HEAD’ (expecting ‘meta’)\")")```.
It turned out that Google returned 301 status.
```
#<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
#<TITLE>301 Moved</TITLE></HEAD><BODY>
#<H1>301 Moved</H1>
#The document has moved
#<A HREF="http://www.google.com/">here</A>.
#</BODY></HTML>
```


thierryvolpiatto:
Thanks! merged.

Issue #2120, opened 2018-12-24
Helm + neotree on spacemacs: unable to use Tab to switch to selected directory


phoe:
## Expected behavior

When using neotree in spacemacs, I used to be able to use Tab to navigate to folders inside helm. Now that functionality is broken.

## Actual behavior (from `emacs-helm.sh` if possible, see note at the bottom)

Error: `split-window: Cannot split side window or parent of side window`

## Steps to reproduce (recipe)

Download current spacemacs
`SPC f t` to open neotree
Select a directory
Hit `Tab`

## Backtraces if any (`M-x toggle-debug-on-error`)

I cannot copypaste the contents of `*Backtrace*` buffer because I cannot select text from it while the lower window is open, and once I close that window, the contents of that window disappear. The best I can give is a screenshot of my Emacs showing the backtrace.

![](https://user-images.githubusercontent.com/15045546/50385771-35b26900-06db-11e9-9626-cca94eafcdab.png)

## Describe versions of Helm, Emacs, operating system, etc.

Spacemacs 0.200 stable

## Are you using `emacs-helm.sh` to reproduce this bug (yes/no):

no

## Other things

Discussed on the spacemacs issue tracker: https://github.com/syl20bnr/spacemacs/issues/11368

thierryvolpiatto:
Please report bugs starting from emacs-helm.sh, and not spacemacs.
Thanks.

phoe:
I have run emacs with `emacs-helm.sh` but neotree is now not loaded. 
The package is downloaded at `~/.emacs.d/elpa/neotree-20180616.1603`
What steps are required to load neotree in this case?

phoe:
I just noticed that the Tab button works just fine unless neotree is open. After I open and then close it, tab button works normally.

thierryvolpiatto:
Michał Herda <notifications@github.com> writes:

> I have run emacs with emacs-helm.sh but neotree is now not loaded.
> The package is downloaded at ~/.emacs.d/elpa/neotree-20180616.1603
> What steps are required to load neotree in this case?

(load-file "/path/to/neotree.el")
M-x neotree

> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub, or mute the thread.*


-- 
Thierry

sdwolfz:
[...]

PR #2122, opened 2018-12-24
Fix window selection with multiple side windows

sdwolfz:
While trying to debug the follwing issues:
- https://github.com/emacs-helm/helm/issues/2120
- https://github.com/syl20bnr/spacemacs/issues/11368

Having:
```
(helm-default-display-buffer-functions '(display-buffer-in-side-window))
```
I ended up in the `helm-persistent-action-display-window` function.

When no other side windows exist here is the list of windows:
```
(window-list): (#<window 8 on *helm find files*> #<window 2 on  *Minibuf-1*> #<window 1 on *Messages*>)
(selected-window): #<window 8 on *helm find files*>
(next-window (selected-window) 1): #<window 1 on *Messages*>
(previous-window (selected-window) 1): #<window 1 on *Messages*>
```

When a side window already exists, for example treemacs:
```
(window-list): (#<window 12 on *helm find files*> #<window 2 on  *Minibuf-1*> #<window 10 on  *Treemacs-Framebuffer-1*> #<window 1 on *Messages*>)
(selected-window): #<window 12 on *helm find files*>
(next-window (selected-window) 1): #<window 10 on  *Treemacs-Framebuffer-1*>
(previous-window (selected-window) 1): #<window 1 on *Messages*>
```

Notice the treemacs window is right after the selected window, not
including the minibuffer, and the window we actually want to check if
it's dedicated or not (window 1) is right at the end.

Using the `previous-window` instead of `next-window` ensures that we
indeed get to check the window we care about (window 1). In the first
place it happend to be the case that both functions returned the same
(correct) value. Also `previous-window` makes sense since `window-list`
behaves like a ring and switching from a normal window to a helm window
means the normal window gets "left behind" in the previous position
while the "next window" would be whatever window was ever touched last.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -6209,7 +6209,7 @@ window to maintain visibility."
   "Return the window that will be used for persistent action.
 If SPLIT-ONEWINDOW is non-`nil' window is split in persistent action."
   (with-helm-window
-    (let (next-win cur-win)
+    (let (prev-win cur-win)
       (setq helm-persistent-action-display-window
             (cond ((and (window-live-p helm-persistent-action-display-window)
                         (not (member helm-persistent-action-display-window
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks!

PR #2124, opened 2019-01-05
Fix broken links in helm-help.el

siraben:
I found them broken links by running `(helm-documentation)` and attempting to convert it to an Org file so that I could export it.

**Important**: there seems to be a link that does not point anywhere, namely `Helm functions vs helmized emacs functions`, where is this link supposed to point to?


siraben:
Please do not merge yet until the link at line 966 is resolved

thierryvolpiatto:
Thanks merged.

PR #2126, opened 2019-01-10
Make helm-mark-all-1 return candidates with their properties

montag451:
`helm-mark-all-1` returns the candidates without their properties even if requested by the source. This pull request fix this issue


thierryvolpiatto:
Thanks!

PR #2127, opened 2019-01-21
Remove trailing whitespace(s)

duianto:
And one case of a whitespace before a closing double quote, the double quotes
are removed when helm-autoloads.el is generated.


thierryvolpiatto:
Thanks!

PR #2129, opened 2019-02-04
helm-shell: Init M-x shell prompt navigation

Ambrevar:
This is mostly a copy-paste from helm-eshell, with subtle differences.


thierryvolpiatto:
Thanks!

PR #2132, opened 2019-02-14
helm-misc: Use multiline source for comint input ring

Ambrevar:


helm-misc.el
```diff
@@ -20,6 +20,7 @@
 (require 'helm)
 (require 'helm-help)
 (require 'helm-types)
+(require 'helm-ring)
```
thierryvolpiatto: I guess you require helm-ring for `helm-kill-ring-max-offset`, I would prefer a new var, something like `helm-comint-max-offset`, this would allow separate values for both (kill-ring and comint) and also avoid requiring helm-ring just for that.
Thanks.

Ambrevar:
OK, see #2135.

PR #2158, opened 2019-06-06
Avoid stringp error

jagrg:
`(helm-basedir "foo")` returns nil so `(string-match "Trash/files/?\\'" nil)` won't work.


jagrg:
Never mind. There's no need for step 2 (C-k).

PR #2164, opened 2019-06-26
Refactor helm-org.el

jagrg:
I've made several changes to helm-org.el. Let me know if you have questions.

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -134,6 +134,17 @@ Also return their position in the buffer as marker objects."
 	      candidates))
       (nreverse candidates))))
 
+(defun helm-org-get-candidates (filenames)
+  "Return a list of candidates in FILENAMES."
+  (apply #'append
+         (mapcar (lambda (filename)
+		   (with-current-buffer
+		       (pcase filename
+                         ((pred bufferp) filename)
+                         ((pred stringp) (find-file-noselect filename t)))
+		     (helm-org--get-candidates-in-file)))
+                 filenames)))
+
 (defun helm-org-in-buffer-preselect ()
   "Preselect the heading at point."
   (if (org-at-heading-p)
@@ -198,16 +209,6 @@ Also return their position in the buffer as marker objects."
     :filtered-candidate-transformer 'helm-org-startup-visibility
     :parents parents
     :candidates filenames))
-(defun helm-org-get-candidates (filenames &optional parents)
-  (apply #'append
-         (mapcar (lambda (filename)
-                   (helm-org--get-candidates-in-file
-                    filename
-                    helm-org-headings-fontify
-                    (or parents (null helm-org-show-filename))
-                    parents))
-                 filenames)))
-
 (defun helm-org--get-candidates-in-file (filename &optional fontify nofname parents)
   (with-current-buffer (pcase filename
                          ((pred bufferp) filename)
--- a/helm-org.el
+++ b/helm-org.el
@@ -44,16 +44,6 @@ NOTE: This will be slow on large buffers."
 NOTE: This has no effect in `helm-org-in-buffer-headings'."
   :type 'boolean)
 
-(defcustom helm-org-headings-min-depth 1
-  "Minimum depth of org headings to start with."
-  :group 'helm-org
-  :type 'integer)
-
-(defcustom helm-org-headings-max-depth 8
-  "Go down to this maximum depth of org headings."
-  :group 'helm-org
-  :type 'integer)
-
 (defcustom helm-org-headings-actions
   '(("Go to heading" . helm-org-goto-marker)
     ("Open in indirect buffer `C-c i'" . helm-org--open-heading-in-indirect-buffer)
--- a/helm-org.el
+++ b/helm-org.el
@@ -167,6 +167,14 @@ when called from the `helm-org-agenda-files-headings' command."
       (outline-previous-visible-heading 1)
       (buffer-substring-no-properties (point-at-bol) (point-at-eol)))))
 
+(defun helm-source-org-headings-for-files (filenames &optional parents)
+  (helm-make-source "Org Headings" 'helm-org-headings-class
+    :candidates (lambda ()
+                  (if parents
+                      (with-helm-current-buffer
+                        (helm-org-get-parent-headings))
+                    (helm-org-get-candidates filenames)))))
+
 (defun helm-org-goto-marker (marker)
   "Go to MARKER showing the entry's context, body and subheadings."
   (switch-to-buffer (marker-buffer marker))
@@ -217,12 +225,6 @@ when called from the `helm-org-agenda-files-headings' command."
           (lambda (candidates)
             (let ((cands (helm-org-get-candidates candidates parents)))
               (if parents (nreverse cands) cands))))))
-
-(defun helm-source-org-headings-for-files (filenames &optional parents)
-  (helm-make-source "Org Headings" 'helm-org-headings-class
-    :filtered-candidate-transformer 'helm-org-startup-visibility
-    :parents parents
-    :candidates filenames))
 (defun helm-org--get-candidates-in-file (filename &optional fontify nofname parents)
   (with-current-buffer (pcase filename
                          ((pred bufferp) filename)
@@ -324,7 +326,12 @@ will be refiled."
   (with-helm-alive-p
     (helm-exit-and-execute-action 'helm-org--refile-heading-to)))
 (put 'helm-org-run-refile-heading-to 'helm-only t)
+
 
+;;; Commands
+;;
+;;
+
 ;;;###autoload
 (defun helm-org-agenda-files-headings ()
   "Preconfigured Helm for agenda files."
@@ -337,9 +344,10 @@ will be refiled."
                             collect (helm-basename f))))
     (when (or (null autosaves)
               helm-org-ignore-autosaves
-              (y-or-n-p (format "%s have auto save data, continue?"
-                                (mapconcat 'identity autosaves ", "))))
-      (helm :sources (helm-source-org-headings-for-files (org-agenda-files))
+              (y-or-n-p (format "%s have auto save data, continue? "
+                                (mapconcat #'identity autosaves ", "))))
+      (helm :sources (helm-source-org-headings-for-files
+                      (org-agenda-files))
             :candidate-number-limit 99999
             :truncate-lines helm-org-truncate-lines
             :buffer "*helm org headings*"))))
@@ -348,26 +356,22 @@ will be refiled."
 (defun helm-org-in-buffer-headings ()
   "Preconfigured Helm for buffer headings."
   (interactive)
-  (let (helm-org-show-filename)
-    (helm :sources (helm-source-org-headings-for-files
-                    (list (current-buffer)))
-          :candidate-number-limit 99999
-          :preselect (helm-org-in-buffer-preselect)
-          :truncate-lines helm-org-truncate-lines
-          :buffer "*helm org inbuffer*")))
+  (helm :sources (helm-source-org-headings-for-files
+                  (list (current-buffer)))
+        :candidate-number-limit 99999
+        :preselect (helm-org-in-buffer-preselect)
+        :truncate-lines helm-org-truncate-lines
+        :buffer "*helm org inbuffer*"))
 
 ;;;###autoload
 (defun helm-org-parent-headings ()
   "Preconfigured Helm for parent headings of the current heading."
   (interactive)
-  ;; Use a large max-depth to ensure all parents are displayed.
-  (let ((helm-org-headings-min-depth 1)
-        (helm-org-headings-max-depth  50))
-    (helm :sources (helm-source-org-headings-for-files
-                    (list (current-buffer)) t)
-          :candidate-number-limit 99999
-          :truncate-lines helm-org-truncate-lines
-          :buffer "*helm org parent headings*")))
+  (helm :sources (helm-source-org-headings-for-files
+                  (list (current-buffer)) t)
+	:candidate-number-limit 99999
+	:truncate-lines helm-org-truncate-lines
+	:buffer "*helm org parent headings*"))
 
 ;;;###autoload
 (defun helm-org-capture-templates ()
--- a/helm-org.el
+++ b/helm-org.el
@@ -96,6 +96,15 @@ NOTE: This has no effect in `helm-org-in-buffer-headings'."
 ;;; Org headings
 ;;
 ;;
+
+(defun helm-org-in-buffer-preselect ()
+  "Preselect the heading at point."
+  (if (org-at-heading-p)
+      (buffer-substring-no-properties (point-at-bol) (point-at-eol))
+    (save-excursion
+      (outline-previous-visible-heading 1)
+      (buffer-substring-no-properties (point-at-bol) (point-at-eol)))))
+
 (defun helm-org-goto-marker (marker)
   "Go to MARKER showing the entry's context, body and subheadings."
   (switch-to-buffer (marker-buffer marker))
@@ -301,13 +310,6 @@ will be refiled."
       (cl-loop for victim in victims
                do (org-with-point-at victim
                     (org-refile nil nil rfloc))))))
-
-(defun helm-org-in-buffer-preselect ()
-  (if (org-on-heading-p)
-      (buffer-substring-no-properties (point-at-bol) (point-at-eol))
-      (save-excursion
-        (outline-previous-visible-heading 1)
-        (buffer-substring-no-properties (point-at-bol) (point-at-eol)))))
 
 (defun helm-org-run-refile-heading-to ()
   "Refile one or more entries to the selected heading."
--- a/helm-org.el
+++ b/helm-org.el
@@ -25,8 +25,7 @@
 (eval-when-compile
   (require 'org-macs))
 
-(declare-function org-agenda-switch-to "org-agenda.el")
-
+
 (defgroup helm-org nil
   "Org related functions for Helm."
   :group 'helm)
@@ -85,7 +84,11 @@ NOTE: This has no effect in `helm-org-in-buffer-headings'."
 ;;; Org capture templates
 ;;
 ;;
+
 (defvar org-capture-templates)
+(defvar org-capture-templates-contexts)
+(declare-function org-capture-upgrade-templates "org-capture" (templates))
+
 (defun helm-source-org-capture-templates ()
   (helm-build-sync-source "Org Capture Templates:"
     :candidates (cl-loop for template in org-capture-templates
--- a/helm-org.el
+++ b/helm-org.el
@@ -108,10 +108,11 @@ NOTE: This has no effect in `helm-org-in-buffer-headings'."
   (org-show-children))
 
 (defun helm-org--open-heading-in-indirect-buffer (marker)
-  (helm-org-goto-marker marker)
   "Go to MARKER and create an indirect buffer of the current subtree."
+  (switch-to-buffer (marker-buffer marker))
+  (goto-char (marker-position marker))
+  (org-show-context)
   (org-tree-to-indirect-buffer)
-
   ;; Put the non-indirect buffer at the bottom of the prev-buffers
   ;; list so it won't be selected when the indirect buffer is killed
   (set-window-prev-buffers nil (append (cdr (window-prev-buffers))
--- a/helm-org.el
+++ b/helm-org.el
@@ -80,11 +80,14 @@ NOTE: This has no effect in `helm-org-in-buffer-headings'."
 (declare-function org-capture-upgrade-templates "org-capture" (templates))
 
 (defun helm-source-org-capture-templates ()
-  (helm-build-sync-source "Org Capture Templates:"
-    :candidates (cl-loop for template in org-capture-templates
+  (helm-build-sync-source "Org Capture Templates"
+    :candidates (cl-loop for template in (org-contextualize-keys
+                                          (org-capture-upgrade-templates org-capture-templates)
+                                          org-capture-templates-contexts)
                          collect (cons (nth 1 template) (nth 0 template)))
-    :action '(("Do capture" . (lambda (template-shortcut)
-                                (org-capture nil template-shortcut))))))
+    :action '(("Select template" . (lambda (template-shortcut)
+                                     (org-capture nil template-shortcut))))))
+
 
 ;;; Org headings
 ;;
--- a/helm-org.el
+++ b/helm-org.el
@@ -188,12 +188,16 @@ when called from the `helm-org-agenda-files-headings' command."
    (keymap :initform 'helm-org-headings-map)
    (group :initform 'helm-org)))
 
+
+;; Actions
+;;
+;;
+
 (defun helm-org-goto-marker (marker)
   "Go to MARKER showing the entry's context, body and subheadings."
   (switch-to-buffer (marker-buffer marker))
   (goto-char (marker-position marker))
   (org-show-context)
-  (re-search-backward "^\\*+ " nil t)
   (org-show-entry)
   (org-show-children))
 
@@ -214,65 +218,6 @@ when called from the `helm-org-agenda-files-headings' command."
   (with-helm-alive-p
     (helm-exit-and-execute-action #'helm-org--open-heading-in-indirect-buffer)))
 (put 'helm-org-run-open-heading-in-indirect-buffer 'helm-only t)
-
-(defun helm-org--get-candidates-in-file (filename &optional fontify nofname parents)
-  (with-current-buffer (pcase filename
-                         ((pred bufferp) filename)
-                         ((pred stringp) (find-file-noselect filename t)))
-    (let ((match-fn (if fontify
-                        #'match-string
-                      #'match-string-no-properties))
-          (search-fn (lambda ()
-                       (re-search-forward
-                        org-complex-heading-regexp nil t)))
-          (file (unless (or (bufferp filename) nofname)
-                  (concat (helm-basename filename) ":"))))
-      (when parents
-        (add-function :around (var search-fn)
-                      (lambda (old-fn &rest args)
-                                (when (org-up-heading-safe)
-                                  (apply old-fn args)))))
-      (save-excursion
-        (save-restriction
-          (unless (and (bufferp filename)
-                       (buffer-base-buffer filename))
-            ;; Only widen direct buffers, not indirect ones.
-            (widen))
-          (unless parents (goto-char (point-min)))
-          ;; clear cache for new version of org-get-outline-path
-          (and (boundp 'org-outline-path-cache)
-               (setq org-outline-path-cache nil))
-          (cl-loop with width = (window-width (helm-window))
-                   while (funcall search-fn)
-                   for beg = (point-at-bol)
-                   for end = (point-at-eol)
-                   when (and fontify
-                             (null (text-property-any
-                                    beg end 'fontified t)))
-                   do (jit-lock-fontify-now beg end)
-                   for level = (length (match-string-no-properties 1))
-                   for heading = (funcall match-fn 4)
-                   if (and (>= level helm-org-headings-min-depth)
-                           (<= level helm-org-headings-max-depth))
-                   collect `(,(propertize
-                               (if helm-org-format-outline-path
-                                   (org-format-outline-path
-                                    ;; org-get-outline-path changed in signature and behaviour since org's
-                                    ;; commit 105a4466971. Let's fall-back to the new version in case
-                                    ;; of wrong-number-of-arguments error.
-                                    (condition-case nil
-                                        (append (apply #'org-get-outline-path
-                                                       (unless parents
-                                                         (list t level heading)))
-                                                (list heading))
-                                      (wrong-number-of-arguments
-                                       (org-get-outline-path t t)))
-                                    width file)
-                                   (if file
-                                       (concat file (funcall match-fn  0))
-                                       (funcall match-fn  0)))
-                               'helm-real-display heading)
-                              . ,(point-marker))))))))
 
 (defun helm-org-insert-link-to-heading-at-marker (marker)
   (with-current-buffer (marker-buffer marker)
--- a/helm-org.el
+++ b/helm-org.el
@@ -175,6 +175,19 @@ when called from the `helm-org-agenda-files-headings' command."
                         (helm-org-get-parent-headings))
                     (helm-org-get-candidates filenames)))))
 
+(defclass helm-org-headings-class (helm-source-sync)
+  ((match :initform (lambda (candidate)
+                      (string-match
+                       helm-pattern
+                       (helm-aif (get-text-property 0 'helm-real-display candidate)
+                           it
+                         candidate))))
+   (candidate-transformer :initform 'helm-org-candidate-transformer)
+   (help-message :initform 'helm-org-headings-help-message)
+   (action :initform 'helm-org-headings-actions)
+   (keymap :initform 'helm-org-headings-map)
+   (group :initform 'helm-org)))
+
 (defun helm-org-goto-marker (marker)
   "Go to MARKER showing the entry's context, body and subheadings."
   (switch-to-buffer (marker-buffer marker))
@@ -202,29 +215,6 @@ when called from the `helm-org-agenda-files-headings' command."
     (helm-exit-and-execute-action #'helm-org--open-heading-in-indirect-buffer)))
 (put 'helm-org-run-open-heading-in-indirect-buffer 'helm-only t)
 
-(defclass helm-org-headings-class (helm-source-sync)
-  ((parents
-    :initarg :parents
-    :initform nil
-    :custom boolean)
-   (match :initform
-          (lambda (candidate)
-            (string-match
-             helm-pattern
-             (helm-aif (get-text-property 0 'helm-real-display candidate)
-                 it
-               candidate))))
-   (help-message :initform 'helm-org-headings-help-message)
-   (action :initform 'helm-org-headings-actions)
-   (keymap :initform 'helm-org-headings-map)
-   (group :initform 'helm-org)))
-
-(defmethod helm--setup-source :after ((source helm-org-headings-class))
-  (let ((parents (slot-value source 'parents)))
-    (setf (slot-value source 'candidate-transformer)
-          (lambda (candidates)
-            (let ((cands (helm-org-get-candidates candidates parents)))
-              (if parents (nreverse cands) cands))))))
 (defun helm-org--get-candidates-in-file (filename &optional fontify nofname parents)
   (with-current-buffer (pcase filename
                          ((pred bufferp) filename)
--- a/helm-org.el
+++ b/helm-org.el
@@ -106,6 +106,18 @@ NOTE: This has no effect in `helm-org-in-buffer-headings'."
         (t
          (match-string-no-properties 0))))
 
+(defun helm-org-get-parent-headings ()
+  "Return all parent headings of the current heading.
+Also return their position in the buffer as marker objects."
+  (let (candidates)
+    (save-excursion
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks for your work, can you explain what have changed exactly?

jagrg:
For the user not much has changed. What I was hoping to accomplish was to make the source code easier to read. More specifically, I changed the way helm-org collects parent headings (helm-org-get-parent-headings), and added a new function for showing filenames (helm-org-candidate-transformer) (I was thinking perhaps we should use a different face for filenames as well). I also changed the capture code to use org-capture-templates-contexts, and introduced a new function (helm-org-format-heading) for displaying candidates' according to the user's settings.

BTW we still have to figure out how to make helm faster when helm-org-headings-fontify is t.

thierryvolpiatto:
Jonathan Gregory <notifications@github.com> writes:

> For the user not much has changed. What I was hoping to accomplish was
> to make the source code easier to read. More specifically, I changed
> the way helm-org collects parent headings
> (helm-org-get-parent-headings), and added a new function for showing
> filenames (helm-org-candidate-transformer) (I was thinking perhaps we
> should use a different face for filenames as well). I also changed the
> capture code to use org-capture-templates-contexts, and introduced a
> new function (helm-org-format-heading) for displaying candidates'
> according to the user's settings.

Ok, let merge this, I wondered if you changes was compatible with
emacs-25<=>26, but we will see if someone complains.

> BTW we still have to figure out how to make helm faster when helm-org-headings-fontify is t.

I have not much time actually, but yes I could look into this at the end
of summer.

Thanks.

> —


--
Thierry

Get my Gnupg key:

thierryvolpiatto:
Thanks, merged.

jagrg:
Thanks, Thierry. If you need any help let me know.

alphapapa:
@thierryvolpiatto @jagrg This is another PR that should have been more carefully reviewed first.  There are breaking changes, inefficient code that will likely reduce performance, and some important features have been simply deleted, which users depend on for managing performance and candidate selection.

> BTW we still have to figure out how to make helm faster when helm-org-headings-fontify is t.

I have no performance problem with this setting enabled, and I use it extensively.  This is why changes like this should be minimal, atomic, and discussed beforehand, rather than potentially introducing serious performance regressions.

I don't know where to start here.  This PR should have been split up into smaller ones and discussed before merging.  

The author's explanation of the changes is also concerning.  It starts with, "What I was hoping to accomplish was to make the source code easier to read," but rather than being aesthetic changes like indentation and line breaks, it *rewrites working code*, which is useless churn that introduces new bugs.  Then it changes and adds features in a PR that is supposed to be about aesthetics and clarity.

I would recommend reverting this PR, because it deletes working features, and unnecessarily rewrites code that appears to introduce regressions in functionality and performance.

jagrg:
On 04 Jul 2019, alphapapa <notifications@github.com> wrote:

> @thierryvolpiatto @jagrg This is another PR that should have been more carefully reviewed first.  There are breaking changes, inefficient code that will likely reduce performance, and some important features have been simply deleted, which users depend on for managing performance and candidate selection.

Can you pinpoint the commits that are problematic for you?

>
>> BTW we still have to figure out how to make helm faster when helm-org-headings-fontify is t.
>
> I have no performance problem with this setting enabled, and I use it extensively.  This is why changes like this should be minimal, atomic, and discussed beforehand, rather than potentially introducing serious performance regressions.

This is an old problem in helm-org so I'm surprised. Are you saying there's no difference in performance for you when helm-org-headings-fontify is t? This happens with large Org files with many headings, but only the first time the Helm buffer is created. Subsequent calls are not affected unless the buffer is killed.

>
> I don't know where to start here.  This PR should have been split up into smaller ones and discussed before merging.  
>
> The author's explanation of the changes is also concerning.  It starts with, "What I was hoping to accomplish was to make the source code easier to read," but rather than being aesthetic changes like indentation and line breaks, it *rewrites working code*, which is useless churn that introduces new bugs.  Then it changes and adds features in a PR that is supposed to be about aesthetics and clarity.

I wasn't making aesthetic changes in this PR as the title suggests. You're clearly twisting my words here. Can you be more specific. What bugs do you see?

alphapapa:
> This is an old problem in helm-org so I'm surprised. Are you saying there's no difference in performance for you when helm-org-headings-fontify is t? 

No, I'm saying that the performance is not a problem for me when the setting is enabled.  This is so even though I have many, very large Org files with thousands of headings each.

> This happens with large Org files with many headings, but only the first time the Helm buffer is created. Subsequent calls are not affected unless the buffer is killed.

That sounds to me like the fontification is being done differently than it has in the past, or in other, similar code.  We should benchmark the performance and try to optimize it.  These macros should be helpful: https://github.com/alphapapa/emacs-package-dev-handbook#profiling--optimization

> I wasn't making aesthetic changes in this PR as the title suggests. You're clearly twisting my words here. Can you be more specific. What bugs do you see?

We may have different ideas of what "making code easier to read" means.  In any case, rewriting working code must be done carefully and reviewed by someone who is familiar with it to avoid introducing regressions.  For example, you deleted the heading depth limit settings, which are important for users to be able to configure.

Since the PRs have been reverted, please consolidate the commits into new PRs that are purpose-specific, and I'll be glad to review them.  e.g. the ones that simply add new actions should be straightforward to merge.  The ones that delete and rewrite code will need careful review and some testing.

jagrg:
> BTW we still have to figure out how to make helm faster when helm-org-headings-fontify is t.

@thierryvolpiatto @alphapapa I know the code needs improvement, but what do you think of this approach. The current buffer doesn't have to be fontified so it's as fast as when helm-org-headings-fontify is nil.

```
(defun helm-org-test-candidates ()
  (let (candidates)
    (save-excursion
      (goto-char (point-min))
      (while (re-search-forward org-complex-heading-regexp nil t)
	(let ((star-face (org-get-at-bol 'face))
	      (level (length (match-string 1))))
	  (when (and (>= level helm-org-headings-min-depth)
		     (<= level helm-org-headings-max-depth))
	    (push (cons
		   (let* ((line (match-string 0))
			  (level-face (intern (format "org-level-%s" level)))
			  (priority (match-string 3))
			  (priority-face (org-get-priority-face priority))
			  ;; FIXME: No fontification if the first word
			  ;; of the headline is upcased
			  (headline (match-string 4))
			  (kwd (match-string 2))
			  (kwd-face (org-get-todo-face kwd))
			  (tags (match-string 5))
			  (col1 (match-end 4))
			  (col2 (match-beginning 5)))
		     (concat
		      ;; This part removes leading stars (just testing...)
		      (org-add-props (replace-regexp-in-string
		      		      "^\\(\\**\\)\\(\\* \\)\\(.+\\)"
		      		      (org-add-props (make-string (length (match-string 1)) ?\s)
		      			  nil 'face star-face)
		      		      line)
		      	  nil 'face level-face)

		      (replace-regexp-in-string
		       "^\\(\\**\\)\\(\\* \\)\\(.+\\)"
		       (concat "\\2"
			       (when kwd
				 (concat (org-add-props kwd nil 'face kwd-face) " "))
			       (when priority
				 (concat (org-add-props priority nil 'face priority-face) " "))
			       (when headline
				 (org-add-props headline nil 'face level-face))
			       ;; Try to align tags correctly
			       (when (and col1 col2)
				 (make-string (- col2 col1) ?\s))
			       (when tags ; tag face?
				 (org-add-props tags nil 'face level-face)))
		       (org-add-props line  nil 'face level-face))))
		   
		   (save-excursion
		     (forward-line 0)
		     (point-marker)))
		  candidates))))
      (nreverse candidates))))

(defun helm-org-test ()
  (interactive)
  (helm :sources (helm-build-sync-source "Pretty fast!"
		   :candidates (lambda ()
				 (with-helm-current-buffer
				   (helm-org-test-candidates))))
	:candidate-number-limit 99999))
```

alphapapa:
@jagrg I don't understand why you're writing so much code.  What problem are you trying to solve?  Why are you writing code that does what existing Org functions can do?

Please, use the `bench-multi` macros I linked to benchmark your code and compare it to the existing code.  You should benchmark the code that returns candidates, which is independent from any Helm UI code.  First, find out if the existing code is actually slow.  If it is, use `elp-profile` to find out where it is slow.  Then you can attempt to optimize it, and then benchmark it to see if it's faster.

jagrg:
Thanks for the feedback.

> What problem are you trying to solve?

I'm trying to make Helm faster when helm-org-headings-fontify is t. In fact, we probably don't need that variable.

> Why are you writing code that does what existing Org functions can do? 

Can you be more specific here.

> Why are you doing things like replace-regexp-in-string to remove leading stars when you've already matched against org-complex-heading-regexp which puts the stars in a separate match group?

It puts all stars in one group. What I was trying to do was to hide the first N-1 stars. In any case, my point with all this is that we don't have to fontify the entire buffer to fontify the candidates.

> Please, use the bench-multi macros I linked to benchmark your code and compare it to the existing code. You should benchmark the code that returns candidates, which is independent from any Helm UI code. First, find out if the existing code is actually slow. If it is, use elp-profile to find out where it is slow. Then you can attempt to optimize it, and then benchmark it to see if it's faster.

These macros are definitely useful, but I'm pretty sure the problem is having to fontify the entire buffer (see jit-lock-fontify-now in the current code).

> If the existing code really needs to be changed, we should first look at existing code in the current version of Org that does approximately the same thing, e.g. the code that gathers refile candidates, and reuse it as much as possible.

I agree with reusing existing code as much as possible. I'll have another look at the source code.

alphapapa:
> I'm trying to make Helm faster when helm-org-headings-fontify is t. In fact, we probably don't need that variable.

We certainly do need that variable, because some users don't want headings fontified.  That's the third setting you've deleted or proposed deleting.  Please stop trying to rip out features you don't personally use.  

> Can you be more specific here.

You're using Helm.  Use it to explore the variety of `org-*` functions provided by `org-mode`.  You'll find many useful functions that we can use.

> we don't have to fontify the entire buffer to fontify the candidates.

No, of course we don't.  See e.g. `helm-org-rifle-fontify-like-in-org-mode` and `org-format-outline-path`.

> These macros are definitely useful, but I'm pretty sure the problem is having to fontify the entire buffer (see jit-lock-fontify-now in the current code).

In `helm-org--get-candidates-in-file` I see:

```el
(jit-lock-fontify-now beg end)
```

In which `beg` and `end` are `(point-at-bol)` and `(point-at-eol)`.

In any case, none of this matters without numbers.  If you want to optimize the existing code, that's great.  Your plan should be roughly this:

1.  Prepare a large Org file with thousands of headings as a source of test data.  Keep this file unchanged throughout all testing and development.
2.  Benchmark the existing code to see if it's slow.  If so, profile it to see where it is slow.
3.  If necessary and possible, attempt to optimize it, then benchmark it to see if performance is improved.
4.  If it is improved, verify that the results are exactly the same as the original code with every combination of relevant settings.  The macros I linked can verify that results are the same.
5.  If the results are the same, send a PR with the improved code.  If it looks good, it will be merged, and everyone will benefit.

thierryvolpiatto:

> In helm-org--get-candidates-in-file I see:
>
> (jit-lock-fontify-now beg end)
>
> In which beg and end are (point-at-bol) and (point-at-eol).

@jagrg Note that here, as Adam pointed we are jit-lock fontifying line
by line, not the whole buffer.
What may be costly and probably unnecessary is this test:

    (null (text-property-any beg end 'fontified t))

as jit-lock probably do the same test or similar before proceeding.

-- 
Thierry

Get my Gnupg key:

jagrg:
On 09 Jul 2019, alphapapa <notifications@github.com> wrote:

> We certainly do need that variable, because some users don't want headings fontified.  That's the third setting you've deleted or proposed deleting.  Please stop trying to rip out features you don't personally use.

That's why I said probably. Helm buffers are usually fontified so I didn't think that variable was there as a feature. I thought it was there as a way for users to avoid the performance issue we are discussing here. In any case, I'm fine with keeping it as a feature. Also, the only variables I removed were the helm-org-headings-*-depth variables and that was a mistake. I thought they were linked to the function that collects parent headings, but obviously they serve their own purpose. I'm also fine with keeping them.

>> we don't have to fontify the entire buffer to fontify the candidates.
>
> No, of course we don't.  See e.g. `helm-org-rifle-fontify-like-in-org-mode` and `org-format-outline-path`.

Thanks. I tried both and indeed both are faster than what we use now. I benchmarked helm-org--get-candidates-in-file and my solution above (I changed it slightly) is still faster (see results below).

;; With the current code
;; "Elapsed time: 10.006602s"

;; With helm-org-rifle-fontify-like-in-org-mode
;; "Elapsed time: 1.883376s"

;; With (org-format-outline-path (org-get-outline-path t t))
;; "Elapsed time: 0.846909s"

;; With jagrg's solution
;; "Elapsed time: 0.176008s"

;; With the current code and helm-org-headings-fontify set to nil
;; "Elapsed time: 0.084493s"

> In any case, none of this matters without numbers.  If you want to optimize the existing code, that's great.  Your plan should be roughly this:
>
> 1.  Prepare a large Org file with thousands of headings as a source of test data.  Keep this file unchanged throughout all testing and development.
> 2.  Benchmark the existing code to see if it's slow.  If so, profile it to see where it is slow.
> 3.  If necessary and possible, attempt to optimize it, then benchmark it to see if performance is improved.
> 4.  If it is improved, verify that the results are exactly the same as the original code with every combination of relevant settings.  The macros I linked can verify that results are the same.
> 5.  If the results are the same, send a PR with the improved code.  If it looks good, it will be merged, and everyone will benefit.

I tried to profile helm-org-in-buffer-headings but for some reason elp is not showing the full result. I'll try again later.

jagrg:
On 10 Jul 2019, Thierry Volpiatto <notifications@github.com> wrote:

> @jagrg Note that here, as Adam pointed we are jit-lock fontifying line
> by line, not the whole buffer.

That is correct.

> What may be costly and probably unnecessary is this test:
>
>     (null (text-property-any beg end 'fontified t))
>
> as jit-lock probably do the same test or similar before proceeding.

Actually, it doesn't change much without it: "Elapsed time: 9.932167s"

alphapapa:
> Thanks. I tried both and indeed both are faster than what we use now. I benchmarked helm-org--get-candidates-in-file and my solution above (I changed it slightly) is still faster (see results below).

That sounds interesting.  Please post all the code you used to run the benchmarks.

alphapapa:
Here's an example of the methodology I'm talking about.  I quickly and easily achieved a 25% performance improvement by replacing `point-at-bol/eol` with `match-beginning/end`.  The macros I used helped locate a bottleneck and verified that the changes are faster while returning the same results.

```org
** Benchmarking =helm-org-in-buffer-headings=
:PROPERTIES:
:ID:       a33dd157-af6f-4665-86dd-253f86f227fd
:END:

#+BEGIN_SRC elisp
  (elp-profile 1 '(org- re- point-at-)
    (helm-org--get-candidates-in-file "~/org/inbox.org" t nil))
#+END_SRC

#+RESULTS:
| Function                | Times called |   Total time | Average time |
|-------------------------+--------------+--------------+--------------|
| org-get-outline-path    |         3728 | 1.4257538450 | 0.0003824447 |
| org--get-outline-path-1 |         2741 | 1.1778320409 | 0.0004297088 |
| org-up-heading-safe     |         1863 | 1.0083470530 | 0.0005412490 |
| org-outline-level       |         2740 | 0.9354403790 | 0.0003414015 |
| re-search-forward       |         1865 | 0.5126548029 | 0.0002748819 |
| org-back-to-heading     |         3727 | 0.2503738079 | 6.717...e-05 |
| point-at-bol            |         1864 | 0.2291523719 | 0.0001229358 |
| re-search-backward      |         3618 | 0.0618359020 | 1.709...e-05 |
| org-format-outline-path |         1864 | 0.0426659889 | 2.288...e-05 |
| org-link-display-format |         1862 | 0.0390958779 | 2.099...e-05 |
| org-add-props           |         3208 | 0.0028750849 | 8.962...e-07 |
| point-at-eol            |         1864 | 0.0011298810 | 6.061...e-07 |

*** Avoiding =point-at=

#+BEGIN_SRC elisp
  (defun helm-org--get-candidates-in-file--avoiding-point-at (filename &optional fontify nofname parents)
    (with-current-buffer (pcase filename
                           ((pred bufferp) filename)
                           ((pred stringp) (find-file-noselect filename t)))
      (let ((match-fn (if fontify
                          #'match-string
                        #'match-string-no-properties))
            (search-fn (lambda ()
                         (re-search-forward
                          org-complex-heading-regexp nil t)))
            (file (unless nofname
                    (concat (helm-basename filename) ":"))))
        (when parents
          (add-function :around (var search-fn)
                        (lambda (old-fn &rest args)
                          (when (org-up-heading-safe)
                            (apply old-fn args)))))
        (save-excursion
          (save-restriction
            (unless (and (bufferp filename)
                         (buffer-base-buffer filename))
              ;; Only widen direct buffers, not indirect ones.
              (widen))
            (unless parents (goto-char (point-min)))
            ;; clear cache for new version of org-get-outline-path
            (and (boundp 'org-outline-path-cache)
                 (setq org-outline-path-cache nil))
            (cl-loop with width = (window-width (helm-window))
                     while (funcall search-fn)
                     for beg = (match-beginning 0)
                     for end = (match-end 0)
                     when (and fontify
                               (null (text-property-any
                                      beg end 'fontified t)))
                     do (jit-lock-fontify-now beg end)
                     for level = (length (match-string-no-properties 1))
                     for heading = (funcall match-fn 4)
                     if (and (>= level helm-org-headings-min-depth)
                             (<= level helm-org-headings-max-depth))
                     collect `(,(propertize
                                 (if helm-org-format-outline-path
                                     (org-format-outline-path
                                      ;; org-get-outline-path changed in signature and behaviour since org's
                                      ;; commit 105a4466971. Let's fall-back to the new version in case
                                      ;; of wrong-number-of-arguments error.
                                      (condition-case nil
                                          (append (apply #'org-get-outline-path
                                                         (unless parents
                                                           (list t level heading)))
                                                  (list heading))
                                        (wrong-number-of-arguments
                                         (org-get-outline-path t t)))
                                      width file)
                                   (if file
                                       (concat file (funcall match-fn  0))
                                     (funcall match-fn  0)))
                                 'helm-real-display heading)
                               . ,(point-marker))))))))

  (elp-profile 1 '(org- re- point-at-)
    (helm-org--get-candidates-in-file--avoiding-point-at "~/org/inbox.org" t nil))
#+END_SRC

#+RESULTS:
| Function                | Times called |   Total time | Average time |
|-------------------------+--------------+--------------+--------------|
| org-get-outline-path    |         3728 | 0.8241513389 | 0.0002210706 |
| org--get-outline-path-1 |         2741 | 0.5730549709 | 0.0002090678 |
| re-search-forward       |         1865 | 0.5099838330 | 0.0002734497 |
| org-up-heading-safe     |         1863 | 0.4038001449 | 0.0002167472 |
| org-outline-level       |         2740 | 0.3312434529 | 0.0001208917 |
| org-back-to-heading     |         3727 | 0.2532743209 | 6.795...e-05 |
| re-search-backward      |         3618 | 0.0618765169 | 1.710...e-05 |
| org-format-outline-path |         1864 | 0.0422899740 | 2.268...e-05 |
| org-link-display-format |         1862 | 0.0394865569 | 2.120...e-05 |
| org-add-props           |         3208 | 0.0028834179 | 8.988...e-07 |

*** Comparing

#+BEGIN_SRC elisp
  (bench-multi-lexical :times 1 :ensure-equal t
    :forms (("Original" (helm-org--get-candidates-in-file "~/org/inbox.org" t nil))
            ("Avoiding point-at-" (helm-org--get-candidates-in-file--avoiding-point-at "~/org/inbox.org" t nil))))
#+END_SRC

#+RESULTS:
| Form               | x faster than next | Total runtime | # of GCs | Total GC runtime |
|--------------------+--------------------+---------------+----------+------------------|
| Avoiding point-at- | 1.26               |      1.282881 |        0 |                0 |
| Original           | slowest            |      1.618544 |        0 |                0 |
```

jagrg:
The current code checks if the match (heading) is fontified and skips fontification if the predicate returns t, so assuming the buffer is kept open all subsequent calls are not affected. Now, if we were to use helm-org-rifle-fontify-like-in-org-mode to fontify the matches, which I think would require the least changes in the current code, the first call would be faster and all subsequent calls slower, so I'm not sure what to do here. Perhaps we could fontify the matches in the background and use helm-org-rifle-fontify-like-in-org-mode until the process is complete. Just a thought.

BTW thanks for breaking down your methodology. Very useful.

jagrg:
Here's an example of what I mean. Thierry, what would you suggest using in place of `run-with-timer` to fontify the headings in the background? I know you're involved with emacs-async, that's why I'm asking, but don't worry if you don't have time now. Also, `helm-org-rifle-fontify-like-in-org-mode` fontifies TODO and DONE keywords only, but I'm guessing that's easy to fix. Tested with ./emacs-helm.sh using Org versions 9.1.8 and 9.2.4:

First call: "Elapsed time: 0.454719s"
Second call: "Elapsed time: 0.027630s"

```
(require 'helm-org)
(setq helm-org-headings-fontify t)

(defconst helm-org-rifle-fontify-buffer-name " *helm-org-rifle-fontify*")

(defun helm-org-rifle-fontify-like-in-org-mode (s &optional odd-levels)
  "Fontify string S like in Org-mode."
  (let ((buffer (get-buffer helm-org-rifle-fontify-buffer-name)))
    (unless buffer
      (setq buffer (get-buffer-create helm-org-rifle-fontify-buffer-name))
      (with-current-buffer buffer
        (org-mode)))
    (with-current-buffer buffer
      (erase-buffer)
      (insert s)
      (let ((org-odd-levels-only odd-levels))
        ;; FIXME: "Warning: ‘font-lock-fontify-buffer’ is for interactive use only; use
        ;; ‘font-lock-ensure’ or ‘font-lock-flush’ instead."
        (font-lock-fontify-buffer)
	(buffer-string)))))

(defun helm-org--get-candidates-in-file (filename &optional fontify nofname parents)
  (with-current-buffer (pcase filename
			 ((pred bufferp) filename)
			 ((pred stringp) (find-file-noselect filename t)))
    (let ((match-fn (if fontify
                        #'match-string
                      #'match-string-no-properties))
	  (search-fn (lambda ()
		       (re-search-forward
			org-complex-heading-regexp nil t)))
	  (file (unless (or (bufferp filename) nofname)
		  (concat (helm-basename filename) ":"))))

      (when parents
	(add-function :around (var search-fn)
		      (lambda (old-fn &rest args)
			(when (org-up-heading-safe)
			  (apply old-fn args)))))
      (save-excursion
	(save-restriction
	  (unless (and (bufferp filename)
		       (buffer-base-buffer filename))
	    ;; Only widen direct buffers, not indirect ones.
	    (widen))
	  (unless parents (goto-char (point-min)))
	  (cl-loop with width = (window-width (helm-window))
		   while (funcall search-fn)
		   for beg = (match-beginning 0)
		   for end = (match-end 0)
		   when (and fontify
			     (null (text-property-any
				    beg end 'fontified t)))
		   do (run-with-timer
		       1 nil `(lambda ()
				(with-current-buffer ,(current-buffer)
				  (jit-lock-fontify-now))))
		   for level = (length (match-string-no-properties 1))
		   for heading = (funcall match-fn 0)
		   if (and (>= level helm-org-headings-min-depth)
			   (<= level helm-org-headings-max-depth))
		   collect `(,(propertize
			       (if helm-org-format-outline-path
                                   (org-format-outline-path
                                    ;; org-get-outline-path changed in signature and behaviour since org's
                                    ;; commit 105a4466971. Let's fall-back to the new version in case
                                    ;; of wrong-number-of-arguments error.
                                    (condition-case nil
                                        (append (apply #'org-get-outline-path
                                                       (unless parents
                                                         (list t level heading)))
                                                (list heading))
                                      (wrong-number-of-arguments
                                       (org-get-outline-path t t)))
                                    width file)
				 (concat (if file file "")
					 (if (text-property-any
					      beg end 'fontified t)
					     heading
					   (helm-org-rifle-fontify-like-in-org-mode heading))))
			       'helm-real-display heading)
			     . ,(point-marker))))))))
```

alphapapa:
There are two issues regarding potentially fontifying headings in the background (or more accurately, fontifying the yet-unfontified parts of the buffer containing the headings we need):

1.  When, or under what conditions, should we begin doing so?  

    If we wait until the user has called one of Helm's commands, it's too late--we need the results now, and the headings must be fontified before they can be displayed, so at that point, there would be no benefit to doing so in the background.  

    Instead, we could fontify the headings before the Helm command is run, but that raises two issues: a) that is outside of Helm's scope; and b) what if the user runs a Helm command that would display those headings before they are finished fontifying?  We would have to check whether they have been fontified and fontify any unfontified ones anyway.

2.  Emacs Lisp doesn't have true multithreading support.  Only one Elisp thread can run at a time.  Elisp threads work by yielding.  

    So while we could write a function to fontify one heading and yield back to the main thread, this again raises the issue above: when a command is running, we need the headings now, so there's no benefit to trying to fontify them in a separate thread.  In fact, doing so would probably introduce additional overhead.

So I think this brings back to where we are now: either we test each heading match and use `jit-lock-fontify-now` on parts of the buffer that aren't already fontified, or we use something like `helm-org-rifle-fontify-like-in-org-mode` or `org-format-outline-path`, which fontify headings separately.  The questions are which of those choices is fastest, and whether any further optimizations can be done.

However, I do like the idea of preemptively, or opportunistically, fontifying Org buffers in a background thread.  If we combined that with a buffer-local variable that is set `t` when fontification is complete, we could avoid testing the fontification of every heading, which might give a nice performance increase.

However, that raises its own potential issues.  For example, buffers' contents can change at any time.  What if new text is inserted into an off-screen portion of a buffer?  It may not be fontified until the user scrolls it into view.  How would we ensure that the new content is fontified before we search it for headings when the user calls a command?  

We probably can't; we could try some hacky stuff with hooks and timers, but there would likely be edge cases and potential performance issues, and we definitely couldn't make that part of Helm by default.

So, once again, `goto now;`.  We either test and `jit-lock`, or fontify with other functions.  Which way is fastest, and can we optimize any of them?

> First call: "Elapsed time: 0.454719s"
> Second call: "Elapsed time: 0.027630s"

That's nice, but again, please post the entire code you used to benchmark them, like I did in the example above.  The results should be reproducible by us on our own data, and the code should be idempotent.  Also, we don't know what data you're running that code on, so those numbers aren't very meaningful to us.  Are you running it on a truly huge Org file?  Multiple files?  You're not letting us see your methodology--you're asking us to take your numbers for granted.

Note as well that calling `jit-lock-fontify-now` on an *entire buffer* when a Helm command is called is *not* practical.  It may be fine in a buffer of a few KB, but in a large buffer with megabytes of data and thousands of lines, it means that we would be *forcing* the *entire buffer* to be fontified, and we would not return *any* results until the *whole thing* is fontified.  Now imagine doing that for several large Org buffers in a single command, and making the user wait for megabytes of Org buffers to be fontified.  That would seem to defeat the purpose of "JIT" fontification.

jagrg:
On 13 Jul 2019, alphapapa <notifications@github.com> wrote:

> There are two issues regarding potentially fontifying headings in the background (or more accurately, fontifying the yet-unfontified parts of the buffer containing the headings we need):
>
> 1.  When, or under what conditions, should we begin doing so?  
>
>     If we wait until the user has called one of Helm's commands, it's too late--we need the results now, and the headings must be fontified before they can be displayed, so at that point, there would be no benefit to doing so in the background.  

I wouldn't say it's too late. We can use helm-org-rifle-fontify-like-in-org-mode on the first call as a fallback.

>     Instead, we could fontify the headings before the Helm command is run, but that raises two issues: a) that is outside of Helm's scope; and b) what if the user runs a Helm command that would display those headings before they are finished fontifying?  We would have to check whether they have been fontified and fontify any unfontified ones anyway.

Not a good idea. I agree.

> 2.  Emacs Lisp doesn't have true multithreading support.  Only one Elisp thread can run at a time.  Elisp threads work by yielding.  
>
>     So while we could write a function to fontify one heading and yield back to the main thread, this again raises the issue above: when a command is running, we need the headings now, so there's no benefit to trying to fontify them in a separate thread.  In fact, doing so would probably introduce additional overhead.

I'm not familiar with multithreading so I'm not going to comment.

> So I think this brings back to where we are now: either we test each heading match and use `jit-lock-fontify-now` on parts of the buffer that aren't already fontified,

Isn't this what the current code does?

> or we use something like `helm-org-rifle-fontify-like-in-org-mode` or `org-format-outline-path`, which fontify headings separately.  The questions are which of those choices is fastest, and whether any further optimizations can be done.

The problem is both are still slower than what we use now (once the matches are fontified that is). IOW the first call is going to faster but all subsequent calls slower. Obviously I'm not taking the optimizations into account here which I'm sure would help.

> However, I do like the idea of preemptively, or opportunistically, fontifying Org buffers in a background thread.  If we combined that with a buffer-local variable that is set `t` when fontification is complete, we could avoid testing the fontification of every heading, which might give a nice performance increase.

Yes! Please give this paragraph some attention.

> However, that raises its own potential issues.  For example, buffers' contents can change at any time.  What if new text is inserted into an off-screen portion of a buffer?  It may not be fontified until the user scrolls it into view.  How would we ensure that the new content is fontified before we search it for headings when the user calls a command?

Well, at least with normal/visible editing, that part of the buffer is already going to fontified, but if that's a problem we can still test fontification as we do now.

There was a similar discussion here (https://github.com/jkitchin/org-ref/issues/647). Perhaps you can skim through it. Apparently we can apply fontification only on parts of the buffer that have been modified by shifting the positions by the length of the text added or removed (cc @vspinu).

[...]

>> First call: "Elapsed time: 0.454719s"
>> Second call: "Elapsed time: 0.027630s"
>
> That's nice, but again, please post the entire code you used to benchmark them, like I did in the example above.  The results should be reproducible by us on our own data, and the code should be idempotent.  Also, we don't know what data you're running that code on, so those numbers aren't very meaningful to us.  Are you running it on a truly huge Org file?  Multiple files?  You're not letting us see your methodology--you're asking us to take your numbers for granted.

Well, the numbers should tell you something comparatively. The difference between the first and second call is a lot smaller. Anyway, my test file has 315 headings, and I used (benchmark 1 '(helm-org--get-candidates-in-file (buffer-file-name) t)) to get those results. I'll see if I can make a test file and share it with you.

> Note as well that calling `jit-lock-fontify-now` on an *entire buffer* when a Helm command is called is *not* practical.

I agree and I'm still grappling with this.

> It may be fine in a buffer of a few KB, but in a large buffer with megabytes of data and thousands of lines, it means that we would be *forcing* the *entire buffer* to be fontified, and we would not return *any* results until the *whole thing* is fontified.  Now imagine doing that for several large Org buffers in a single command, and making the user wait for megabytes of Org buffers to be fontified.  That would seem to defeat the purpose of "JIT" fontification.

You're missing the point. The idea is to use helm-org-rifle-fontify-like-in-org-mode as a fallback until the buffer is fontified, so in theory we should see results even before fontification is complete. Another idea from that thread is to cache all headings and their position and only reparse modified parts of the buffer.

thierryvolpiatto:
Jonathan Gregory <notifications@github.com> writes:

> Here's an example of what I mean. Thierry, what would you suggest
> using in place of run-with-timer

I don't see how using a timer helps to speed up here.

> to fontify the headings in the background? I know you're involved with
> emacs-async, that's why I'm asking, but don't worry if you don't have
> time now.

I don't think beeing async helps here, anyway with emacs-async you can't
work on buffers.

Other alternative to fontify the whole buffer before calling helm-org
doesn't help as well because emacs will hang waiting the fontification
finish.

Could you provide the huge org buffer you are using so that @alphapapa an I
could test as well (remove all confidential stuff).

Also please provide diffs against the current code instead of the whole
function, it is easier to see what changed, even better create a new PR,
doesn't matter if code is wrong, we will delete it afterward if needed.

Thanks for your work on this.

--
Thierry

Get my Gnupg key:

jagrg:
Create a blank test.org file and do M-x testfile. Save and close the file. Then open the file and evaluate (benchmark 1 '(helm-org--get-candidates-in-file (buffer-file-name) t)) to get the result of the first call. Then evaluate once again to get the result of the second call. Then compare the results with the current code.

(defun testfile ()
  (interactive)
  (save-excursion
    (goto-char (point-min))
    (insert "#+title: Test\n")
    (cl-loop for i from 1 to 1000
	     do (insert (format "\n* TODO Heading %d\n** DONE Subheading %d.1" i i)))))

vspinu:
> Apparently we can apply fontification only on parts of the buffer that have been modified by shifting the positions by the length of the text added or removed (cc @vspinu).

Not sure how I can help here but just pointing out that there is no need to track changed regions explicitely for the sake of the font-lock. Font lock already does that in its after-change hook and marks all necessary regions with `fontified nil` property which is checked by re-display in due time. So, if helm-org does need to re-fontify the buffer before re-display does so, all what is needed is to iterate over `fontified nil` regions and call `font-lock-fontify-region` on those.

alphapapa:
@jagrg 

> Yes! Please give this paragraph some attention.

It's a nice idea, but out-of-scope for Helm.  If you want to write a third-party package to fontify buffers in the background, go for it.  If it works well, Helm could add optional support for it.  If it works really well, you could propose it for addition to Org or Emacs itself.

> Another idea from that thread is to cache all headings and their position and only reparse modified parts of the buffer.

Also a nice idea, but also out-of-scope for Helm.  There have also been discussions about in-buffer caching on the org-mode mailing list.  Such a feature would best be implemented in Org directly, but you could implement a prototype as a separate package, and if it worked well, Helm could use it.

> Well, the numbers should tell you something comparatively. The difference between the first and second call is a lot smaller.

> You're missing the point. The idea is to use helm-org-rifle-fontify-like-in-org-mode as a fallback until the buffer is fontified, so in theory we should see results even before fontification is complete.

No, you're missing my point, which is to compare the existing code with your proposals.  I've asked you 3 or 4 times now, and given you a complete example of how to do so with concrete results, but you still haven't done so.  Instead, you've given us a function that inserts 2000 headings without entry content and told us to "compare the results with the current code."  I already have Org files with thousands of headings and tens of thousands of lines of content.  I don't need help with preparing test data.

If you won't benchmark and profile your code as I've requested, I'm not going to spend any more time reviewing your code and going in circles with this discussion.

jagrg:
@vspinu Thanks making that clear. What we're trying to do is to optimize the way we collect headings in an Org file and display them in Helm. ATM Emacs hangs for a few of seconds (depending on the size of file) the first time the Helm buffer is created because we are using [jit-lock-fontify-now](https://github.com/emacs-helm/helm/blob/005c51abcea467f29154810869a884b9ff69c46b/helm-org.el#L249) to fontify each heading before displaying the results.

I proposed a solution [here](https://github.com/emacs-helm/helm/pull/2164#issuecomment-509617750) which doesn't require using jit-lock-fontify-now, but perhaps there's a better way of doing this. For example, I liked what jkitchin did [here](https://github.com/jkitchin/org-ref/commit/9ce113f8e702816a1dc1e380f9a050f24c64786c) with your idea of caching labels and reparsing only modified parts of the buffer, so if you don't mind me borrowing your knowledge, what would you suggest doing here to address the lag?

@alphapapa This is not my full-time job, so please be a bit more patient here.

alphapapa:
> This is not my full-time job, so please be a bit more patient here.

Neither is it mine.  I've been very patient, reviewing your code, correcting regressions you introduced, and writing extensive explanations, but you've ignored my requests to present and test your proposals in a way that we can verify, which isn't respectful of my time.

jagrg:
Thank you all for bearing with me. I was able to reproduce the lag using `./emacs-helm.sh` with Emacs version 26.2 and Org version 9.2.4 with the following settings:

```
(require 'package)
(package-initialize)
(add-to-list 'load-path "~/path/to/org-mode/lisp")
(add-to-list 'load-path "~/path/to/org-mode/contrib/lisp" t)
(require 'helm-config)
(require 'helm-org)
(setq helm-org-headings-fontify t)
(setq org-highlight-latex-and-related '(latex entities))
```

I ran out of breath but AFAICT there's nothing urgent that needs to be done.

PR #2173, opened 2019-07-16
Don't (require 'org) unnecessarily

kiennq:
`(require 'org)` unnecessarily slows the first `helm-goto-char`.
I think we should not force user to pay for that.

Also, I added reveal function for `markdown-mode` too.


kiennq:
cc @alphapapa

thierryvolpiatto:
Thanks.

PR #2174, opened 2019-07-18
Fix bug when preselecting headings

jagrg:
cc: @alphapapa

### DIFF
--- a/helm-org.el
+++ b/helm-org.el
@@ -307,11 +307,14 @@ will be refiled."
                     (org-refile nil nil rfloc))))))
 
 (defun helm-org-in-buffer-preselect ()
-  (if (org-on-heading-p)
-      (buffer-substring-no-properties (point-at-bol) (point-at-eol))
-      (save-excursion
-        (outline-previous-visible-heading 1)
-        (buffer-substring-no-properties (point-at-bol) (point-at-eol)))))
+  "Return the current or closest visible heading as a regexp string."
+  (save-excursion
+    (cond ((org-at-heading-p) (forward-line 0))
+	  ((org-before-first-heading-p)
+	   (outline-next-visible-heading 1))
+	  (t (outline-previous-visible-heading 1)))
+    (regexp-quote (buffer-substring-no-properties (point-at-bol)
+						  (point-at-eol)))))
 
 (defun helm-org-run-refile-heading-to ()
   (interactive)


alphapapa:
1.  What is the bug you're trying to fix?  I guess it's that an error happens if point is before the first heading, but we should be clear about what the problem is.
2.  Why are you using  `regexp-quote`?
3.  I would suggest writing the function like this:
```el
(defun helm-org-in-buffer-preselect ()
  (unless (outline-on-heading-p)
    (if (org-before-first-heading-p)
        (outline-next-heading)
      (outline-previous-visible-heading 1)))
  (org-get-heading t t))
```

jagrg:
Thanks for reviewing the code.

On 18 Jul 2019, alphapapa <notifications@github.com> wrote:

> 1.  What is the bug you're trying to fix?  I guess it's that an error happens if point is before the first heading, but we should be clear about what the problem is.

Sorry, did you not read the commit message, or do you not see the bug?

> 2.  Why are you using  `regexp-quote`?

Please read the commit message.

> 3.  I would suggest writing the function like this:
> ```el
> (defun helm-org-in-buffer-preselect ()
>   (unless (outline-on-heading-p)
>     (if (org-before-first-heading-p)
>         (outline-next-heading)
>       (outline-previous-visible-heading 1)))
>   (org-get-heading t t))
> ```

If we were to use org-get-heading we'd also need tags and TODO keywords to ensure the heading is unique. For example, if there's a heading above the current one with the same headline but different tags Helm would preselect the wrong candidate. The same is true for keywords. Ideally, we could use the point-marker to preselect the right candidate, although I'm not sure this is possible with the current code. Probably not.

Also, with org-get-heading you also get face properties etc. We just need the heading as a string wrapped with regexp-quote to ensure brackets, asterisk etc. are escaped correctly.

alphapapa:
Thanks, I didn't see the commit message.  It's customary to describe the purpose of a PR in the PR message, even if that just means copying the commit message.  :)

Using the point marker is a good idea, if it were possible.  

So this code looks good to me.  I only have one suggestion: since you're calling `(forward-line 0)`, we should be able to use `point` instead of `point-at-bol`, like this:
```el
(defun helm-org-in-buffer-preselect ()
  "Return the current or closest visible heading as a regexp string."
  (save-excursion
    (cond ((org-at-heading-p) (forward-line 0))
	  ((org-before-first-heading-p)
           (outline-next-visible-heading 1))
	  (t (outline-previous-visible-heading 1)))
    (regexp-quote (buffer-substring-no-properties (point)
                                                  (point-at-eol)))))
```

Thanks for fixing this bug.

thierryvolpiatto:
Thanks, merged.
For the `point-at-bol` vs `point` the first `cond` clause ensure `point` is at `bol` with `(forward-line 0)`, but is the next 2 clauses ensuring this?

alphapapa:
@thierryvolpiatto Yes, the commands `outline-next-visible-heading` and `outline-previous-visible-heading` leave point at `bol`.

PR #2199, opened 2019-09-12
Use file local variable to make log use outline mode by default.

cireu:
Add File local variable to debug log when save it to file. So outline mode will be used by default while opening a log file.


thierryvolpiatto:
Thanks, merged!

PR #2207, opened 2019-10-08
Extract mode-line updating functionality.

jjzmajic:
It makes the code more readable and maintainable. It also makes it
possible to disable the mode-line updates with `advice-add` should a
user choose to.


thierryvolpiatto:
Thanks!

PR #2222, opened 2019-11-11
Smonnier defmethod patch

thierryvolpiatto:
Update helm-mode to work with Stefan patch using generic functions.

Note: This don't work on any published Emacs (even Emacs-27 as of now).


thierryvolpiatto:
Related to issue #2165

PR #2239, opened 2019-11-29
fix: Attach load file action to dynamic modules.

cireu:
Only `.elc` and `.el` file have load file action now. Add this for dynamic module file if supported.

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -3214,8 +3214,9 @@ Return candidates prefixed with basename of `helm-input' first."
                . helm-find-files-byte-compile)
               ("Load File(s) `M-L'" . helm-find-files-load-files))
             2))
-          ((string-match "\\.elc?\\'" (helm-aif (helm-marked-candidates)
-                                          (car it) candidate))
+          ((string-match (concat (regexp-opt load-suffixes) "\\'")
+                         (helm-aif (helm-marked-candidates)
+                             (car it) candidate))
            (helm-append-at-nth
             actions
             '(("Load File(s) `M-L'" . helm-find-files-load-files))


thierryvolpiatto:
Zhu Zihao <notifications@github.com> writes:

> Only .elc and .el file have load file action now. Add this for dynamic module file if supported.

Sorry, but I don't understand what you want to do, actually it is
possible to load *.el and *.elc files, with your patch it is only
possible to load *.elc, what did I miss?
What are dynamic module files?

Thanks.

-- 
Thierry

Get my Gnupg key:

cireu:
> Zhu Zihao <notifications@github.com> writes:
> Only .elc and .el file have load file action now. Add this for dynamic module file if supported.
> Sorry, but I don't understand what you want to do, actually it is possible to load *.el and *.elc files, with your patch it is only possible to load *.elc, what did I miss? What are dynamic module files? Thanks.
> […](#)
> -- Thierry Get my Gnupg key: gpg --keyserver pgp.mit.edu --recv-keys 59F29997

It was introduced in Emacs 25. Allows user to load some specific dynamic libraries to extend Emacs's ability. 

It requires user to build Emacs with configure switch `--with-modules`. It is not enable by default so you didn't notice this.

For linux, the value of  variable `load-suffixes` of a Emacs with dynamic module support  is

```
ELISP> load-suffixes
(".elc" ".el" ".so")
```

You can also check manual for more information 

https://www.gnu.org/software/emacs/manual/html_node/elisp/Dynamic-Modules.html

If

thierryvolpiatto:
I guess what you want is instead of:

+          ((string-match (concat (regexp-opt (remove ".el" load-suffixes))
+                                 "\\'")

is:

(string-match (concat (regexp-opt load-suffixes) "\\'")

-- 
Thierry

Get my Gnupg key:

cireu:
@thierryvolpiatto  Please take a look.

thierryvolpiatto:
This time it is correct, thanks!

thierryvolpiatto:
Zhu Zihao <notifications@github.com> writes:

> ELISP> load-suffixes
> (".elc" ".el" ".so")

Ok, thanks I have this in emacs-27 but not in emacs-26.

-- 
Thierry

Get my Gnupg key:

PR #2240, opened 2019-11-29
Travis CI

peterbecich:
Some CI coverage.  This succeeds: https://travis-ci.org/peterbecich/helm/builds/618781081

I've combined some things from https://github.com/abo-abo/swiper/blob/master/.travis.yml and other examples.

Perhaps the other packages in this repository could be covered by additional stages.


helm.el
```diff
@@ -1,5 +1,8 @@
 ;;; helm.el --- Emacs incremental and narrowing framework -*- lexical-binding: t -*-
 
+;; Version: 1.0
```
peterbecich: required for Cask

thierryvolpiatto: Why not putting the right number version e.g. 3.5.6 which is the next version number should be released soon now)?

thierryvolpiatto:
Thanks! Merged.

PR #2277, opened 2020-02-07
Tweaking of `other' helm-split-window-default-side when 1 window

bramadams:
Hi,

When `helm-split-window-default-side` is set to `'other` and only one window is visible in the current frame, `helm-split-window-default-fn` is hard-coded to use 'below for its buffer. For consistency, it seems better to make that default configurable.

This patch adds the `helm-split-window-other-side-when-one-window` defcustom, then uses that variable instead of the hard-coded 'below value in
`helm-split-window-default-fn` when only one window is visible.

Possible values for the defcustom are 'below, 'above, 'left and 'right. The default is 'below, which corresponds to the current behavior.

I hope these changes prove useful for inclusion in helm.

Bram

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -491,7 +491,24 @@ If `helm-full-frame' is non-`nil', it take precedence over this setting.
 See also `helm-split-window-inside-p' and `helm-always-two-windows' that
 take precedence over this.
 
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks, merged.

bramadams:
Thanks a lot!

Issue #2280, opened 2020-02-18
helm-list-elisp-packages wants to upgrade and delete too much by default


obar:
## Expected behavior
Running `helm-list-elisp-packages` should be something like running `package-list-packages`. Nothing should be marked to delete or install by default (or at least with optional configuration for that behavior), no prompt asking y/n for deleting packages.

## Actual behavior (from `emacs-helm.sh` if possible, see note at the bottom)
From `emacs-helm.sh`, first thing that happens is a prompt asking if I really want to delete all packages.

In normal operation I believe this is a prompt asking to delete any package not in `package-selected-packages`, including those installed other ways.

After that, it shows the filtered view for upgradable packages (that's OK, docs explain M-A to view all), but with every package that can be upgraded marked for installation.

## Steps to reproduce (recipe)
`M-x helm-list-elisp-packages`

## Backtraces if any (`M-x toggle-debug-on-error`)
n/a

## Describe versions of Helm, Emacs, operating system, etc.
emacs 26.2 with helm-20200128.1831, Ubuntu

## Are you using `emacs-helm.sh` to reproduce this bug? (yes/no):
yes

## Are you using Spacemacs? (yes/no):
no

thierryvolpiatto:
Ori <notifications@github.com> writes:

> Expected behavior
>
> Running helm-list-elisp-packages should be something like running
> package-list-packages. Nothing should be marked to delete or install
> by default (or at least with optional configuration for that
> behavior), no prompt asking y/n for deleting packages.

You can say no, if you don't want to delete packages.

> Actual behavior (from emacs-helm.sh if possible, see note at the bottom)
>
> From emacs-helm.sh, first thing that happens is a prompt asking if I really want to delete all packages.

All packages? If it asks to delete all packages it is you did something
wrong (e.g. modifying manually package-selected-packages).

> In normal operation I believe this is a prompt asking to delete any
> package not in package-selected-packages, including those installed
> other ways.

If you delete selected package A with dependencies B and C, helm will
ask you if you want to delete packages B and C if they are not used
elsewhere.

So I wonder what is the problem with this as long as you can say yes or
no.

--
Thierry

Get my Gnupg key:
gpg --keyserver pgp.mit.edu --recv-keys 59F29997

obar:
> All packages? 

Yes, `emacs-helm.sh` doesn't have my package list.

What about all of the packages that can be upgraded being marked for installation?

obar:
[...]

PR #2281, opened 2020-02-20
Option for helm-list-elisp-packages filtering to upgrade on start

obar:
Add a new user preference to filter only packages available for
upgrade on start (nil by default).

Related to #2280


thierryvolpiatto:
Thanks, merged!

PR #2283, opened 2020-03-18
Fix a bug when the host part of a TRAMP path ends with a method name

montag451:
Hi @thierryvolpiatto 

I found out a bug in Helm when the host part of a TRAMP path ends by a method name (e.g `/ssh:user@my.domain.ssh:`). Helm considers the path as an "Invalid tramp file name" although it's a valid TRAMP path. This pull request fixes this issue.


thierryvolpiatto:
Thanks, merged!

PR #2325, opened 2020-06-19
Preserve source ordering in `helm-fuzzy-match-sort-fn-preserve-ties-order'

dme:
With:
```
(setq helm-buffers-fuzzy-matching t
  helm-fuzzy-sort-fn 'helm-fuzzy-match-sort-fn-preserve-ties-order)
```
the ordering of buffers in helm-buffers-list is still "shortest match first" rather than the expected "recent match first" once some input is entered. This seems to be because the call to helm-fuzzy-matching-default-sort-fn-1 is incorrect.

I'm new to helm - if this analysis is wrong I'm happy to learn more!

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -4285,7 +4285,7 @@ The default function, `helm-fuzzy-matching-default-sort-fn',
 sorts ties by length, shortest first.  This function may be more
 useful when the order of the candidates is meaningful, e.g. with
 `recentf-list'."
-  (helm-fuzzy-matching-default-sort-fn-1 candidates nil t))
+  (helm-fuzzy-matching-default-sort-fn-1 candidates nil nil t))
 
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
David Edmondson <notifications@github.com> writes:

> With:
>
> (setq helm-buffers-fuzzy-matching t
>   helm-fuzzy-sort-fn 'helm-fuzzy-match-sort-fn-preserve-ties-order)
>
> the ordering of buffers in helm-buffers-list is still "shortest match
> first" rather than the expected "recent match first" once some input
> is entered. This seems to be because the call to
> helm-fuzzy-matching-default-sort-fn-1 is incorrect.

Thanks to catch this, merged, this is wrong since 2017! When this was
introduced. I never noticed this because I don't use this function.

-- 
Thierry

Get my Gnupg key:

PR #2332, opened 2020-06-28
Add hidden flag to rg command

jagrg:
Search hidden files and directories. By default, hidden
files and directories are skipped. Note that if a hidden
file or a directory is whitelisted in an ignore file,
then it will be searched even if this flag isn’t
provided.

### DIFF
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -1444,12 +1444,12 @@ colorization of backend, however it is still supported.
 
 For ripgrep here is the command line to use:
 
-    rg --color=always --smart-case --no-heading --line-number %s %s %s
+    rg --color=always --hidden --smart-case --no-heading --line-number %s %s %s
 
 And to customize colors (always for ripgrep) use something like this:
 
     rg --color=always --colors 'match:bg:yellow' --colors 'match:fg:black'
-\--smart-case --no-heading --line-number %s %s %s
+\--hidden --smart-case --no-heading --line-number %s %s %s
 
 This will change color for matched items from foreground red (the
 default) to a yellow background with a black foreground.  Note


manuel-uberti:
Shouldn't this be for the user to decide? 

I, for instance, wouldn't want `rg` search results with candidates from hidden files, because I may have hidden those files for a reason.

jagrg:
Why is the hidden flag in the default ag command then? I think it's
pretty common for users to grep "hidden" files (.emacs, .bashrc, etc.).

thierryvolpiatto:
Manuel Uberti <notifications@github.com> writes:

> Shouldn't this be for the user to decide?

Yes.

> I, for instance, wouldn't want rg search results with candidates from
> hidden files, because I may have hidden those files for a reason.

I agree with Manuel, perhaps just mentioning the --hidden flag exists is
enough, but well it is in the RG documentation anyway...

-- 
Thierry

Get my Gnupg key:

manuel-uberti:
> Why is the hidden flag in the default ag command then? I think it's pretty common for users to grep "hidden" files (.emacs, .bashrc, etc.).

I don't use `ag` so I didn't know about that default setting. 

Anyway, IIUC your change affects the documentation only here. Is this the intended aim of this PR?

thierryvolpiatto:
Jonathan Gregory <notifications@github.com> writes:

> Why is the hidden flag in the default ag command then?

It should not be the default, up to user to set it up.

> I think it's pretty common for users to grep "hidden" files (.emacs,
> .bashrc, etc.).

So instead let's say it in help if you want but briefly, the goal of
helm documentation is not to duplicate rg/ag documentation, but just to
provide the minimum needed to make ag/rg working in helm.

-- 
Thierry

Get my Gnupg key:

jagrg:
On 28 Jun 2020, Thierry Volpiatto <notifications@github.com> wrote:

>> Why is the hidden flag in the default ag command then?
>
> It should not be the default, up to user to set it up.

In this case, should we remove --hidden from helm-grep-ag-command? It
doesn't make sense to use --hidden in ag and advertise rg without it.

PR #2338, opened 2020-07-13
Do not highlight a sequence of 0’s as a file extension

d125q:
Currently, only a single 0 would be exempt from being highlighted as a
file extension.  E.g., in

```
foo.txt.0
foo.txt.1
foo.txt.2
```

none of the 0, 1, and 2 will be highlighted.  However, in

```
bar.txt.00
bar.txt.01
bar.txt.02
```

the 00 will be highlighted, but the 01 and 02 won’t.  This commit
fixes that.


thierryvolpiatto:
Thanks! Merged.

Issue #2343, opened 2020-07-20
C-x C-b fails if helm-fuzzy-sort-fn/helm-buffer-sort-fn or helm-fuzzy-matching-highlight-fn is nil


Miciah:
## Expected behavior

Customizing `helm-fuzzy-sort-fn` and `helm-fuzzy-matching-highlight-fn` to the nil value should cause Helm not to perform fuzzy sorting or match highlighting, per those options' documentation strings.

## Actual behavior (from `emacs-helm.sh` if possible, see note at the bottom)

Customizing `helm-fuzzy-sort-fn` and `helm-fuzzy-matching-highlight-fn` to the nil value causes some Helm commands, such as `helm-list-buffers`, to fail.

## Steps to reproduce (recipe)

I configure Helm using use-package:

```emacs-lisp
(use-package helm
  :defer nil
  :bind (([remap eshell-list-history] . helm-eshell-history)
         ([remap find-file]           . helm-find-files)
         ([remap list-buffers]        . helm-buffers-list))
  :custom
  (helm-fuzzy-sort-fn nil)
  (helm-fuzzy-matching-highlight-fn nil)
  :config (helm-mode 1))
```

When I start Emacs with the above configuration and press <kbd>C-x C-b</kbd> or <kbd>C-h v</kbd>, Helm fails with an error: `funcall: Symbol’s function definition is void: nil`.

## Backtraces if any (`M-x toggle-debug-on-error`)

```
Debugger entered--Lisp error: (void-function nil)
  nil(#("*Messages*             " 0 10 (face helm-non-file-buffer help-echo "~/" type nofile)))
[...]

PR #2344, opened 2020-07-20
Handle nil value for helm-buffers-sort-fn

Miciah:
...or for `helm-fuzzy-matching-highlight-fn`.  By default, `helm-buffers-sort-fn` is initialized with the value of `helm-fuzzy-sort-fn`, and the documentation string for `helm-fuzzy-sort-fn` states that it may be set to `nil` to indicate that no sorting should be done.  Similarly, the documentation string for `helm-fuzzy-matching-highlight-fn` states that it may be set to `nil` to indicate that no highlighting should be done.  Thus functions that use `helm-buffers-sort-fn` or `helm-fuzzy-matching-highlight-fn` to sort or highlight a list of candidates must handle nil values for these options by using the candidates directly.

This change fixes https://github.com/emacs-helm/helm/issues/2343.

* `helm-buffers.el` (`helm-highlight-buffers`): Handle nil value for `helm-fuzzy-matching-highlight-fn`.
(`helm-buffers-sort-transformer`): Handle nil value for `helm-buffers-sort-fn`.
* `helm.el` (`helm-fuzzy-highlight-matches`): Handle nil value for `helm-fuzzy-matching-highlight-fn`.

### DIFF
--- a/helm-buffers.el
+++ b/helm-buffers.el
@@ -541,15 +541,20 @@ buffers)."
                                         (and helm-buffers-fuzzy-matching ""))))
                      (cons (if helm-buffer-details-flag
                                (concat
-                                (funcall helm-fuzzy-matching-highlight-fn
-                                         truncbuf)
+                                (if helm-fuzzy-matching-highlight-fn
+                                    (funcall helm-fuzzy-matching-highlight-fn
+                                             truncbuf)
+                                  truncbuf)
                                 helm-buffers-column-separator
                                 formatted-size
                                 helm-buffers-column-separator
                                 fmode
                                 helm-buffers-column-separator
                                 meta)
-                             (funcall helm-fuzzy-matching-highlight-fn name))
+                             (if helm-fuzzy-matching-highlight-fn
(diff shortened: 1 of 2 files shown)


thierryvolpiatto:
Miciah Dashiel Butler Masters <notifications@github.com> writes:

> ...or for helm-fuzzy-matching-highlight-fn. By default, helm-buffers-sort-fn is initialized with the value of helm-fuzzy-sort-fn, and the documentation string for
> helm-fuzzy-sort-fn states that it may be set to nil to indicate that no sorting should be done. Similarly, the documentation string for helm-fuzzy-matching-highlight-fn
> states that it may be set to nil to indicate that no highlighting should be done. Thus functions that use helm-buffers-sort-fn or helm-fuzzy-matching-highlight-fn to sort
> or highlight a list of candidates must handle nil values for these options by using the candidates directly.

No, we should not change this. The only issue here is a docstring issue,
none of these variables should be set to nil but to a function.

-- 
Thierry

Get my Gnupg key:

Miciah:
Understood.  Thanks!

Issue #2350, opened 2020-07-29
Customise how shell buffer is displayed


lionel-:
**Is your feature request related to a problem? Please describe.**
I always use shells in the right window of a 2-split configuration.

For some reason this used to work with `<f2>` in helm-find-files, whether a shell buffer was open or not. It stopped working after updating the Emacs packages, but I can't figure out why.

**Describe the solution you'd like**
`helm-ff-switch-to-shell` uses `switch-to-buffer` when the shell buffer already exists. This prevents it from being opened in a dedicated window.

I believe this patch would allow users to customise how shell buffers are opened:

```diff
--- a/helm-files.el
+++ b/helm-files.el
@@ -1562,7 +1562,9 @@ prefix arg shell buffer doesn't exists, create it and switch to it."
                        (helm-comp-read "Switch to shell buffer: " bufs
                                        :must-match t)
                      (car bufs)))
-        (switch-to-buffer it)
+        (if-let ((window (display-buffer it)))
+            (select-window window)
+          (switch-to-buffer it))
       (cl-case helm-ff-preferred-shell-mode
         (eshell-mode
          (eshell helm-current-prefix-arg))
```

**Describe alternatives you've considered**

Temporarily setting the current window as dedicated works for me because switch-to-buffer then delegates to pop-to-buffer, but that doesn't seem the right solution:

```elisp
(advice-add #'helm-ff-switch-to-shell :around #'my-around-dedicated-window)

(defun my-around-dedicated-window (fn &rest args)
  (let ((win (selected-window))
        (old (window-dedicated-p)))
    (unwind-protect
        (progn
          (set-window-dedicated-p win t)
          (apply fn args))
      (set-window-dedicated-p win old))))
```

thierryvolpiatto:
Lionel Henry <notifications@github.com> writes:

> Is your feature request related to a problem? Please describe.
> I always use shells in the right window of a 2-split configuration.

Ok, if you split in two vertical windows and you start helm from the
right window, your shell will be displayed in the right window.

> For some reason this used to work with <f2> in helm-find-files,

F2 never switch to shell, perhaps you have another configuration?

> whether a shell buffer was open or not. It stopped working after
> updating the Emacs packages, but I can't figure out why.

Please describe what was working and what is no more working.

From where do you start helm? right window? left window?
Where is the shell now once you switch to it?

Also please start from emacs-helm.sh so that we use same configuration,
same bindings etc...

[...]

PR #2352, opened 2020-07-30
Use `pop-to-buffer` instead of `display-buffer`

lionel-:
As discussed in https://github.com/emacs-helm/helm/issues/2350#issuecomment-666216567, sorry I wasn't more clear.


thierryvolpiatto:
Thanks! merged.

PR #2361, opened 2020-08-13
helm-eshell.el: Fix eshell-pcomplete keymap setting.

akreisher:
The `helm-esh-completion-map` is still being set in the construction of the helm instance, and was not being applied.

This PR moves the setting of the keymap from the `helm` call to the `helm-make-source` call in `helm-esh-pcomplete-default-source`.


thierryvolpiatto:
Thanks! Merged.

PR #2364, opened 2020-08-15
Check if symbol is bound before trying to get its value

### LINKED ISSUE #2363: minibuffer-local-must-match-filename-map: void-variable in emacs28
The recent emacs built from git removes obsolete
`minibuffer-local-must-match-filename-map` variable. 

Deleting line 31 from `helm-global-bindings.el`:
```
minibuffer-local-must-match-filename-map; Older emacsen
```

solves the problem.

d125q:
Fixes #2363.


thierryvolpiatto:
Thanks! merged.

Issue #2375, opened 2020-08-30
backtrace displayed using lsp-find-reference in native-comp emacs branch


gcla:
## Expected behavior

A helm-ified list of references to the programming language symbol under point.

## Actual behavior (from `emacs-helm.sh` if possible, see note at the bottom)

With debug-on-error set to t, I see a backtrace (shown in section below)

## Steps to reproduce (recipe)

- use emacs built from the native-comp branch. See https://akrl.sdf.org/gccemacs.html. "gccemacs is a modified Emacs capable of compiling and running Emacs Lisp as native code in form of re-loadable elf files. As the name suggests this is achieved blending together Emacs and the gcc infrastructure."
- after starting emacs for the first time, give it some minutes to natively-compile all your el/elc files (I checked with htop)
- confirm this function is natively-compiled:
```
(subr-native-elisp-p (symbol-function 'helm-xref-goto-xref-item))
```
- the problem may show with any lsp reference lookup, but in my particular case, I was working with Go, and used a hydra to invoke lsp-find-reference when point was placed over a Go variable

## Backtraces if any (`M-x toggle-debug-on-error`)

```
Debugger entered--Lisp error: (wrong-type-argument symbolp #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_8>)
  helm-symbol-name(#<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_8>)
  helm-source--header-line(#<helm-source-sync helm-source-sync-1582bda979a6>)
  #f(compiled-function (source) #<bytecode 0x19d62c8ed2c2da34>)(#<helm-source-sync helm-source-sync-1582bda979a6>)
  apply(#f(compiled-function (source) #<bytecode 0x19d62c8ed2c2da34>) #<helm-source-sync helm-source-sync-1582bda979a6>)
  #f(compiled-function (&rest args) #<bytecode -0x3feeee56fec87b4>)(#<helm-source-sync helm-source-sync-1582bda979a6>)
  apply(#f(compiled-function (&rest args) #<bytecode -0x3feeee56fec87b4>) #<helm-source-sync helm-source-sync-1582bda979a6> nil)
  helm--setup-source(#<helm-source-sync helm-source-sync-1582bda979a6>)
[...]

PR #2376, opened 2020-08-30
Fix issues running emacs built from the native-comp branch (#2375)

gcla:
The emacs native-comp branch provides "a modified Emacs capable of
compiling and running Emacs Lisp as native code in form of re-loadable
elf files" (https://akrl.sdf.org/gccemacs.html). Functions that are
natively-compiled do not return t to byte-code-function-p, but do return
t to subr-native-elisp-p, a function that is built-in to the native-comp
emacs.

The problem here is that helm-symbol-name tries to call symbol-name on
an object if it is not functionp OR byte-code-function-p; when running
with native-comp, this now needs to include subr-native-elisp-p
too. I've also modified some other functions that use a similar pattern.


thierryvolpiatto:
Thanks! Merged.

PR #2392, opened 2021-03-15
Handle persistent-action-if in helm-source--header-line

d125q:
Currently persistent-action-if is ignored in the default header line.
This PR includes it so that a source configured with

```elisp
:persistent-action-if 'foo
:action '(("bar" . bar)
          ("baz" . baz))
```

shows foo in the header line instead of bar.


thierryvolpiatto:
Thanks! merged.

PR #2398, opened 2021-04-13
* doc/helm-manual-1.org: Based on helm.wiki/Home.org

rnchzn:
* doc/helm-manual-1.org: Based on helm.wiki/Home.org

I have also added a TOC to the earlier manuals.  This  way the individual sections will be browsable on GitHub


thierryvolpiatto:
Thanks! merged.

PR #2403, opened 2021-04-20
* doc/helm.org: Use relative path

rnchzn:
* doc/helm.org: Use relative path
    * doc/Makefile: Fix errors with `make  clean-install-info'


thierryvolpiatto:
Thanks!

rnchzn:
That was pretty quick ... Thanks.

PR #2404, opened 2021-04-21
* doc/helm-manual.org (Pattern matching in helm-buffers): New

rnchzn:
* doc/helm-manual.org (Pattern matching in helm-buffers): New

I have re-written the buffer matching rules.


thierryvolpiatto:
Thanks merged!

PR #2408, opened 2021-04-25
* doc/helm-classes.org:  Prune redundant entries for slots.

rnchzn:
* doc/helm-classes.org:  Prune redundant entries for slots.

i.e., For child classes, document a slot only if it's  unique
to that class or it's init form is different from it's  parents.

* doc/ox-texinfo+.el: Add support for generating custom indices.
For eg, doc/helm-classes.org uses `hc' as an index for helm classes.

* Makefile: Load doc/ox-texinfo+.el


thierryvolpiatto:
Thanks, merged!

PR #2412, opened 2021-05-01
Ensure eshell-pcomplete directory is expanded relative to parent

akreisher:
In an Eshell buffer in a remote directory, completing with `helm-esh-pcomplete` works correctly with multiple candidates, as shown below:
screenshot-05-01-21-14:34:49

If the only valid completion is a single directory, however, the directory is repeated, and not immediately completed.
screenshot-05-01-21-14:35:17

`pcomplete` succesfully completes the candidate, so I figured there must be something happening in `helm-esh-get-candidates`. After tracing the code a bit, it seems like if a file candidate is a directory, it is being expanded relative to itself, causing the behavior above. Ensuring that we expand a file to its parent directory fixes this issue for me. I'm not sure why it was only a problem in remote directories with TRAMP, but this fix seems to work both locally and remote.


thierryvolpiatto:
Thanks! merged.

thierryvolpiatto:
Alexander Kreisher ***@***.***> writes:

> pcomplete succesfully completes the candidate, so I figured there must
> be something happening in helm-esh-get-candidates. After tracing the
> code a bit, it seems like if a file candidate is a directory, it is
> being expanded relative to itself, causing the behavior
> above. Ensuring that we expand a file to its parent directory fixes
> this issue for me.

Yes, your change is safer.


> I'm not sure why it was only a problem in remote directories with
> TRAMP, but this fix seems to work both locally and remote.

Don't know too, depend on what try-completion returns perhaps.

Thanks.

-- 
Thierry

PR #2418, opened 2021-05-17
Fix unexpected quit in terminal emacs

### LINKED ISSUE #2417: When using helm from terminal emacs, typing too quickly may result in helm unexpectedly closing
## Expected behavior
Helm should not quit when typing too quickly in terminal emacs, when using a package such as helm-dash, which generates candidates using an external process, but who's source does not say that it uses a process.

## Actual behavior (from `emacs-helm.sh` if possible, see note above)
Helm will unexpectedly quit after a bell is generated from typing too fast. Inhibit-quit is not enabled by `helm-get-candidates` because helm doesn't know the source uses an external process.

## Steps to reproduce (recipe)
Install `helm-dash`.
Run emacs in the terminal like `emacs -nw`.
Install a docset using `helm-dash-install-docset`.
Run `helm-dash` and type `playbook` really quickly, or mash keys.
At this stage the helm has probably quit unexpectedly with no error message.

in `*messages*`, you may see the following:
```
Waiting for process to die...(type C-g again to kill it instantly)
Waiting for process to die...done
```

## Describe versions of Helm, Emacs, operating system, etc.
- helm     20210320.1405
- GNU Emacs 28.0.50 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.20, cairo version 1.16.0) of 2021-03-23
- helm-dash 20190527.1118
- 
## Are you using `emacs-helm.sh` to reproduce this bug? (yes/no):
no

## Potential fix:
Change inhibit-quit to be permanently t, or do a check for terminal emacs frame.
```
(defun helm-get-candidates (source)
  "Retrieve and return the list of candidates from SOURCE."
  (let* ((candidate-fn (assoc-default 'candidates source))
         (candidate-proc (assoc-default 'candidates-process source))
         ;; See comment in helm-get-cached-candidates (Bug#2113).
         ;; (inhibit-quit candidate-proc)
         (inhibit-quit t)
...
```

mullikine:
Fixes #2417

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -4011,7 +4011,8 @@ WARNING: Do not use this mode yourself, it is internal to Helm."
   (let* ((candidate-fn (assoc-default 'candidates source))
          (candidate-proc (assoc-default 'candidates-process source))
          ;; See comment in helm-get-cached-candidates (Bug#2113).
-         (inhibit-quit candidate-proc)
+         (inhibit-quit (or candidate-proc
+                           (not (display-graphic-p))))
          cfn-error
          (notify-error
           (lambda (&optional e)


thierryvolpiatto:
This have been fixed by 6be278a77fbe0f9a2507608da9e0eb032dcb8a67 so i am closing.

Thanks.

PR #2427, opened 2021-06-30
Fix: quote an initform in defclass helm-source

akater:
Emacs 28 is going to get an update that will break code with unquoted symbols in initforms.  This seems to be the only one such in Helm.


thierryvolpiatto:
Thanks, merged.

PR #2429, opened 2021-07-03
Make visible mark prefix optional

akater:
Marks on margins shift the whole list of candidates in my case (console Emacs; most visible elements turned off), when they appear.  This is quite distracting.  Keeps happening even if I set `helm-visible-mark-prefix` to `""`.

I offer a quick solution but if you think the issue could be resolved without much hassle while keeping the type of `helm-visible-mark-prefix` as just `string`, I'll create an Issue.

### DIFF
--- a/helm.el
+++ b/helm.el
@@ -889,9 +889,10 @@ You can toggle later `truncate-lines' with
   :type 'boolean)
 
 (defcustom helm-visible-mark-prefix "*"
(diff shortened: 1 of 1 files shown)


akater:
Hmm, it doesn't help, even though it did yesterday.  Investigating.

akater:
No, it's just wrong.  I don't know how to fix this.

PR #2439, opened 2021-09-13
highlight filename extension when find files

taquangtrung:
Hi,

I created this PR to enable Helm to highlight the file name extensions of files when calling `helm-mini` for `helm-find-files` 
(by using the face `'helm-ff-file-extension`).

This is useful for users to quickly see what file types they want to look for.

- Here is a screenshot of the current display of file names (in recent files) when calling `helm-mini`:

helm-before

- This is how it looks like after applying my patch:

helm-after

Can you advise if this PR is useful and can be merged?

Thank you!


thierryvolpiatto:
Thanks! merged.

PR #2440, opened 2021-09-13
highlight file extension in buffer names

taquangtrung:
Hi,

I created another PR that is similar to https://github.com/emacs-helm/helm/pull/2439, but this new PR is to highlight the file name extension of opened buffers.

Here is a screenshot of how it looks like, on the first 3 buffers of the list.

helm-buffer

Please consider merging it too.

Thank you!

### DIFF
--- a/helm-buffers.el
+++ b/helm-buffers.el
@@ -431,9 +431,18 @@ The list is reordered with `helm-buffer-list-reorder-fn'."
   (append
    (list
     (concat prefix
-            (propertize buf-name 'face face1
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Thanks! I made a small modification to your previous changes, using helm-file-name-extension instead of file-name-extension, this to prevent highlighting extension with a number which are not meaningful mime types e.g. Changelog.1.

taquangtrung:
Thanks for the update! :-)

PR #2442, opened 2021-09-21
helm-buffer--format-mode-name: format-mode-line in buffer context

progfolio:
Fix an issue where elisp files with lexical binding enabled were
incorrectly displayed with a mode of Elisp/d in `helm-mini'.

`format-mode-line' defaults to the selected window and its current buffer.
Since `helm-buffer--format-mode-name' is called from the *helm mini* window,
the mode line is not formatted in the context of the buffer
(despite setting the current buffer via `with-current-buffer').


thierryvolpiatto:
Thanks! Merged.

PR #2444, opened 2021-10-08
Fix paren mismatch

obar:


thierryvolpiatto:
Thanks for the quick fix, but this commit had other problems and have been reverted.

PR #2445, opened 2021-10-08
Remove extra ")" on helm-files

zetahernandez:
Invalid read syntax: ")" on helm-files


thierryvolpiatto:
Already fixed, thanks.

PR #2456, opened 2021-11-08
helm-read-file-name: Fallback to $HOME when default-directory nil

progfolio:
Otherwise `helm-complete-file-name-at-point` fails in buffers which are not backed
by a file. e.g. \*scratch*.


progfolio:
We could let-bind `default-dir` directly in `helm-complete-file-name-at-point` directly if you think that's a better approach, but I figured `helm-read-file-name` should work in buffers without backing files as well.

thierryvolpiatto:
Merged thanks!

PR #2458, opened 2021-11-13
helm-misc.el (helm-minibuffer-history-old-key): Check existence

leungbk:
The latest Emacs 29 does not have
minibuffer-local-must-match-filename-map, so we must guard any attempt
at looking up its symbol-value.


thierryvolpiatto:
Thanks! Merged.

PR #2468, opened 2021-12-21
Extend helm-grep-ag-1 to accept input

jagrg:
For example: `(helm-grep-ag-1 dir '("-t txt") "pattern")`.

### DIFF
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -1639,8 +1639,9 @@ returns if available with current AG version."
                  (assoc-default 'follow helm-source-grep-ag))
       (setf (slot-value source 'follow) it)))
 
-(defun helm-grep-ag-1 (directory &optional type)
-  "Start helm ag in DIRECTORY maybe searching in files of type TYPE."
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Jonathan Gregory ***@***.***> writes:

>
> For example: (helm-grep-ag-1 dir '("-t txt") "pattern").

You don't need this, what if you don't like what is at point?
You will have to delete input to write something else instead, it is why
I generally don't provide INPUT in helm commands.
What you want to use is M-n, but if you have not modified default
setting of `helm-sources-using-default-as-input` you even do not need to
press M-n.

-- 
Thierry

jagrg:
This has nothing to do with thing-at-point. I want to pass pattern 
or helm-pattern to helm-grep-ag-1 from another function. This can 
be done with helm-do-grep-1:

(helm-do-grep-1 dir nil nil '("-t txt") nil "pattern")

but not with helm-grep-ag-1.

-- 
Jonathan

thierryvolpiatto:
Ok merged, thanks.

PR #2498, opened 2022-04-09
Update helm-this-command for emacs-28

0x60df:
Before emacs-28, `backtrace-frames` in `interactive` form looks like:
```
...
(descriptor of interactive form)
call-interactively(command)
command-execute(command)
...
```
In emacs-28, `backtrace-frame` of `call-interactively` does not exist.
This causes `helm-this-command` returns a wrong command.

I am not sure why `call-interactively` is missing between the `interactive` descriptor and `command-execute`, but anyway, we can obtain the command by the frame of `command-execute`.
Thus, `helm-this-command` is updated to check the `backtrace-frame` of `command-execute` and return the right command even while it is called in `interactive` form on emacs-28.

This update resolves problems on looking up `helm-mode` handler for the command invoked via `execute-extended-command` and other `M-x` commands as well as `helm-M-x`.

### DIFF
--- a/helm-lib.el
+++ b/helm-lib.el
@@ -425,6 +425,9 @@ Like `this-command' but return the real command, and not
            else
            if (and (eq fn 'call-interactively)
                    (> (length btf) 2))
+           return (cadr (cdr btf))
+           if (and (eq fn 'command-execute)
+                   (> (length btf) 2))
            return (cadr (cdr btf))))
 
 
--- a/helm-lib.el
+++ b/helm-lib.el
@@ -404,8 +404,7 @@ available APPEND is ignored."
   '(helm-maybe-exit-minibuffer
     helm-confirm-and-exit-minibuffer
     helm-exit-minibuffer
-    exit-minibuffer
-    helm-M-x))
+    exit-minibuffer))
 
 (defun helm-this-command ()
   "Return the actual command in action.


thierryvolpiatto:
Thanks, looks good.  Is there a recipe starting from M-x or helm-M-x to reproduce ?

0x60df:
Thank you for your review!

Here are the recipes.

For `M-x`, while using default `helm-completing-read-handlers-alist` (who has `(execute-extended-command . nil)`), invoke commands like the following example.

```el
(defun example (arg)
  (interactive (list (completing-read "Example: " '(candidate example list)))))
```

This results in using default `completing-read` instead of `helm-completing-read-default-handler`.

For `helm-M-x`, actually it works well and there are no issues, but for understanding, the same behavior could be reproduced by the following procedure. 

1. Add `(helm-M-x . nil-or-any-handler)` to `helm-completing-read-handlers-alist`.
2. Remove `helm-M-x` from `helm-this-command-black-list`.
3. Increase upper boundary for the count in `helm-this-command` (50 -> 200 in my trial).
4. Invoke commands like `example` above.

Then, specified handler for `helm-M-x` is used for the invoked commands like `example`.

thierryvolpiatto:
0x60DF ***@***.***> writes:

> Thank you for your review!
>
> Here are the recipes.
>
> For M-x, while using default helm-completing-read-handlers-alist (who
> has (execute-extended-command . nil)), invoke commands like the
> following example.
>
> (defun example (arg)
>   (interactive (list (completing-read "Example: " '(candidate example list)))))
>
> This results in using default completing-read instead of helm-completing-read-default-handler.

Of course, this is what expected.

> For helm-M-x, actually it works well and there are no issues, but for
> understanding, the same behavior could be reproduced by the following
> procedure.
>
>  1. Add (helm-M-x . nil-or-any-handler) to
>  helm-completing-read-handlers-alist.

Why one would add helm-M-x to helm-completing-read-handlers-alist?

>  
>  2. Remove helm-M-x from helm-this-command-black-list.  3. Increase
>  upper boundary for the count in helm-this-command (50 -> 200 in my
>  trial).  4. Invoke commands like example above.
>
> Then, specified handler for helm-M-x is used for the invoked commands like example.

You lost me here, I wonder how helm-M-x would be affected at all by an
entry in helm-completing-read-handlers-alist.

I am just wondering what you are trying to fix actually.

-- 
Thierry

0x60df:
Thank you for your comment!

I expect `execute-extended-command` read command with default `completing-read`, **AND** invoked command by `execute-extended-command` (`example` in the case above) read its argument with `helm-completing-read-default-handler`.
Before emacs-28, helm-mode works as I expected, but in emacs-28 both `execute-extended-command` and `example` does reading with default `completing-read`.

This is because `helm-this-command` in `(interactive (list (completing-read "Example: " '(candidate example list))))` returns `execute-extended-commnad` instead of `example` or `nil`.

Actually, I use `smex` for `M-x`, and the similar things happen.

For `helm-M-x`, there are no actual issues as you pointed out. The procedure described above is just for experiment and understanding behavior of forms.

thierryvolpiatto:
As I still don't understand what you want to fix, I am therefore closing this, feel free to open a bug report.
Thanks.

0x60df:
Thank you for your response, and sorry for my bad explanation...
Let me try again to describe what I want, but just for sharing information.

What I expected, and the behavior before emacs-28:
1. Call `smex`.
2. Select `example` in `smex`.
3. `arg` of `example` is asked with `helm` interface.
Emacs-27

On emacs-28:
1. Call `smex`.
2. Select `example` in `smex`.
3. `arg` of `example` is asked with default `completing-read` interface.
Emacs-28

It looks like caused by the different behavior of `helm-this-command` between emacs-28 and before.

thierryvolpiatto:
0x60DF ***@***.***> writes:

> Thank you for your response, and sorry for my bad explanation...
> Let me try again to describe what I want, but just for sharing information.
>
> What I expected, and the behavior before emacs-28:
>
>  1. Call smex.
>  2. Select example in smex.
>  3. arg of example is asked with helm interface.
>     Emacs-27
>
> On emacs-28:
>
>  1. Call smex.
>  2. Select example in smex.
>  3. arg of example is asked with default completing-read interface.
>     Emacs-28
>
> It looks like caused by the different behavior of helm-this-command between emacs-28 and before.

No, I couldn't reproduce, I downloaded and loaded smex.el, then M-x smex
=> example => and the helm-buffer is named "*helm-mode-example*" as
expected.

Thanks for the recipe.

> —


-- 
Thierry

0x60df:
Sorry, my fault, less checking...
This seems to occur only on the Emacs compiled with native compilation option.

Though there are few, let me share the workaround for `helm-mode` users who use `smex` or other similar `M-x` commands on Gcc Emacs.

```el
(add-to-list 'helm-this-command-black-list 'smex)
```

Thank you for reviewing!

thierryvolpiatto:
0x60DF ***@***.***> writes:

> Sorry, my fault, less checking...
> This seems to occur only on the Emacs compiled with native compilation option.
>
> Though there are few, let me share the workaround for helm-mode users who use smex or other similar M-x commands on Gcc Emacs.
>
> (add-to-list 'helm-this-command-black-list 'smex)

Probably using `command-execute` instead of `execute-extended-command` in
smex would fix the problem (not verified).


-- 
Thierry

0x60df:
Thank you for the comment!
I'll make some trial with `command-execute` and `execute-extended-command`.

thierryvolpiatto:
Also for other people reading this, modifying `helm-this-command-black-list` may break helm, this variable is not public, it is a defconst, so please don't modify it unless you know what you are doing.

PR #2500, opened 2022-04-15
Remove whitespace in helm-M-x

Horrih:
If show-trailing-whitespace is set to true globally, it leads to an ugly
completion buffer.


thierryvolpiatto:
Merged thanks!

PR #2509, opened 2022-05-17
More precise idle-for in helm-elisp--format-timer

renatofdds:
Better format and consider milliseconds for idle timers.


thierryvolpiatto:
Thanks, merged.

PR #2515, opened 2022-05-26
Add insert face action

jagrg:
This is useful when editing a theme. For example, `M-x helm-apropos RET ^dired`, then move to the Faces source, press `M-a` to select all and f4 to insert them.

### DIFF
--- a/helm-elisp.el
+++ b/helm-elisp.el
@@ -616,7 +616,9 @@ double quote."
     :action '(("Describe face" . helm-describe-face)
               ("Find face" . helm-find-face-definition)
               ("Customize face" . (lambda (candidate)
-                                    (customize-face (helm-symbolify candidate)))))))
+                                    (customize-face (helm-symbolify candidate))))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
You have C-c C-i for this (for any source in helm) and the helm-colors source (M-x helm-colors) have more specific commands for this).
Thanks.
PS: Answered long time ago to this by mail but it seems during sometimes mail response was not ending here)

jagrg:
Thanks, I wasn't aware of C-c C-i/TAB, and I see it in the 
helm-help buffer. Sorry about that.

thierryvolpiatto:
Jonathan Gregory ***@***.***> writes:

> This is useful when editing a theme. For example, M-x helm-apropos RET
> ^dired, then move to the Faces source, press M-a to select all and f4
> to insert them.

You already have `C-c C-i` for this, isn't it the same?

-- 
Thierry

PR #2523, opened 2022-06-16
Fix issue of "Invalid function: (helm-apropos-default-sort-fn)"

leuven65:
The issue pops up on the command "helm-apropos"


thierryvolpiatto:
Oops! yes thanks you, merged.

PR #2530, opened 2022-07-11
Fix: `completing-read` does not pick the default item when `helm-patt…

twlz0ne:
…ern` is not empty

* helm-core.el(helm-initial-setup): set `helm-pattern` before
`helm-compute-attr-in-sources` to prevent it from interfering the
`completing-read` pick the default item.

Steps to reproduce:

1. Evaluate following code:

```elisp
(progn
  (defun test-helm-completing-read ()
    (interactive)
    (completing-read "> " '("1" "2" "3") nil t nil nil "3"))

  ;; Simulate `M-x test-helm-completing-read RET'
  (minibuffer-with-setup-hook
      (lambda ()
        (let ((hook (make-symbol "hook")))
          (fset hook (lambda ()
                       (remove-hook 'post-command-hook hook)
                       (exit-minibuffer)))
          (add-hook 'post-command-hook hook)))
    (helm-M-x-read-extended-command obarray nil '("test-helm-completing-read"))))
```

2. Check the default selected item (expected "3", actual "1").

Another simpler way to reproduce:

```elisp
(let ((helm-pattern "not empty"))
  (completing-read "> " '("1" "2" "3") nil t nil nil "3"))
;; => 1
;;    2
;;    3

(let ((helm-pattern ""))
  (completing-read "> " '("1" "2" "3") nil t nil nil "3"))
;; => 3
;;    1
;;    2
```

### DIFF
--- a/helm-core.el
+++ b/helm-core.el
@@ -3611,8 +3611,6 @@ For RESUME INPUT DEFAULT and SOURCES see `helm'."
               'vertical 'horizontal))
     (setq helm--window-side-state
           (or helm-split-window-default-side 'below)))
-  ;; Call the init function for sources where appropriate
-  (helm-compute-attr-in-sources 'init sources)
   (setq helm-pattern (or (and helm-maybe-use-default-as-input
                               (or (if (listp default)
                                       (car default) default)
@@ -3620,6 +3618,8 @@ For RESUME INPUT DEFAULT and SOURCES see `helm'."
                                     (thing-at-point 'symbol))))
                          ""))
   (setq helm-input "")
+  ;; Call the init function for sources where appropriate
+  (helm-compute-attr-in-sources 'init sources)
   (clrhash helm-candidate-cache)
   (helm-create-helm-buffer)
   (helm-clear-visible-mark)


twlz0ne:
Still can not set the default item sometimes, e.g.:

``` emacs-lisp
(let ((helm-completion-style '(helm-flex)))
  (completing-read "> " '("1" "2" "3") nil t nil nil "3"))
;; => 1
;;    2
;;    3
```

According to the following code, the `:default` will be set as `nil` if  the`helm-completion-style` not equals `emacs`:

https://github.com/emacs-helm/helm/blob/7c2f08063910ef15a63cd58e6db0ffa2e4ddc12d/helm-mode.el#L1074-L1076

But the default in candidates seems not working.

thierryvolpiatto:
twlz0ne ***@***.***> writes:

> Still can not set the default item sometimes, e.g.:
>
> (let ((helm-completion-style '(helm-flex)))

You are confused with `helm-completion-style` and `completion-styles`.
See docstring of `helm-completion-style`.
For `completion-styles`, it has effect only if
helm-completion-style==emacs, for its value use only a list, note that
'(helm-flex) as value is only for emacs-26, with recentest emacs(27+)
use '(flex).

So this is working as you expect:

    (let ((helm-completion-style 'helm-fuzzy))
      (completing-read "> " '("1" "2" "3") nil t nil nil "3")) 

and this as well:

    (let ((helm-completion-style 'emacs)
          (completion-styles '(flex)))
      (completing-read "> " '("1" "2" "3") nil t nil nil "3"))

Note also that when setting `completion-styles`, you don't have to add
`helm` style, it is added by helm automatically, so when specifying
'(flex), helm will fallback to helm style (multimatching) as soon as you
enter a space.

HTH.

-- 
Thierry

twlz0ne:
> You are confused with `helm-completion-style` and `completion-styles`.

@thierryvolpiatto Thanks for correcting.

But the following behavior (I reported at the beginning https://github.com/emacs-helm/helm/pull/2530#issue-1300087519) should be an issue:

``` emacs-lisp
(let ((helm-pattern "not empty"))
  (completing-read "> " '("1" "2" "3") nil t nil nil "3"))
;; => 1
;;    2
;;    3
```

thierryvolpiatto:
twlz0ne ***@***.***> writes:

>     You are confused with helm-completion-style and completion-styles.
>    
> @thierryvolpiatto Thanks for correcting.
>
> But the following behavior (I reported at the beginning #2530 (comment)) should be an issue:
>
> (let ((helm-pattern "not empty"))
>   (completing-read "> " '("1" "2" "3") nil t nil nil "3"))
> ;; => 1
> ;;    2
> ;;    3

Why one would do this?

> —


-- 
Thierry

twlz0ne:
> Why one would do this?

Of course no one writes code just to read 1 2 3, that is an example of reproducing the issue.

Consider that I defined the following function:

``` emacs-lisp
(defun lsp-XXX-install-server (node-ver)
  "Install XXX language Server for node NODE-VER."
  (interactive
   (list (completing-read
          "Node version: "
          (cdr (shim-versions 'node))
          nil t nil nil
          ;; Set default (current "17.9.1")
          (shim-version 'node))))
  (npm-install "XXX-language-server" node-ver))
```

Then when I execute it via `M-x lsp-typescript-install RET` (`M-x` binded to `helm-M-x`), I expect the default item to be selected, but it is not:

``` emacs-lisp
;; Actaul:              Expected:
;;
;; => 6.14.3            => 17.9.1
;;    8.11.3               6.14.3
;;    10.9.0               8.11.3
;;    11.15.0              10.9.0
;;    12.20.0              11.15.0
;;    16.15.1              12.20.0
;;    17.9.1               16.15.1
```

(Note: `M-: (call-interactively 'lsp-XXX-install-server) RET` is as expected.)

The causes for this issue is that the `helm-pattern` becomes **non-empty** after `M-x`, which is equivalent to:

``` emacs-lisp
(let ((helm-pattern "whatever"))
  (completion-read ...))
```

thierryvolpiatto:
twlz0ne ***@***.***> writes:

>     Why one would do this?
>    
> Of course no one writes code just to read 1 2 3, that is an example of reproducing the issue.

Thanks, I finally understand what you mean (took more than one month),
you should have started by this.

I pushed a simple fix that reset pattern on exit/quit.

I suspect there is other use cases other than calling init fns before
resetting pattern that could trigger this bug, I have no recipes though,
but I guess resetting pattern when quitting/exiting helm should cover all cases.

-- 
Thierry

tmalsburg:
The change in ae8721ca331a10a3ce3a8d86570108f11709381b breaks multiple things in helm-mu, which relies on 
`helm-pattern` storing the input for the last search.  We can work around this in helm-mu, but I wanted to make you aware of this because other packages may also rely on the old behavior.

thierryvolpiatto:
Hello Titus,

Titus von der Malsburg ***@***.***> writes:

> The change in ae8721c breaks multiple things in helm-mu, which relies on
> helm-pattern storing the input for the last search.

What is breaking exactly?

> We can work around this in helm-mu, but I wanted to make you aware of
> this because other packages may also rely on the old behavior.

Thanks, but I think one should not relay on helm-pattern as its previous
value is not always related to the helm command you use.

e.g. With M-x helm-mu next helm-mu session would have helm-mu as initial
input (with old behavior) which is probably not what you want.

I guess what you want is to reuse your last search as default for your
next helm-mu session isn't it ?

> —


-- 
Thierry

tmalsburg:
> What is breaking exactly?

The action for opening an e-mail (`helm-mu-display-email` uses the current `helm-pattern` to instruct mu4e to display the e-mail. Mu4e then opens a headers view for the help pattern and then opens the e-mail (which appears in that headers view).  Background is that Mu4e apparently can only display e-mail messages that are listed in the current headers view.  If `helm-pattern` is reset, the headers view does not correspond to the current search.

I think another case is the function `helm-mu-open-headers` view, which opens the current helm-mu search in Mu4e.  It uses `helm-run-after-quit` but at that time the `helm-pattern` is gone.

There may be a third case in `helm-mu-init` (which calls `helm-mu-get-search-pattern` which itself uses `helm-pattern`), but I haven't thought this case through and it may not be an issue there.

> Thanks, but I think one should not relay on helm-pattern as its previous
> value is not always related to the helm command you use.
> e.g. With M-x helm-mu next helm-mu session would have helm-mu as initial
> input (with old behavior) which is probably not what you want.

I didn't have any problems yet using `helm-pattern`, but it's good to know that its use is not advised. I guess helm-mu itself has to keep track of search terms or is there an easier solution using helm infrastructure?

tmalsburg:
> I guess what you want is to reuse your last search as default for your
> next helm-mu session isn't it ?

Not exactly that but something similar: If helm-mu is called from an e-mail buffer or headers view, the default helm-mu search term is taken from the current headers view. Very useful feature. Can be used to refine search terms in existing headers views.

thierryvolpiatto:
Hello Titus,
I made several changes to helm-mu, have a look.

Thanks.

-- 
Thierry

tmalsburg:
Thank you, Thierry. Everything seems to work as intended now.

thierryvolpiatto:
twlz0ne ***@***.***> writes:

> (let ((helm-pattern "not empty"))
>   (completing-read "> " '("1" "2" "3") nil t nil nil "3"))
> ;; => 1
> ;;    2
> ;;    3
>
> (let ((helm-pattern ""))
>   (completing-read "> " '("1" "2" "3") nil t nil nil "3"))
> ;; => 3
> ;;    1
> ;;    2

Ok, but why helm-pattern would be not empty unless you specify
INITIAL-INPUT?

helm-pattern is an internal var which should not be modified by the end
user of completing-read.

What do you want to fix exactly?

-- 
Thierry

thierryvolpiatto:
Titus von der Malsburg ***@***.***> writes:

>     I guess what you want is to reuse your last search as default for your
>     next helm-mu session isn't it ?
>    
> Not exactly that but something similar: If helm-mu is called from an e-mail buffer or headers view, the default helm-mu search term is taken from the current headers view. Very useful
> feature. Can be used to refine search terms in existing headers views.

IIUC you use helm-pattern in an action called from helm-mu, is this
right?
If so is using helm-local-input instead of helm-pattern working for you?

> —


-- 
Thierry

tmalsburg:
Hi Thierry, my problem is solved. I ended up using "helm-last-query". Thank you!

PR #2540, opened 2022-08-26
Let dabbrev complete function names from PHP function calls

cweiske:
Function calls in PHP are chained with "->":

    $this->someFunc()->someOtherFunc()->varname

helm-dabbrev does not complete such function names.
When typing "someF" and invoking helm-dabbrev, nothing gets completed.

The solution is to add ">" to the regex definining the
start of an expansion.

I hope this does not break other languages.


thierryvolpiatto:
Thanks, merged.

PR #2541, opened 2022-08-28
Set "helm-pattern" to "input" to avoid the source conetent refreshed by "helm-resume"

leuven65:
When using "helm-resume" to resume previous "helm-do-grep-ag", the grep content is always updated (re-executed) and previous selection is lost.
The problem is caused by that the "helm-pattern" is not reset to previous "input", and it causes "helm-check-new-input" check that it need to update the helm buffer.


thierryvolpiatto:
Good catch, merged.

Thanks.

Issue #2544, opened 2022-09-05
Modification of helm-file broke it for Emacs 27


LevPetrovitch:
### What happened?

Last commit in tramp-file [https://github.com/emacs-helm/helm/commit/5496dbf263e6d91c6360bb4a278710ab9479810d](link) 
broke helm for Emacs 27. I get a `(void-function tramp-file-name-unify)` error when loading helm.
This is due to the fact that Emacs 27, the version installed by Ubuntu, still uses tramp version 2.4.3, which doesn't define `tramp-file-name-unify`

### How to reproduce?

Load Helm in an Emacs 27.1

### Helm Version

Melpa or NonGnuElpa

### Emacs Version

Emacs-27.1

### OS

GNU/Linux

### Relevant backtrace (if possible)

```text
Debugger entered--Lisp error: (void-function tramp-file-name-unify)
  func-arity(tramp-file-name-unify)
  byte-code("\300\301!\302\232\203\n\0\303\207\304\207" [func-arity tramp-file-name-unify (1 . 2) helm-ff--file-accessible-directory-p file-accessible-directory-p] 2)
  (defvar helm-ff--file-accessible-directory-p-fn (byte-code "\300\301!\302\232\203\n\0\303\207\304\207" [func-arity tramp-file-name-unify (1 . 2) helm-ff--file-accessible-directory-p file-accessible-directory-p] 2))
  require(helm-files)
  eval-buffer(#<buffer  *load*> nil "/home/hobbit/.emacs.d/elpa/helm-20220905.430/helm-..." nil t)  ; Reading at buffer position 863
  load-with-code-conversion("/home/hobbit/.emacs.d/elpa/helm-20220905.430/helm-..." "/home/hobbit/.emacs.d/elpa/helm-20220905.430/helm-..." nil t)
  (helm-mode 1)
[...]

PR #2545, opened 2022-09-05
do not try using tramp-file-name-unify if isn't there

LevPetrovitch:
Solves issue #2544 by checking if `tramp-file-name-unify` is defined as a function before trying to use it.


thierryvolpiatto:
Thanks. Merged.

PR #2551, opened 2022-10-01
Fix helm-help-internal

renatofdds:


thierryvolpiatto:
Thanks to catch this, Merged!

PR #2560, opened 2022-10-22
Implemented default mode selection for buffer substring 

berdanakyurek:
Added a new custom called "Helm Occur Buffer Substring Default Mode" which can get ‘buffer-substring’ or ‘buffer-substring-no-properties’ values. The major modes that do not appear in the "Helm Occur Buffer Substring Fn For Modes" list get this value instead of 'buffer-substring-no-properties'. This way, it is possible to customize default behavior. In the current version, it is necessary to add all major modes to the list one by one. For the people who do not use "buffer-substring-no-properties" much, it would be easier to change the default to "buffer-substring" and add it to the list when necessary.


thierryvolpiatto:
Thanks, merged.

PR #2561, opened 2022-10-24
Add helm-fd-mode-line-function customization

lerouxrgd:
Add the `helm-fd-mode-line-function` customization with default `helm-fd-default-mode-line` which is the same behavior as before.

This is inspired by what `helm-ag` does with `helm-ag-show-status-function`.

It is also useful for `doom-modeline` as there is no other way to customize the mode-line with `helm-fd`.
See [seagle0128/doom-modeline#584](https://github.com/seagle0128/doom-modeline/issues/584)


thierryvolpiatto:
Merged, thanks!

PR #2583, opened 2023-01-11
Fix helm-define-key-with-subkeys error on keys with meta

mosquito-magnet:
char-to-string fails on key combinations including the meta/alt key, e.g. C-M-u, use single-key-description instead.

Sample for reproduction:
(helm-define-key-with-subkeys helm-map
      (kbd "C-w") ?\C-w 'helm-yank-text-at-point '((?\C-\M-u . helm-undo-yank-text-at-point)))


thierryvolpiatto:
Merged, thanks.  However note that the purpose of subkeys is generally to bind short keys (like self-insert-command) e.g. "n", "p" etc...

PR #2598, opened 2023-04-24
Add a command helm-do-grep-ag-project-root

huangfeiyu:
unlike helm-do-grep-ag which search the pattern default in directory of current buffer, helm-do-grep-ag-project-root search the given pattern in project root determined by emacs build-in project.el.

### DIFF
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -1787,6 +1787,18 @@ version."
   (helm-grep-ag (expand-file-name default-directory) arg))
 
 ;;;###autoload
+(defun helm-do-grep-ag-project-root (arg)
+  "Preconfigured `helm' for grepping with AG in `project-root'.
+With prefix arg prompt for type if available with your AG
+version."
+  (interactive "P")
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
huangfeiyu ***@***.***> writes:

> unlike helm-do-grep-ag which search the pattern default in directory
> of current buffer, helm-do-grep-ag-project-root search the given
> pattern in project root determined by emacs build-in project.el.

We don't need this, you can use helm-find-files 'C-l' (n
times) and 'M-g a' to do the same, thus project.el is not available in
older emacs, anyway I don't want dependency on project.el.

Thanks.

-- 
Thierry

huangfeiyu:
@thierryvolpiatto thanks for mention helm-find-files and 'M-g a', I still think that it is necessary to have a specific command which search pattern default in project's root directory, which is used very common.
Out of curiosity, may I know is there any other reason which makes you don't want to depend on project.el besides it is not available in older emacs you mentioned?

PR #2603, opened 2023-05-28
show all blocks if necessary when go to char

taquangtrung:
Hi,

Currently, `helm-goto-char` only show the current hidden block (when `hideshow` mode is enabled), which might make it incorrectly jump to a location, if more blocks are hidden.

I fix this issue by calling `hs-show-all` instead.

Can you check and consider merging this PR?

Thanks!


thierryvolpiatto:
Thanks, merged.

PR #2611, opened 2023-08-10
fix the initial input content of minibuffer for helm-occur command

yuchunzhou:


yuchunzhou:
Hello @thierryvolpiatto this PR will initialize the minibuffer with the value of `(thing-at-point 'symbol)` if there is no default input, please review it.

thierryvolpiatto:
Hello yuchunzhou,

yuchunzhou ***@***.***> writes:

> Hello @thierryvolpiatto this PR will initialize the minibuffer with
> the value of (thing-at-point 'symbol) if there is no default input,
> please review it.

(thing-at-point 'symbol) is always provided as default input with M-n,
thus when the corresponding source is in
`helm-sources-using-default-as-input` helm-occur should start
immediately with (thing-at-point 'symbol) even if the prompt is still empty
waiting eventually something new as input without having to delete
initial-input.

This is why we do not use initial-input in helm-occur, helm-grep etc...

Thanks.

> —


-- 
Thierry

yuchunzhou:
> 

Okay, @thierryvolpiatto , i get you, but why `helm-projectile-find-file` command has the initial input at minibuffer? i think it would be better if `helm-occur` can display the default input.

thierryvolpiatto:
yuchunzhou ***@***.***> writes:

>
> Okay, @thierryvolpiatto , i get you, but why helm-projectile-find-file
> command has the initial input at minibuffer?

IIRC helm-projectile-find-file uses the same mechanism as
helm-find-files and yes we need some input here (for many reasons I will
not explain here).

> i think it would be better if helm-occur can display the default
> input.

No, this have been discussed several times and for such searching
functions I don't want an initial-input for the reasons explained in
previous post.

> —


-- 
Thierry

yuchunzhou:
> yuchunzhou ***@***.***> writes:
> 1. ( ) text/plain (*) text/html Okay, @thierryvolpiatto , i get you, but why helm-projectile-find-file command has the initial input at minibuffer?
> IIRC helm-projectile-find-file uses the same mechanism as helm-find-files and yes we need some input here (for many reasons I will not explain here).
> i think it would be better if helm-occur can display the default input.
> No, this have been discussed several times and for such searching functions I don't want an initial-input for the reasons explained in previous post.
> […](#)
> — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you were mentioned.*Message ID: ***@***.***>
> -- Thierry

okay, i will look at your discussions.

PR #2631, opened 2023-12-06
Fix helm-completion--initial-filter for candidates with no affixation

renatofdds:
For cases such as [helm-completion-library-affixation](https://github.com/emacs-helm/helm/blob/e9cc7a80d0c01d6714400e662136733af7ddeee3/helm-mode.el#L1302) which can return nil.


thierryvolpiatto:
Thanks! Merged.

PR #2640, opened 2024-03-04
Remove 'resize-mini-windows' local binding that causes UI bug.

karbiv:
It prevented minibuffer from adjusting its height for 'help-key-binding' face in 'helm-hide-minibuffer-maybe' function.


thierryvolpiatto:
Ok, merged thanks.

Issue #2643, opened 2024-03-10
follow-mode messes up Emacs' buffer-list, pseudo-randomized order of buffers on each Helm invocation.


karbiv:
### What happened?

Helm's persistent action functions use `switch-to-buffer` and each follow action puts current selected buffer to top of `buffer-list`. So on each invocation of `helm-mini` buffers are shown in some random order depending on how they were viewed in previous `helm-mini` session with `follow-mode` on.

### How to reproduce?

The usual `helm-mini` buffer with follow-mode on. It's more obvious when more buffers are opened, 5-7.

### Helm Version

Master branch

### Emacs Version

Emacs-29.1

### OS

GNU/Linux

### Minimal configuration

- [X] I agree using a minimal configuration

thierryvolpiatto:
Alexander Karbivnichiy ***@***.***> writes:

> What happened?
>
> Helm's persistent action functions use switch-to-buffer and each
> follow action puts current selected buffer to top of buffer-list. So
> on each invocation of helm-mini buffers are show in some random order
> depending on how they were viewed in previous helm-mini session with
> follow-mode on.

Ah yes, never think at this as I don't use follow in buffers source.

Thanks.

> How to reproduce?
>
> The usual helm-mini buffer with follow-mode on. It's more obvious when more buffers are opened, 5-7.
>
> Helm Version
>
> Master branch
>
> Emacs Version
>
> Emacs-29.1
>
> OS
>
> GNU/Linux
>
> Relevant backtrace (if possible)
>
> No response
>
> Minimal configuration
>
>   • [*] I agree using a minimal configuration
>
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> You are receiving this because you are subscribed to this thread.*Message ID: ***@***.***>

-- 
Thierry

PR #2644, opened 2024-03-10
'norecord argument and display-buffer instead of switch-to buffer in …

karbiv:
…helm-buffers-list-persistent-action

https://github.com/emacs-helm/helm/issues/2643


thierryvolpiatto:
Thanks, merged.

PR #2647, opened 2024-03-20
suppress dired-async-processes being void

kkatsuyuki:
When you do `helm-ff-compress-marked-files` before loading dired-async.el, the error of `dired-async-processes` being void happens.


thierryvolpiatto:
Yes, good catch.   Thanks, Merged.

Issue #2652, opened 2024-03-24
Endless loop in 'helm-fuzzy-default-highlight-match-1'


karbiv:
### What happened?

When typing in helm minibuffer a pattern like this: "**", Helm enters endless loop and consumes 100% CPU.

It could happen on any other broken regex pattern. The problem is in `re-search-forward` that with some regexes doesn't move point forward and if put inside a `while` loop will repeat forever.
And that's normal behavior for `re-search-forward` because the function is supposed to work with a regex argument created and tested by a programmer, not from a user input.

There's a PR with a fix.

### How to reproduce?

Input "**" in `helm-mini` when _recentf_ source is enabled.

### Helm Version

Master branch

### Emacs Version

Emacs-29.1

### OS

GNU/Linux

### Minimal configuration

- [X] I agree using a minimal configuration

karbiv:
It was a reason of other bugs, and I guess that's probably why `helm-update-blacklist-regexps` was created.
https://github.com/emacs-helm/helm/pull/2653

karbiv:
Fixed second `re-search-forward` in `helm-fuzzy-default-highlight-match-1`.
It entered endless loop on input "^ ". (caret and space)

karbiv:
> helm-update-blacklist-regexps doesn't affect highlighting matches.

I'll get to it maybe, I was exploring why Helm buffer is erased by some one character patterns like "!", "^".

PR #2653, opened 2024-03-24
Fix endless loop in 'helm-fuzzy-default-highlight-match-1'.

karbiv:
https://github.com/emacs-helm/helm/issues/2652

### DIFF
--- a/helm-core.el
+++ b/helm-core.el
@@ -4696,10 +4696,14 @@ useful when the order of the candidates is meaningful, e.g. with
             (progn
               ;; Try first matching against whole pattern.
               (unless (string= pattern "")
-                (while (re-search-forward regex nil t)
-                  (cl-incf count)
-                  (helm-add-face-text-properties
-                   (match-beginning 0) (match-end 0) 'helm-match)))
+                (cl-loop
+                 while (re-search-forward regex nil t)
+                 when (eql (match-beginning 0) (match-end 0))
+                 do (cl-return)
+                 do
+                 (cl-incf count)
+                 (helm-add-face-text-properties
+                  (match-beginning 0) (match-end 0) 'helm-match)))
               ;; If no matches start matching against multiples or fuzzy matches.
               (when (zerop count)
                 (cl-loop with multi-match = (string-match-p " " pattern)
--- a/helm-core.el
+++ b/helm-core.el
@@ -4720,10 +4720,13 @@ useful when the order of the candidates is meaningful, e.g. with
                          ;; Multi matches (regexps patterns).
                          if multi-match do
                          (progn
-                           (while (re-search-forward p nil t)
-                             (helm-add-face-text-properties
-                              (match-beginning 0) (match-end 0)
-                              'helm-match))
+                           (cl-loop
+                            while (re-search-forward p nil t)
+                            when (eql (match-beginning 0) (match-end 0))
+                            do (cl-return)
+                            do
+                            (helm-add-face-text-properties
+                             (match-beginning 0) (match-end 0) 'helm-match))
                            (goto-char (point-min)))
                          ;; Fuzzy matches (literal patterns).
                          else do


karbiv:
There are many other places in Helm where `re-search-forward` is used, and possibly can be affected by the bug.
But the PR fixes only one place. There's other usage of `re-search-forward` in the same func, didn't hit that case in debugging yet.

thierryvolpiatto:
Here an alternative patch avoiding calls to match-* fns:

```diff
diff --git a/helm-core.el b/helm-core.el
index 9b672d77..474eebc2 100644
--- a/helm-core.el
+++ b/helm-core.el
@@ -4696,10 +4696,14 @@ useful when the order of the candidates is meaningful, e.g. with
             (progn
               ;; Try first matching against whole pattern.
               (unless (string= pattern "")
-                (while (re-search-forward regex nil t)
-                  (cl-incf count)
-                  (helm-add-face-text-properties
-                   (match-beginning 0) (match-end 0) 'helm-match)))
+                (let ((pos (point)))
+                  (helm-awhile (re-search-forward regex nil t)
+                    (if (eql it pos)
+                        (cl-return)
+                      (setq pos it)
+                      (cl-incf count)
+                      (helm-add-face-text-properties
+                       (match-beginning 0) (match-end 0) 'helm-match)))))
               ;; If no matches start matching against multiples or fuzzy matches.
               (when (zerop count)
                 (cl-loop with multi-match = (string-match-p " " pattern)
@@ -4715,11 +4719,14 @@ useful when the order of the candidates is meaningful, e.g. with
                          for p in patterns
                          ;; Multi matches (regexps patterns).
                          if multi-match do
-                         (progn
-                           (while (re-search-forward p nil t)
-                             (helm-add-face-text-properties
-                              (match-beginning 0) (match-end 0)
-                              'helm-match))
+                         (let ((pos (point)))
+                           (helm-awhile (re-search-forward p nil t)
+                             (if (eql it pos)
+                                 (cl-return)
+                               (setq pos it)
+                               (helm-add-face-text-properties
+                                (match-beginning 0) (match-end 0)
+                                'helm-match)))
                            (goto-char (point-min)))
                          ;; Fuzzy matches (literal patterns).
                          else do


```

thierryvolpiatto:
I merged your PR, but I will commit on top of it a change which allow using your fix in other places in Helm.
Thanks to catch this.

karbiv:
Still checking those macros version, whatever makes it better, if it does the work exiting the loop.
I think checking match data is more robust, not relying on return of `re-search-forward`.

thierryvolpiatto:
match data is anyway related to what re-search-forward does, isn't it?
Let me know if commit 8068f530 fix #2652 as well.
Thanks.

karbiv:
> match data is anyway related to what re-search-forward does, isn't it?
> Let me know if commit [8068f53](https://github.com/emacs-helm/helm/commit/8068f5301cd1a8495dee6882de99ffec6f89f55f) fix #2652 as well.

Yes it fixes it. I just didn't notice at first that `re-search-forward` returns point and made my fix to check match data directly.

thierryvolpiatto:

>
>     match data is anyway related to what re-search-forward does, isn't it?
>     Let me know if commit 8068f53 fix #2652 as well.
>    
> Yes it fixes it.

Great, thanks to confirm.
Should prevent hanging as well when searching candidates-buffer.  If we
find other places where re-search-forward may hang we can now use
helm-re-search-forward instead.

> I just didn't notice at first that re-search-forward returns point and
> made my fix to check match data directly.

Your fix were good as well, two different approches.  But I guess we
should not have to do this, in a better world (re-search-forward "**")
should return an error (nil when NOERROR is specified), actually it
behave like (re-search-forward "") which is wrong IMO, not sure if it is
a bug or a feature, it seems string-match does the same:

(let ((str "foo")) (when (string-match "**" str) (match-string 0 str)))
=> ""

so I guess it is how the emacs regexp engine behave (not documented).

> —

-- 
Thierry

michael-heerdegen:
Thierry Volpiatto ***@***.***> writes:


> Your fix were good as well, two different approches. But I guess we 
> should not have to do this, in a better world (re-search-forward "**") 
> should return an error (nil when NOERROR is specified), actually it 
> behave like (re-search-forward "") which is wrong IMO, not sure if it is 
> a bug or a feature, it seems string-match does the same: 
>
> (let ((str "foo")) (when (string-match "**" str) (match-string 0 str))) 
> => "" 
>
> so I guess it is how the emacs regexp engine behave (not documented). 

It is: (info "(elisp) Regexp Special"):

   For historical compatibility, a repetition operator is treated as
ordinary if it appears at the start of a regular expression or after
‘^’, ‘\`’, ‘\(’, ‘\(?:’ or ‘\|’.  For example, ‘*foo’ is treated as
‘\*foo’, and ‘two\|^\{2\}’ is treated as ‘two\|^{2}’.  It is poor
practice to depend on this behavior; use proper backslash escaping
anyway, regardless of where the repetition operator appears.

Btw, also `xr' knows about this:

  (xr "**") ==> (zero-or-more "*")

thierryvolpiatto:
Michael Heerdegen ***@***.***> writes:

> Thierry Volpiatto ***@***.***> writes:
>
>> Your fix were good as well, two different approches. But I guess we
>> should not have to do this, in a better world (re-search-forward "**")
>> should return an error (nil when NOERROR is specified), actually it
>> behave like (re-search-forward "") which is wrong IMO, not sure if it is
>> a bug or a feature, it seems string-match does the same:
>>
>> (let ((str "foo")) (when (string-match "**" str) (match-string 0 str)))
>> => ""
>>
>> so I guess it is how the emacs regexp engine behave (not documented).
>
> It is: (info "(elisp) Regexp Special"):

Ah! Thanks, didn't find it.

> For historical compatibility, a repetition operator is treated as
> ordinary

What means ordinary in this case?

> if it appears at the start of a regular expression or after ‘^’, ‘\`’,
> ‘\(’, ‘\(?:’ or ‘\|’. For example, ‘*foo’ is treated as ‘\*foo’, and
> ‘two\|^\{2\}’ is treated as ‘two\|^{2}’. It is poor practice to depend
> on this behavior; use proper backslash escaping anyway, regardless of
> where the repetition operator appears.
>
> Btw, also `xr' knows about this:
>
> (xr "**") ==> (zero-or-more "*")

Do you mean `rx`?

> —

-- 
Thierry

michael-heerdegen:
Michael Heerdegen ***@***.***> writes:

> (xr "**") ==> (zero-or-more "*")

What I wanted to say is that even `xr' knows that the first * in the
regexp "**" is not special.  `xr' is a strict inverse of `rx': if
S is a regexp in string form, (rx-to-string (xr S)) is always an
equivalent regexp in string form - the `xr' code tries hard to guarantee
that.

That means that `xr' is a great tool to check for the real semantics of
doubtful string regexps like your case.

thierryvolpiatto:
Michael Heerdegen ***@***.***> writes:

>
> Michael Heerdegen ***@***.***> writes:
> That means that `xr' is a great tool to check for the real semantics of
> doubtful string regexps like your case.

Indeed, thanks, I didn't know this package.

> —

-- 
Thierry

PR #2655, opened 2024-03-26
Don't erase helm buffer on some one-char input patterns.

karbiv:
When typing some short patterns like "!", "?", "@!", then Helm erases buffer, makes it empty.
It may be a distraction for some tastes. A list of buffers disappears that could be a visual clue sometimes.

The only change is in **`helm-update-source-p`**.
`helm-update-blacklist-regexps` is removed as not used.
One line is removed from `helm-update`, that erases Helm buffer with overly simplistic condition:

`(unless sources (erase-buffer))`

### DIFF
--- a/helm-core.el
+++ b/helm-core.el
@@ -99,9 +99,6 @@ in `helm-initialize'.")
   "The value of point at selection.")
 (defvar helm-alive-p nil)
 (defvar helm-visible-mark-overlays nil)
-(defvar helm-update-blacklist-regexps '("^" "^ *" "$" "!" " " "\\b"
-                                        "\\<" "\\>" "\\_<" "\\_>" ".*"
-                                        "??" "?*" "*?" "?"))
 (defvar helm--force-updating-p nil
   "[INTERNAL] Don't use this in your programs.")
 (defvar helm-exit-status 0
@@ -5070,12 +5067,6 @@ without recomputing them, it should be a list of lists."
                 (cl-loop for src in helm-sources
                          when (helm-update-source-p src)
                          collect src))
-          ;; When no sources to update erase buffer
-          ;; to avoid duplication of header and candidates
-          ;; when next chunk of update will arrive,
-          ;; otherwise the buffer is erased AFTER [1] the results
-          ;; are computed.
-          (unless sources (erase-buffer))
           ;; Compute matches without rendering the sources.
           ;; This prevent the helm-buffer flickering when constantly
           ;; updating.
@@ -5084,7 +5075,7 @@ without recomputing them, it should be a list of lists."
           ;; If computing matches finished and is not interrupted
           ;; erase the helm-buffer and render results (Fix #1157).
           (when matches ;; nil only when interrupted by while-no-input.
-            (erase-buffer)             ; [1]
+            (erase-buffer)
             (cl-loop for src in sources
                      for mtc in matches
                      do (helm-render-source src mtc))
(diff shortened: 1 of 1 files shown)


helm-core.el
```diff
@@ -5121,16 +5112,17 @@ without recomputing them, it should be a list of lists."
...
-         (not (member (replace-regexp-in-string "\\s\\ " " " helm-pattern)
-                      helm-update-blacklist-regexps)))))
+         ;; bugs were mentioned here before: Bug#1802, bug#2423
+         ;; must be fixed already by 'helm-re-search-forward'
+         ;;
+         ;; short patterns that are alone not expected to return any results,
+         ;; ignore in order to not erase Helm buffer on such patterns.
+         (not
+          (let ((split-string-default-separators "[ \f\t\n\r\v]+")
+                (subpatterns (string-split helm-pattern)))
```
thierryvolpiatto, line 5122: Your split-string is not affected by split-string-default-separators because of let and not let*.
BTW why do you want such a separator? isn't splitting with helm-mm-split-pattern enough?

helm-core.el
```diff
@@ -5121,16 +5112,17 @@ without recomputing them, it should be a list of lists."
...
-         ;; so defer update. Maybe replace spaces quoted when using
-         ;; multi-match.
-         (not (member (replace-regexp-in-string "\\s\\ " " " helm-pattern)
-                      helm-update-blacklist-regexps)))))
+         ;; bugs were mentioned here before: Bug#1802, bug#2423
+         ;; must be fixed already by 'helm-re-search-forward'
+         ;;
+         ;; short patterns that are alone not expected to return any results,
+         ;; ignore in order to not erase Helm buffer on such patterns.
+         (not
```
thierryvolpiatto, line 5120: You don't need not, see below

helm-core.el
```diff
@@ -5121,16 +5112,17 @@ without recomputing them, it should be a list of lists."
...
+          (let ((split-string-default-separators "[ \f\t\n\r\v]+")
+                (subpatterns (string-split helm-pattern)))
+            (seq-find
+             (lambda (p) (member p '("!" "!^" "?" "@!")))
+             subpatterns nil))))))
```
thierryvolpiatto, line 5125: I avoid using seq in Helm for many reasons:
- compatibility with other Emacs versions (different seq versions depending Emacs versions)
- Performances
- Readability
- seq is not needed unless the argument passed to it may be of different kind (list, vector or whatever)

I prefer using cl-loop:

```elisp
(cl-loop for p in (helm-mm-split-pattern helm-pattern)
                  never (member p '("!" "!^" "?" "@!")))
``` 
Using never avoid the not on top.

thierryvolpiatto:

> When typing some short patterns like "!", "?", "@!", then Helm erases buffer, makes it empty.
> It may be a distraction for some tastes. A list of buffers disappears that could be a visual clue sometimes.

Thanks, I added some comments in the diff, have a look.

> The only change is in helm-update-source-p.
> helm-update-blacklist-regexps is removed as not used.
> One line is removed from helm-update, that erases Helm buffer with overly simplistic condition:
>
> (unless sources (erase-buffer))

Ok, lets remove this line, in theory it should not be needed anymore, I
remember of these duplicate headers that this line is intended to fix
but can't remember in which conditions this happened, maybe async
sources?
Perhaps you could leave the line but commented and with a little comment
about what it was intended to do previously?

Thanks.


-- 
Thierry

helm-core.el
karbiv: > Your split-string is not affected by split-string-default-separators because of let and not let*.

Yeah, added it and then got distracted by some other aspect. That binding probably not needed at all. Chances that something will change separator definitions are almost nonexistent.

helm-core.el
karbiv: > I prefer using cl-loop:

That's better. Checking.

karbiv:
> Perhaps you could leave the line but commented and with a little comment about what it was intended to do previously?

It could be something in a function in `helm-after-update-hook`.

The comment of that line states: "_;; when next chunk of update will arrive_".
Chunks don't make sense in this context because after `erase-buffer` a buffer must be recreated completely anew by
`helm-render-source` which is invoked only in `helm-update`.

thierryvolpiatto:
Thanks, merged!

thierryvolpiatto:

>     Perhaps you could leave the line but commented and with a little comment about what it was intended to do previously?
>    
> It could be something in a function in helm-after-update-hook.
>
> The comment of that line states: ";; when next chunk of update will arrive".
> Chunks don't make sense in this context because after erase-buffer a buffer must be recreated completely anew by
> helm-render-source which is invoked only in helm-update.

Maybe I was refering to the process-filter in async sources when I wrote
this comment, but still, I hardly see how this could happen at this
point.  Never mind, if it happens in the future, we will know where it
comes from.

> —

-- 
Thierry

PR #2668, opened 2024-05-10
fix(helm-files): outdated cache and errors when finding files in Docker

taquangtrung:
Hi,

I notice that when finding files in a Docker container, new changes in its file system aren't notified back to Helm in the host machine, or some errors occur when the guest system tries to notify the changes.

I fixed this issue by ignoring the errors, so `helm-ff-directory-files` can still return the result.

I also added a condition to only use cached result on non-remote file system.

Can you help to review and merge this PR?

Thanks!

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -1,6 +1,6 @@
 ;;; helm-files.el --- helm file browser and related. -*- lexical-binding: t -*-
 
-;; Copyright (C) 2012 ~ 2023 Thierry Volpiatto 
+;; Copyright (C) 2012 ~ 2023 Thierry Volpiatto
 
 ;; This program is free software; you can redistribute it and/or modify
 ;; it under the terms of the GNU General Public License as published by
@@ -239,7 +239,7 @@ Should not be used among other sources.")
     (define-key map (kbd "C-c C-a")       'helm-ff-run-mail-attach-files)
     (define-key map (kbd "C-c p")         'helm-ff-run-print-file)
     (define-key map (kbd "C-c /")         'helm-ff-run-find-sh-command)
-    (define-key map (kbd "C-/")           'helm-ff-run-fd) 
+    (define-key map (kbd "C-/")           'helm-ff-run-fd)
     ;; Next 2 have no effect if candidate is not an image file.
     (define-key map (kbd "M-l")           'helm-ff-rotate-left-persistent)
     (define-key map (kbd "M-r")           'helm-ff-rotate-right-persistent)
@@ -1858,7 +1858,7 @@ this working."
                 (setq cmd-line (format command mapfiles)) ; See [1]
               (setq cmd-line (format "%s %s" command mapfiles)))
             (eshell-command cmd-line))
-        
+
         ;; Run eshell-command sequencially on EACH marked files.
         ;; To work with tramp handler we have to call
         ;; COMMAND on basename of each file, using
@@ -2137,7 +2137,7 @@ prefix arg shell buffer doesn't exists, create it and switch to it."
                when (file-directory-p r)
                ;; We can use this as long as this filtering function
                ;; is called after `helm-ff-fct' otherwise candidates
-               ;; may not be cons cell at first call [1]. 
+               ;; may not be cons cell at first call [1].
                collect (cons d r))
     candidates))
 
@@ -3607,7 +3607,7 @@ debugging purpose."
           ;; Check here if directory is accessible (not working on Windows).
           ((and (file-directory-p path) (not (file-readable-p path)))
            ;; Prefix error message with @@@@ for safety
-           ;; (some files may match file-error See bug#2400) 
+           ;; (some files may match file-error See bug#2400)
            (list (cons (format "@@@@file-error: Opening directory permission denied `%s'" path)
                        path)))
           ;; A fast expansion of PATH is made only if `helm-ff-auto-update-flag'
@@ -3785,7 +3785,7 @@ later in the transformer."
 ;; bug#2542.
 ;; Store here associations of (truename . symlink) when opening a
 ;; symlinked directory, then add the watch to the truename; When the
-;; watcher ring on the truename remove the symlinked directory from cache.  
+;; watcher ring on the truename remove the symlinked directory from cache.
 (defvar helm-ff--list-directory-links nil)
 
 (defun helm-ff-directory-files (directory &optional force-update)
@@ -3805,6 +3805,7 @@ in cache."
       (cl-pushnew (cons truename directory)
                   helm-ff--list-directory-links :test 'equal))
     (or (and (not force-update)
+             (not (file-remote-p directory))
              (gethash directory helm-ff--list-directory-cache))
         (let* (file-error
                (ls   (condition-case err
@@ -3818,7 +3819,7 @@ in cache."
                        (file-error
                         (prog1
                             ;; Prefix error message with @@@@ for safety
-                            ;; (some files may match file-error See bug#2400) 
+                            ;; (some files may match file-error See bug#2400)
                             (list (format "@@@@%s:%s"
                                           (car err)
                                           (mapconcat 'identity (cdr err) " ")))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Ta Quang Trung ***@***.***> writes:

> Hi,
>
> I notice that when finding files in a Docker container, new changes in its file system aren't notified back to Helm in the host machine,
> or some errors occur when the guest system tries to notify the changes.
>
> I fixed this issue by ignoring the errors, so helm-ff-directory-files can still return the result.
>
> I also added a condition to only use cached result on non-remote file system.
>
> Can you help to review and merge this PR?

Isn't (setq helm-ff-use-notify nil) enough to fix your issue?

> Thanks!
>
> -------------------------------------------------------------------------------------------------------------------------
>
>
>   https://github.com/emacs-helm/helm/pull/2668
>
> Commit Summary
>
> * 208e177 fix(helm-files): outdated cache, errors occur when finding files in Docker
>
> File Changes 
>
> (1 file)
>
> * M helm-files.el (84) 
>
> Patch Links:
>
> * https://github.com/emacs-helm/helm/pull/2668.patch
> * https://github.com/emacs-helm/helm/pull/2668.diff
>
> —

-- 
Thierry

taquangtrung:
Previously, I also tried to use that setting `(setq helm-ff-use-notify nil)`, but it also disables notifying file changes in the local file system. For example, in an Emacs session, if I create a new folder from terminal, this folder won't appear in `helm-find-files` until I restart Emacs.

The current patch fixes the issue that even if the file notification in the code from lines 3837-3863 fails, `prog1` will still return the result by lines 3832-3836, by wrapping lines 3837-3863 inside an `ignore-errors`:

https://github.com/emacs-helm/helm/blob/85aae8a8bee8940e761cbabc92d6af7dcf1a92e2/helm-files.el#L3831-L3863

thierryvolpiatto:
Ta Quang Trung ***@***.***> writes:

> Previously, I also tried to use that setting (setq helm-ff-use-notify
> nil), but it also disables notifying file changes in the local file
> system. For example, in an Emacs session, if I create a new folder
> from terminal, this folder won't appear in helm-find-files until I
> restart Emacs.

No, you just have to C-c C-u (force update).

> The current patch fixes the issue that even if the file notification
> in the code from lines 3837-3863 fails, prog1 will still return the
> result by lines 3832-3836, by wrapping lines 3837-3863 inside an
> ignore-errors:

I understand what the patch does, but using here ignore-errors would
defeat error handling in notify, so I don't want to use this.
Thus you are adding a condition previously for file-remote that I don't
understand.

What about allowing helm-ff-use-notify to use something else than a
boolean value i.e. (setq helm-ff-use-notify 0) in this case it would
disable checking the cache as well in addition to disabling inotify:

    diff --git a/helm-files.el b/helm-files.el
    index 096b6a9e..9e2cdb21 100644
    --- a/helm-files.el
    +++ b/helm-files.el
    @@ -3805,6 +3805,7 @@ in cache."
           (cl-pushnew (cons truename directory)
                       helm-ff--list-directory-links :test 'equal))
         (or (and (not force-update)
    +             (booleanp helm-ff-use-notify)
                  (gethash directory helm-ff--list-directory-cache))
             (let* (file-error
                    (ls   (condition-case err

Even better would be to have a predicate to know if we are in an
environment incompatible with notify like a container.

> https://github.com/emacs-helm/helm/blob/85aae8a8bee8940e761cbabc92d6af7dcf1a92e2/helm-files.el#L3831-L3863

PS: Could you remove whitespace changes in your patches? This is really
annoying when reviewing.

Thanks.

> —

-- 
Thierry

thierryvolpiatto:
I didn't understand the `(not (file-remote-p directory))` part because you didn't tell me how you visit files in your container (I never used container so my knowledge with this is limited).  I guess you are using the tramp docker method, right?
If so the following patch may work for you:

```elisp
---
 helm-files.el | 25 +++++++++++++++++++++----
 1 file changed, 21 insertions(+), 4 deletions(-)

diff --git a/helm-files.el b/helm-files.el
index 096b6a9e..412261aa 100644
--- a/helm-files.el
+++ b/helm-files.el
@@ -850,13 +850,23 @@ directories belonging to each visible windows."
                  nil)))
 
 (defcustom helm-ff-use-notify t
-  "Watch directories visited with `helm-find-files' when non nil.
+  "Control watching and caching directories visited with `helm-find-files'.
+
+When `t' watch and cache directories.
+When `nil' stop watching directories but continue caching directories.
+Any other value disable watching directories and disable caching as well.
+
 If your system have no file notification package available turn this
-to nil to avoid error messages when using `helm-find-files'."
-  :type 'boolean
+to nil to avoid error messages when using `helm-find-files'.
+
+Warning: Do not use `setq' to set this variable."
+  :type '(radio
+          (const :tag "Enable watching and caching directories" t)
+          (const :tag "Disable watching only" nil)
+          (const :tag "Disable watching and caching" 0))
   :set (lambda (var val)
 	 (set-default var val)
-	 (unless (symbol-value var)
+	 (unless (eq (symbol-value var) t)
            (cl-loop for dir being the hash-keys of helm-ff--file-notify-watchers
                     do (remhash dir helm-ff--list-directory-cache)))))
 
@@ -3788,6 +3798,10 @@ later in the transformer."
 ;; watcher ring on the truename remove the symlinked directory from cache.  
 (defvar helm-ff--list-directory-links nil)
 
+(defvar helm-ff-incompatible-notify-methods
+  '("docker" "toolbox" "podman" "kubernetes")
+  "Tramp methods icompatible with (i)notify.")
+
 (defun helm-ff-directory-files (directory &optional force-update)
   "List contents of DIRECTORY.
 Argument FULL mean absolute path.
@@ -3805,6 +3819,9 @@ in cache."
       (cl-pushnew (cons truename directory)
                   helm-ff--list-directory-links :test 'equal))
     (or (and (not force-update)
+             (booleanp helm-ff-use-notify)
+             (not (member (file-remote-p directory 'method)
+                          helm-ff-incompatible-notify-methods))
              (gethash directory helm-ff--list-directory-cache))
         (let* (file-error
                (ls   (condition-case err
-- 
2.34.1
```
Let me know if it fit your needs.

taquangtrung:
No, it doesn't work. I tried to set `helm-ff-use-notify` to all the 3 values 0, t, nil, but none of them works .

I'm using helm tramp over both ssh, and docker. There are 2 scenarios of finding files by Helm as bellow:

1. In an Emacs session, create a new file/folder from a terminal (both in a Docker or in a host session). This scenario needs file watcher to be enabled.

2. Browse a new folder that hasn't been browsed before by helm tramp. Lines 3832-3836 code in `prog1` correctly finds the nested files/folders and update to cache. However, an error occur in the next part configuring file watchers. Hence, no result is returned in the first time browsing that folder. 
 
   The second time browsing is fine since it can get the result from the cache. That's why I used `ignore-errors` previously to ensure the first time browsing also returns result.

https://github.com/emacs-helm/helm/blob/85aae8a8bee8940e761cbabc92d6af7dcf1a92e2/helm-files.el#L3832-L3836

thierryvolpiatto:
Ta Quang Trung ***@***.***> writes:

> No, it doesn't work. I tried to set helm-ff-use-notify to all the 3 values 0, t, nil, but none of them works .
>
> I'm using helm tramp over both ssh, and docker. There are 2 scenarios of finding files by Helm as bellow:
>
> 1 In an Emacs session, create a new file/folder from a terminal (both in a Docker or in a host session). This scenario needs file
>  watcher to be enabled.
>
> 2 Browse a new folder that hasn't been browsed before by helm tramp.

I still don't know how you "browse a new folder".
What mean browsed before by helm tramp? What is helm tramp?

> Lines 3832-3836 code in prog1 correctly finds the nested files/folders
> and update to cache. However, an error occur in the next part
> configuring file watchers.

Please forget this part, just send a precise recipe describing what you
are doing exactly.

> Hence, no result is returned in the first time browsing that folder.

Where is "that folder"?

-- 
Thierry

thierryvolpiatto:
Ta Quang Trung ***@***.***> writes:

>
> No, it doesn't work. I tried to set helm-ff-use-notify to all the 3 values 0, t, nil, but none of them works .

Also are you sure you are using the last patch?
I made an error in first version and edited the post to update the
patch, be sure to copy/paste the patch from github (not what you
received by email).

-- 
Thierry

thierryvolpiatto:
I have completely forgotten the existence of `helm-ff-inotify-unsupported-methods`, so this should be enough to fix your issue with your "docker" tramp method, you will have to update your directories under your container with `C-c C-u` though.

thierryvolpiatto:
Ta Quang Trung ***@***.***> writes:

>  I have completely forgotten the existence of helm-ff-inotify-unsupported-methods, so this should be enough to fix your issue
>  with your "docker" tramp method, you will have to update your directories under your container with C-c C-u though.
>
> Nice! I added (add-to-list 'helm-ff-inotify-unsupported-methods "docker") into my configuration, and now Helm can return
> correct sub dirs and files inside a Docker container!

Great!

> And indeed, the cache isn't updated when I create a new file/folder
> inside the Docker. Would it be possible to add a new option like
> helm-ff-cache-unsupported-methods below to control when not using the
> cache, so that I don't have to press C-u every time?

Done.

> I'm running an experiment inside a Docker and it will generate new
> output files every time.
>
> (and (not force-update)
>              (not (member method helm-ff-cache-unsupported-methods))
>              (gethash directory helm-ff--list-directory-cache))

Thanks.

> —

-- 
Thierry

taquangtrung:
Thanks a lot for your effort! Perfect patch!

Issue #2669, opened 2024-05-11
Extra space is inserted to helm completion in shell mode


kamikado:
### What happened?

I have a trouble that is basically same with the old closed issue<https://github.com/emacs-helm/helm/issues/2274>.

### How to reproduce?

$ cd ~/.emacs.d/.cask/29.1/elpa/helm-20240510.1833
$ ./emacs-helm.sh 
M-x shell
$ cd ~/D<TAB>
Then three candidates appear in the helm-mini buffer
<Desktop/>, <Downloads/>,<Documents/>
If I choose <Desktop/> then an unnecessary space is inserted at the end
$ cd ~/Desktop/ 

I can reproduce this behavior on both MaxOS and Ubuntu.

### Helm Version

Master branch

### Emacs Version

Emacs-29.1

### OS

MacOSX

### Minimal configuration

- [X] I agree using a minimal configuration

kamikado:
The root cause of this problem is that exit-fun is called with 'finished at <https://github.com/emacs-helm/helm/blob/85aae8a8bee8940e761cbabc92d6af7dcf1a92e2/helm-mode.el#L2716>.

try-completion returns non desired value (probably we expect "~/Desktop/" but it returns "~/D" ) and "string-match "/\\'" it" is false.
I confirmed this by adding printf in the code.


<img width="1407" alt="Screen Shot 2024-05-11 at 14 28 39" src="https://github.com/emacs-helm/helm/assets/22497404/6d1c4f70-7ea4-44f4-9a6b-7d36e2dc5f7e">

<img width="695" alt="Screen Shot 2024-05-11 at 14 28 10" src="https://github.com/emacs-helm/helm/assets/22497404/eb792213-809a-4460-93ff-3cb7e7c0d964">

thierryvolpiatto:
Kazuhiko Kamikado ***@***.***> writes:

> What happened?
>
> I have a trouble that is basically same with the old closed issue#2274.
>
> How to reproduce?
>
> $ cd ~/.emacs.d/.cask/29.1/elpa/helm-20240510.1833
> $ ./emacs-helm.sh
> M-x shell
> $ cd ~/D
> Then three candidates appear in the helm-mini buffer
> , ,
> If I choose then an unnecessary space is inserted at the end
> $ cd ~/Desktop/

Let me know if last change (c2f9bee9) fixes it, thanks.

> I can reproduce this behavior on both MaxOS and Ubuntu.
>
> Helm Version
>
> Master branch
[...]

PR #2670, opened 2024-05-11
fix(#2669): exit-fun is not called with 'exact in helm-mode.el

kamikado:
Hi, 

As I reported in <https://github.com/emacs-helm/helm/issues/2669>, there seems undesired behavior in helm on shell mode. 

>Function is called with 'exact if the return value of
>        ;; `try-completion' is a string ending with / (possibly a directory
>        ;; Bug#2274),

I thought, for the purpose there is no need to call try-completion and to check selected string is enough.
I confirmed my modification works well in my environment.
Can you help to review and merge this PR?

### DIFF
--- a/helm-mode.el
+++ b/helm-mode.el
@@ -2714,11 +2714,10 @@ Can be used for `completion-in-region-function' by advicing it with an
         ;; really bad (see bug#2646).
         (when (and (stringp string) exit-fun)
           (funcall exit-fun string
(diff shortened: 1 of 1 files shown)


kamikado:
As I confirmed that https://github.com/emacs-helm/helm/commit/c2f9bee9f437401e9229a8debafe05f6f3ae9689 solved the issue, I close this PR.

PR #2680, opened 2024-07-23
fix base may be nil on Windows in helm-common-dir

zbelial:
On windows, files may be in different partitions, like c:/f1, d:/f2. In this case, they don't have common directory, and helm-common-dir will fail with error "(wrong-type-argument stringp nil)" as `base` is nil.

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -6006,8 +6006,9 @@ and `dired-compress-files-alist'."
   "Return the longest common directory path of FILES list"
   (cl-loop with base = (car files)
            for file in files
+           while base
            do (setq base (fill-common-string-prefix base file))
-           finally return (file-name-directory base)))
+           finally return (when base (file-name-directory base))))
 
 (defun helm-ff--dired-compress-file (file)
   ;; `dired-compress-file' doesn't take care of binding `default-directory' when


thierryvolpiatto:
太阳 ***@***.***> writes:

> On windows, files may be in different partitions, like c:/f1,
> d:/f2. In this case, they don't have common directory, and
> helm-common-dir will fail with error "(wrong-type-argument stringp
> nil)" as base is nil.

Thanks to catch this. Unfortunately your patch will fail in many cases.
This issue raise as well another bug in helm-common-dir which doesn't
fit with its docstring description:
    "Return the longest common directory path of FILES list."
It returns the file-name-directory of the real longest common directory!
This is now fixed.
For your issue I have changed helm-common-dir so that it returns a list
of path in the case of files not belonging to same drive:

    (helm-common-dir '("c:/foo" "c:/foo/bar/baz"
                       "f:/foo" "e:/foo" "f:/foo/bar"
                       "d:/foo" "d:/foo/bar/baz"
                       "e:/foo/bar/baz"))
=>    ("c:/foo" "e:/foo" "f:/foo" "d:/foo")

    (helm-common-dir '("/home/you/download/foo/"
                       "/home/you/download/foo/bar/baz"
                       "/home/you/tmp/foo"))
=>    "/home/you"

This way you can call grep (with several C-s IIUC) on each dir until you
find what you want.

I will commit the changes in a few minutes.

Thanks.

>
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
>
>   https://github.com/emacs-helm/helm/pull/2680
>
> Commit Summary
>
>   • 59fe2fa fix base may be nil on Windows in helm-common-dir
>
> File Changes
>
> (1 file)
>
>   • M helm-files.el (3)
>
> Patch Links:
>
>   • https://github.com/emacs-helm/helm/pull/2680.patch
>   • https://github.com/emacs-helm/helm/pull/2680.diff
>
> —
> You are receiving this because you are subscribed to this thread.*Message ID: ***@***.***>

-- 
Thierry

zbelial:
Thanks. I didn't bother to tackle multiple drives to return a list of common directories, one for each drive.  One reason is no occuring multiple buffers in my workflow, the other and the more important reason is it's too complicated to implement the grep-in-parent-dir feature for me, I don't know how to grep multiple directories( for example, `helm-grep-ag-1` only greps in one directory) using facilities available in helm.

thierryvolpiatto:
太阳 ***@***.***> writes:

> Thanks. I didn't bother to tackle multiple drives to return a list of
> common directories, one for each drive. One reason is no occuring
> multiple buffers in my workflow, the other and the more important
> reason is it's too complicated to implement the grep-in-parent-dir
> feature for me, I don't know how to grep multiple directories( for
> example, helm-grep-ag-1 only greps in one directory) using facilities
> available in helm.

Let start back from the beginning with a possible workflow:

1) Launch helm-mini, select 3 buffers:
file1, file2, file3, each belong to a different directory, respectively
dir1, dir2, dir3.

2) Hit C-s which launch helm-occur on these 3 buffers and search the
term "foo" in these 3 buffers. If there is no results at all you are
done and you have better time hitting "C-x b", unmark and hit "C-x C-f" on
one of the buffers and then "M-g a" to search buffer's directory.

3) On the other hand if you had result in 2) with helm-occur, you should end up
with 1/3 sources with results for each buffer, from each source you
could provide an action launching helm-ag in the directory belonging to
the buffer of this source.

What do you think?

-- 
Thierry

zbelial:
> What do you think?

I need some time to organize my thought and sentences, I think I will reply to you by local time tomorrow. 

But I do have questions about your new `helm-common-dir`. In the case of multiple drives, what it returns actually is not `common directory` any more, so it may not work in `helm-find-files-do-action` as expected, where it's supposed to be a single directory.

zbelial:
> It is about making archive with several files or directories, doing this makes sense only in the same drive, if the user add several files coming from different drive an error is returned: (helm-aand (eq action 'compress) (helm-common-dir ifiles) (if (stringp it) it (error "Try to compress files not belonging to same drive")))

I didn't notice your change here, maybe because I was sleepy then. Sorry. 

About the workflow you posted, it's a little complicated for me. From the very beginning of implementing the grep-in-parent-dir feature, what I wanted was a simple, straightforward operation, so I can just press "C-s", "C-s", "C-s" .... until I found what I needed or nothing. Frankly, the workflow you proposed is not what I wanted. 

With your new `helm-common-dir`, following are how I implement the feature, and what my workflow looks like. 
```elisp
;;; my changes to helm-occur.el
;;; 1. I added a new command `helm-occur-grep-in-parent-dir-ag`, 
(defun helm-occur-grep-in-parent-dir-ag ()
  "Grep in the parent directory. If you were occuring multiple buffers
before invoking this command, it will grep from the common directory."
  (interactive)
  (let* ((all-files (delq nil (mapcar (lambda (buf) (buffer-file-name buf)) helm-occur--buffers)))
         (parent-dir (helm-common-dir all-files))
         (input helm-input))
    (if (and parent-dir (stringp parent-dir))
        (helm-run-after-exit #'helm-grep-ag-1 parent-dir nil input)
      (message "No common parent directory to grep."))))
;;; 2. helm-occur-grep-in-parent-dir-ag uses a new added variable `helm-occur--buffers', which stores which buffers occur is running in. 
(defvar helm-occur--buffers nil
  "Used to help grep in parent directory in `helm-occur-grep-in-parent-dir-ag'.")

;;; my changes to helm-grep.el
;;; 1. I added a new command `helm-grep-in-parent-dir-ag'
(defun helm-grep-in-parent-dir-ag ()
  "Grep in the parent directory."
  (interactive)
  (let ((parent-dir (file-name-parent-directory helm-grep-ag-dir))
        (type helm-grep-ag-type)
        (input helm-input))
    (when parent-dir
      (helm-run-after-exit #'helm-grep-ag-1 parent-dir type input))))
;;; 2. `helm-grep-in-parent-dir-ag' uses two new added variables `helm-grep-ag-dir', `helm-grep-ag-type'.
(defvar helm-grep-ag-dir nil)
(defvar helm-grep-ag-type nil)

;;; in `helm-grep-ag-1', it will stores `directory' and `type' to  `helm-grep-ag-dir' and `helm-grep-ag-type' respectively. 
```

Above are all what I have changed.  

In my personal config, I added some key bindings. 
```elisp
(global-set-key (kbd "C-s")  #'helm-occur)
(global-set-key (kbd "C-c s r")  #'helm-do-grep-ag)
(define-key helm-occur-map (kbd "C-s") #'helm-occur-grep-in-parent-dir-ag)
(define-key helm-grep-map (kbd "C-s") #'helm-grep-in-parent-dir-ag)
```

Here is how I use this feature. 
1. In a buffer `buf1` of file `f1`, I press "C-s" (aka. helm-occur) to occur something, If What I find is not what I need, I'll press "C-s" to grep in `f1`'s directory, and then "C-s" to grep in the parent directory, and the parent-of-parent directory. 
2. Sometime I'll use `helm-do-grep-ag` directly to grep something in current buffer's directory, and similiarly if I could find what I need, I'll press "C-s" to grep in upper directories.

So for me the key point is I can use one "C-s" to do the occur and grep, and I don't have to switch buffers and use other commands. 

Hope I've clearly described my workflow and why I don't want what you proposed.  And if this feature is not suitable to be merged into helm, I'm totally understand, since it's very specific to my personal workflow. 

By the way, your new `helm-common-dir` does not work as expected when FILES is a single file list, for example `(helm-common-dir '("/home/xxx/.emacs.d/init.el"))` will return "/home/xxx/.emacs.d/init.el" (I think it should return "/home/xxx/.emacs.d", right? ).  
You may change the last line of `helm-common-dir-1` to `(list (and base (if (file-directory-p base) (directory-file-name base) (file-name-directory base))))`.

thierryvolpiatto:
太阳 ***@***.***> writes:

>     It is about making archive with several files or directories, doing this makes sense only in the same drive, if the user add several files coming from different drive an error is
>     returned: (helm-aand (eq action 'compress) (helm-common-dir ifiles) (if (stringp it) it (error "Try to compress files not belonging to same drive")))
>    
> I didn't notice your change here, maybe because I was sleepy then. Sorry.
>
> About the workflow you posted, it's a little complicated for me. From
> the very beginning of implementing the grep-in-parent-dir feature,
> what I wanted was a simple, straightforward operation, so I can just
> press "C-s", "C-s", "C-s" .... until I found what I needed or
> nothing. Frankly, the workflow you proposed is not what I wanted.

Probably we are miscommucating as the workflow you describe here is more
or less what I have shown. Also with the code you sent helm-common-dir
is not needed.

After updating Helm, can you try this code:

    (defun helm-occur-grep-ag-buffer-directory (candidate)
      (let* ((src (with-helm-buffer (car helm-sources)))
             (buf (helm-get-attr 'buffer-name src)))
             (directory (with-current-buffer buf
                         default-directory))
             (input helm-pattern))
        (helm-grep-ag-1 directory nil input)))

    (add-to-list 'helm-occur-actions '("Ag grep directory buffer" . helm-occur-grep-ag-buffer-directory) t)

    (helm-make-command-from-action helm-run-occur-grep-ag-buffer-directory
        "Ag grep directory buffer." 'helm-occur-grep-ag-buffer-directory)

    (define-key helm-occur-map (kbd "C-s") 'helm-run-occur-grep-ag-buffer-directory)

    (defun helm-grep-ag-grep-parent-directory (candidate)
      (let* ((src (with-helm-buffer (car helm-sources)))
             (directory (helm-basedir (helm-get-attr 'directory src) t))
             (input helm-pattern))
        (helm-grep-ag-1 directory nil input)))

    (add-to-list 'helm-grep-actions '("Ag grep parent directory" . helm-grep-ag-grep-parent-directory) t)

    (helm-make-command-from-action helm-grep-run-ag-grep-parent-directory
        "Ag grep parent directory." 'helm-grep-ag-grep-parent-directory)

    (define-key helm-grep-map (kbd "C-s") 'helm-grep-run-ag-grep-parent-directory)


-- 
Thierry

zbelial:
> Probably we are miscommucating as the workflow you describe here is more or less what I have shown. Also with the code you sent helm-common-dir is not needed. After updating Helm, can you try this code:

Hi @thierryvolpiatto . Sure, I'll try your code tomorrow.  And huge thanks for your help! 
I just found I know little about helm, and I feel sorry that this function "has wasted" you so much time.  Thanks.

zbelial:
> 太阳 ***@***.***> writes:
> It is about making archive with several files or directories, doing this makes sense only in the same drive, if the user add several files coming from different drive an error is returned: (helm-aand (eq action 'compress) (helm-common-dir ifiles) (if (stringp it) it (error "Try to compress files not belonging to same drive"))) I didn't notice your change here, maybe because I was sleepy then. Sorry. About the workflow you posted, it's a little complicated for me. From the very beginning of implementing the grep-in-parent-dir feature, what I wanted was a simple, straightforward operation, so I can just press "C-s", "C-s", "C-s" .... until I found what I needed or nothing. Frankly, the workflow you proposed is not what I wanted.
> Probably we are miscommucating as the workflow you describe here is more or less what I have shown. Also with the code you sent helm-common-dir is not needed. After updating Helm, can you try this code: (defun helm-occur-grep-ag-buffer-directory (candidate) (let* ((src (with-helm-buffer (car helm-sources))) (buf (helm-get-attr 'buffer-name src))) (directory (with-current-buffer buf default-directory)) (input helm-pattern)) (helm-grep-ag-1 directory nil input))) (add-to-list 'helm-occur-actions '("Ag grep directory buffer" . helm-occur-grep-ag-buffer-directory) t) (helm-make-command-from-action helm-run-occur-grep-ag-buffer-directory "Ag grep directory buffer." 'helm-occur-grep-ag-buffer-directory) (define-key helm-occur-map (kbd "C-s") 'helm-run-occur-grep-ag-buffer-directory) (defun helm-grep-ag-grep-parent-directory (candidate) (let* ((src (with-helm-buffer (car helm-sources))) (directory (helm-basedir (helm-get-attr 'directory src) t)) (input helm-pattern)) (helm-grep-ag-1 directory nil input))) (add-to-list 'helm-grep-actions '("Ag grep parent directory" . helm-grep-ag-grep-parent-directory) t) (helm-make-command-from-action helm-grep-run-ag-grep-parent-directory "Ag grep parent directory." 'helm-grep-ag-grep-parent-directory) (define-key helm-grep-map (kbd "C-s") 'helm-grep-run-ag-grep-parent-directory)
> […](#)
> -- Thierry

This works flawlessly. During a short time of testing, I didn't find any problem, except I had to adjust parenthesis in `helm-occur-grep-ag-buffer-directory`, where there is a redundant parenthesis at the end of line `(buf (helm-get-attr 'buffer-name src))`.  
I think I can replace my code with yours.  Thanks.

thierryvolpiatto:
太阳 ***@***.***> writes:

> This works flawlessly. During a short time of testing, I didn't find
> any problem, except I had to adjust parenthesis in
> helm-occur-grep-ag-buffer-directory, where there is a redundant
> parenthesis at the end of line (buf (helm-get-attr 'buffer-name src)).
> I think I can replace my code with yours. Thanks.

Ok, thanks for your feedback.
I have made several other changes since yesterday to integrate properly
in Helm, I will merge those changes right now.

Thanks.

> —

-- 
Thierry

zbelial:
Today I used your latest commit and helm worked very well so far. Thanks. 
Closing this PR.

PR #2690, opened 2024-12-03
helm-frame-alpha: Set default value to actual value used if nil

Thaodan:
The default value of `helm-frame-alpha' didn't match the specified
custom type. Since the default value which is in use if nil is 100
setting it to 100 is the best choice as it doesn't break any existing
code where setting to nil is used to set it to the default value of
100.


thierryvolpiatto:
Thanks merged!

PR #2698, opened 2025-01-22
Add affixations for Info files

pkryger:
I've taken a liberty to create this PR to provide similar functionality to what
`helm` already provides for `helm-locate-library` and `find-library`
commands. I've taken a very similar approach here by scanning files to get a
nice Info file description, based on the Info dir entry.

I've also used this opportunity to update affixations' face in
`helm-info-at-point`.

I've tried to split the work into logical parts into separate commits. Please
let me know if you have any suggestions.

### DIFF
--- a/helm-info.el
+++ b/helm-info.el
@@ -144,6 +144,7 @@ If line have a node use the node, otherwise use directly first name found."
    (filtered-candidate-transformer
     :initform
     (lambda (candidates _source)
+      (require 'helm-mode)
       (cl-loop for line in candidates
                when (string-match helm-info--node-regexp line)
                do (progn
@@ -153,7 +154,7 @@ If line have a node use the node, otherwise use directly first name found."
                      nil line)
                     (helm-add-face-text-properties
                      (match-beginning 2) (match-end 2)
-                     'font-lock-warning-face
+                     'helm-completions-detailed
                      nil line))
                collect line)))
    (display-to-real :initform #'helm-info-display-to-real)
--- a/helm-info.el
+++ b/helm-info.el
@@ -34,6 +34,7 @@
 (declare-function ring-ref "ring")
 (defvar Info-history)
 (defvar Info-directory-list)
+(defvar helm-completions-detailed)
 ;; `Info-minibuf-history' is not declared in Emacs, see emacs bug/58786.
 (when (and (> emacs-major-version 28)
            (not (boundp 'Info-minibuf-history)))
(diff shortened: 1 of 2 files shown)


pkryger:
> You did it by advice, but it is not the way to go, you don't need advices, use
> instead `helm-completing-read-extra-metadata` and
> `helm-completing-read-command-categories`, add a category to the former and use
> this category in the other

This is clearly a nice idea. Thank you for pointing it out - I think the code
became neater when using it

> Also I have here error when there is an info file not installed on system,
> didn't have the time yet to debug it.

Please let me know if I can be of any help.

FWIW, I've tried the following scenario, to test the not-installed Info file:
1. Open a not installed Info file, for me it was Embark:
``` emacs-lisp
(info "/Users/pkryger/gh/oantolin/embark/embark.info")
```
2. Run `M-x info-display-manual`, select `embark`
3. The Info buffer from point 1 is selected.

helm-info.el
```diff
@@ -240,15 +241,71 @@ helm-info-<CANDIDATE>."
...
+(defvar helm-info--files-doc-cache (make-hash-table :test 'equal))
+
+(defun helm-info-file-doc (file)
+  "Return dir entry from the Info FILE."
+  (or
+   (with-temp-buffer
+     (ignore-errors (info-insert-file-contents file))
+     (goto-char (point-min))
+     (when (re-search-forward (rx line-start "START-INFO-DIR-ENTRY" line-end)
+                              nil t)
```
thierryvolpiatto, line 254: Please use plain regexps, not RX, I don't understand it, it's easier for me to read plain regexps.

helm-info.el
```diff
@@ -240,15 +241,71 @@ helm-info-<CANDIDATE>."
...
+       (forward-line 1)
+       (when (re-search-forward
+              (rx line-start
+                  "*" (one-or-more whitespace)
+                  (group (one-or-more (not ":")))
+                  ":" (one-or-more whitespace)
+                  "(" (one-or-more (not ")"))
+                  ")." (one-or-more whitespace)
+                  (group (one-or-more any)))
```
thierryvolpiatto, line 263: Same as above.

helm-info.el
```diff
@@ -240,15 +241,71 @@ helm-info-<CANDIDATE>."
...
+                                   (make-progress-reporter
+                                    "Scanning libraries..." 0 (length candidates)))
+                 for c in candidates
+                 for count from 0
+                 for sep = (helm-make-separator c longest)
+                 for file = (or
+                             (assoc-default c helm-info--files-cache)
+                             (let ((file (Info-find-file c)))
+                               (push (cons c file) helm-info--files-cache)
+                               file))
```
thierryvolpiatto, line 291: When `Info-find-file` fails to find a file it popup the top dir info buffer as default, which override the helm buffer, I fixed it like this:

```elisp
(let ((file (Info-find-file c t)))
   (when file (push (cons c file) helm-info--files-cache))
   file)
```
Use the noerror arg of Info-find-file.

thierryvolpiatto:
Przemysław Kryger ***@***.***> writes:

>     Also I have here error when there is an info file not installed on system,
>     didn't have the time yet to debug it.
>    
> Please let me know if I can be of any help.

I added some comments in your patches to fix this and also the info
buffer displaying top dir which override the helm window.
Seems it is working fine like this, I will have a closer look tomorrow
for further cleaning.

Thanks.


-- 
Thierry

pkryger:
> I added some comments in your patches to fix this and also the info buffer
> displaying top dir which override the helm window.  Seems it is working fine
> like this, I will have a closer look tomorrow for further cleaning.

Have implemented suggestions. Let me know if you find anything else to fix.

thierryvolpiatto:
Thanks, merged!

thierryvolpiatto:
However something is wrong in the affixation function for helm-mode, when an Info buffer is found its description is used for all entries, this:

```elisp
                  ;; The `info-display-manual' favours `Info-mode'
                  ;; buffers. However, each buffer like that may be opened from
                  ;; a file that has not been installed and is not visible to a
                  ;; `Info-find-file'. Retrieve the such a file name from the
                  ;; buffer, and use it to scan for affixation, however don't
                  ;; cache the result in `helm-info--files-cache' to avoid using
                  ;; wrong file. For example, the latter can happen after a
                  ;; development version of info file has been opened.
                  (let ((blist (buffer-list))
                        (manual-re (concat "/\\|\\`" comp "\\.\\|\\'"))
                        (case-fold-search t)
                        found)
                    (dolist (buffer blist)
                      (with-current-buffer buffer
                        (when (and (derived-mode-p 'Info-mode)
		                   (stringp Info-current-file)
		                   (string-match manual-re
                                                 Info-current-file))
	                  (setq found Info-current-file
                                blist nil))))
                    found)
```
Not sure to understand what the intention is here, but it is wrong anyway.
Can we remove this piece of code?

thierryvolpiatto:
Ah! it is the regexp (manual-re) which is wrong, lets try to fix it.

thierryvolpiatto:
Ok done by fixing the regexp.

thierryvolpiatto:
Przemysław Kryger ***@***.***> writes:

>     Ah! it is the regexp (manual-re) which is wrong, lets try to fix it.
>     Ok done by fixing the regexp.
>    
> Sorry about that. Has pushed to hastily after converting from re
> syntax. Thank you for debugging and correcting!
>
>     Not sure to understand what the intention is here, but it is wrong anyway.
>     Can we remove this piece of code?
>    
> I think I can see at least two scenarios supporting this buffer scan.
>
>  1. User opens a not installed Info file, with (info "/some/not/installed/file.info"). Then, upon calling info-display-manual,
>     the value of Info-current-file in the buffer displaying the Info file becomes
>     one of the completing-read candidates. Alas, the file.info is not
>     installed, so neither it will be cached nor it will be found by
>     Info-find-file.
>    
>  2. User opens a not installed Info file, with (info "/some/not/installed/file.info"). However, the /some/not/installed/file.info
>     shadows an already installed one, i.e., the file.info (with possible suffix)
>     is found somewhere Info-directory-list or in
>     Info-additional-directory-list. In such a case the buffer version takes
>     precedence over installed one, when calling info-display-manual. I guess the
>     latter is helpful when one develops the Info content and want to inspect it.
>    
> In cases like that the affixation should be taken from the not installed Info
> file (i.e., to display anything, or to display an updated Info dir entry).

Ok, thanks for explanations, that's fine now we have fixed the regexp.

> —

-- 
Thierry

PR #2699, opened 2025-01-23
Add prefix to denote buffers in info-display-manual affixation

pkryger:
Following up on #2698, a small idea to give a visual indicator in
`info-display-manual` when a candidate is a buffer.

I have only thought about it when writing the most recent explanation. Let me
know if this is something worth having.

### DIFF
--- a/helm-mode.el
+++ b/helm-mode.el
@@ -1425,6 +1425,7 @@ is used."
   (require 'helm-info)
   (lambda (comp)
     (let* ((sep (helm-make-separator comp))
+           (prefix "")
            (file (or
                   ;; The `info-display-manual' favours `Info-mode'
                   ;; buffers. However, each buffer like that may be opened from
@@ -1440,10 +1441,11 @@ is used."
                     (cl-loop for buffer in (buffer-list) thereis
                              (with-current-buffer buffer
                                (when (and (derived-mode-p 'Info-mode)
-		                          (stringp Info-current-file)
-		                          (string-match
+		                                  (stringp Info-current-file)
+		                                  (string-match
                                            manual-re Info-current-file))
-	                         Info-current-file))))
+                                 (setq prefix "*")
+	                             Info-current-file))))
                   (assoc-default comp helm-info--files-cache)
                   (helm-aif (Info-find-file comp t)
                       (prog1 it (push (cons comp it) helm-info--files-cache)))))
@@ -1451,7 +1453,8 @@ is used."
                         (puthash file (helm-info-file-doc file)
                                  helm-info--files-doc-cache))))
       (list comp
-            ""
+            (helm-aand (propertize prefix 'face 'helm-completions-detailed)
+                       (propertize " " 'display (format "%-2s" it)))
             (helm-aand (propertize summary 'face 'helm-completions-detailed)
                        (propertize " " 'display (concat sep it)))))))
 


thierryvolpiatto:
Przemysław Kryger ***@***.***> writes:

>
> Following up on #2698, a small idea to give a visual indicator in
> info-display-manual when a candidate is a buffer.
>
> I have only thought about it when writing the most recent explanation. Let me
> know if this is something worth having.

Yes, why not, however some commentary:

- It will match only the first buffer found (like the previous behavior
though) i.e. if there is for some reasons "*info*" and "*info<2>*"
etc... Doing some more extended search (on all buffers) may slow down
this code, so probably it fine as it is.
 
- We will have always a leading space in display, need:
@@ -1454 +1456,3 @@
-            ""
+            (helm-aand (not (string= prefix ""))
+                       (propertize prefix 'face 'helm-completions-detailed)
+                       (propertize " " 'display (format "%-2s" it)))

- Also there is whitespaces in your patch (bad indent only).

> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
>
>   https://github.com/emacs-helm/helm/pull/2699
>
> Commit Summary
>
>   • 6a0b087 Add prefix to denote buffers in info-display-manual affixation
>
> File Changes
>
> (1 file)
>
>   • M helm-mode.el (11)
>
> Patch Links:
>
>   • https://github.com/emacs-helm/helm/pull/2699.patch
>   • https://github.com/emacs-helm/helm/pull/2699.diff
>
> —
> You are receiving this because you are subscribed to this thread.*Message ID: ***@***.***>

-- 
Thierry

pkryger:
> - It will match only the first buffer found (like the previous behavior though) i.e. if there is for some reasons "*info*" and "*info<2>*" etc... Doing some more extended search (on all buffers) may slow down this code, so probably it fine as it is

I think it's all right as it is, I opened a few info files and when I did `helm-display-manual`, typed `el\|emacs\|eieio` I got result like in the following screenshot:
Screenshot 2025-01-23 at 11 32 15

>. - We will have always a leading space in display, need:
> @@ -1454 +1456,3 @@
> - "" + (helm-aand (not (string= prefix ""))
> + (propertize prefix 'face 'helm-completions-detailed)
> + (propertize " " 'display (format "%-2s" it)))

As soon as I did that, and repeated the command and the query as above, I got:
Screenshot 2025-01-23 at 11 39 08

My original intention was to present user with a prefix, similar to what `helm`'s affixation for `completing-read` for `describe-function` does. Please let me know if you think the latter is better and I will update. 

>  - Also there is whitespaces in your patch (bad indent only).

I think i did it on purpose (marked the whole function and then hit `C-i`). This way the condition of `when` form is clearly indented at different level that the body of the form. But can take it out, or split into a separate commit if you prefer

thierryvolpiatto:
Przemysław Kryger ***@***.***> writes:

>
>       □ It will match only the first buffer found (like the previous behavior though) i.e. if there is for some reasons "info" and "info<2>" etc... Doing some more extended search (on all
>         buffers) may slow down this code, so probably it fine as it is
>
> I think it's all right as it is, I opened a few info files and when I
> did helm-display-manual, typed el\|emacs\|eieio I got result like in
> the following screenshot: Screenshot.2025-01-23.at.11.32.15.png (view
> on web)

Yes, you are right, I tried and got the same behavior.

>
>
>     . - We will have always a leading space in display, need:
>     @@ -1454 +1456,3 @@
>    
>       □ "" + (helm-aand (not (string= prefix ""))
>       □ (propertize prefix 'face 'helm-completions-detailed)
>       □ (propertize " " 'display (format "%-2s" it)))
>
> As soon as I did that, and repeated the command and the query as above, I got:
> Screenshot.2025-01-23.at.11.39.08.png (view on web)

Ah yes, we lost alignment, so it's fine as it is.

>       □ Also there is whitespaces in your patch (bad indent only).
>
> I think i did it on purpose (marked the whole function and then hit
> C-i). This way the condition of when form is clearly indented at
> different level that the body of the form.

No it is not, for some reasons we loose indentation after the when.

> But can take it out, or split into a separate commit if you prefer

No it's ok I will merge manually and fix it on the way.

Thanks.

> —

-- 
Thierry

thierryvolpiatto:
Merged, thanks! (have fixed indentation).

PR #2703, opened 2025-01-28
Add autoload to helm-highlight-files

pkryger:
Function `helm-highlight-files' is used in type `helm-type-file'. Ensure that
the definition is available for clients, should they need it.

See https://github.com/bbatsov/helm-projectile/issues/184.


thierryvolpiatto:
Thanks merged.

PR #2706, opened 2025-02-04
Add affixations to completions in package-vc commands

pkryger:
I have scrubbed the `package-vc.el` for all interactive commands that either
call `package-vc--read-package-desc` or `package-vc--read-package-name`.


thierryvolpiatto:
Thanks, merged!

PR #2716, opened 2025-05-10
fix: exception occurring when trimming empty summary string

taquangtrung:
Hi,

This PR fixes the exception when Helm tries to get the summary of libraries whose summary strings are empty.

Example: the library [haskell-c2hs.el](https://github.com/haskell/haskell-mode/blob/master/haskell-c2hs.el) has empty summary string: the content between `---` and `-*-` is empty.

```emacs-lisp
;; haskell-c2hs.el --- -*- lexical-binding: t; -*-
```

This causes Helm to throw an exception when running `find-library`. The main culprit function that throws the exception is `split-string`, when it attempts to trim the string `"[ \11\n\15-]+"` from an empty summary .

```emacs-lisp
Debugger entered--Lisp error: (args-out-of-range "-*- lexical-binding: t; -*-" 1 0)
  split-string("-*- lexical-binding: t; -*-" "-\\*-" nil "[ \11\n\15-]+")
  helm-locate-lib-get-summary("/home/trung/.config/emacs/.local/straight/build-31.0.50/haskell-mode/haskell-c2hs.el")
  #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_48>("haskell-c2hs")
  helm-completion--decorate(("haskell" "hashcash"  ...))
  ....
 ```

There might be other libraries that contain similar issues.

I fixed this bug by checking if the summary string is non-empty before trimming it.

Please consider merging this PR.

Thanks!

### DIFF
--- a/helm-lib.el
+++ b/helm-lib.el
@@ -2024,11 +2024,9 @@ Directories expansion is not supported."
       (goto-char (point-min))
       (when (re-search-forward "^;;;?\\(.*\\) ---? \\(.*\\)" (pos-eol) t)
         (setq desc (match-string-no-properties 2)))
-      (if-let ((_ (and desc (length> desc 0)))
-               (summary (car (split-string desc "-\\*-" nil)))
-               (_ (length> summary 0)))
-          (string-trim summary "[ \t\n\r-]+" "[ \t\n\r-]+")
-        "Not documented"))))
+      (if (or (null desc) (string= "" desc) (string-match "\\`-*-" desc))
+          "Not documented"
+        (car (split-string desc "-\\*-" nil "[ \t\n\r-]+"))))))
 
 (defun helm-local-directory-files (directory &rest args)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by thierryvolpiatto** [comment]
Thanks, isn't this simpler:

```diff
@@ -2027 +2027 @@
-      (if (or (null desc) (string= "" desc))
+      (if (or (null desc) (string= "" desc) (string-match "\\`-*-" desc))
```

taquangtrung:
@thierryvolpiatto: you're right! :)

I just updated the patch with your fix!

thierryvolpiatto:
I forget to quote the "*", sorry, it should be instead:

```elisp
(or (null desc) (string= "" desc) (string-match "\\`-\\*-" desc))
```

taquangtrung:
@thierryvolpiatto : thanks! I just updated the fix again.

thierryvolpiatto:
Thanks, merged!

PR #2724, opened 2025-06-27
Generate lexical cooky in the adaptive-history file

amno1:
Can we possibly get a lexical-binding cookie generated when adaptive-history file is generated so we skip the annoying warnings in Emacs? If you want this patch or want it add it yourself, I would be happy with any.

### DIFF
--- a/helm-adaptive.el
+++ b/helm-adaptive.el
@@ -173,7 +173,7 @@ Returns nil if `helm-adaptive-history-file' doesn't exist."
   (when helm-adaptive-history-file
     (with-temp-buffer
       (insert
-       ";; -*- mode: emacs-lisp -*-\n"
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
Arthur Miller ***@***.***> writes:

> Can we possibly get a lexical-binding cookie generated when
> adaptive-history file is generated so we skip the annoying warnings in
> Emacs?

Ah yes, good catch. I never had this problem because I set
helm-adaptive-history-file to nil and let psession.el store the data itself.

> If you want this patch or want it add it yourself, I would be happy
> with any.

Will merge your PR, thanks a lot.

-- 
Thierry

thierryvolpiatto:
Merged! Thanks.

PR #2731, opened 2025-07-14
helm-grep: run ag from the current project root

taquangtrung:
Hi,

I added a new function to run helm-grep-ag from the current project root.
Could you help to review it?

Thanks!

### DIFF
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -1857,9 +1857,15 @@ With prefix arg prompt for type if available with your AG version."
   (require 'helm-files)
   (require 'project)
   (helm-aif (project-current)
-      (let* ((project-root (expand-file-name (car (last it))))
-             (default-directory project-root))
-        (helm-grep-ag project-root arg))
+      (let* ((project-type (car it))
+             (project-root (cond ((equal project-type 'vc)
+                                  (car (last it)))
+                                 ((equal project-type 'projectile)
+                                  (cdr it))
+                                 (t default-directory)))
+             (project-root-abs (expand-file-name project-root))
+             (default-directory project-root-abs))
+        (helm-grep-ag project-root-abs arg))
     (message "Not in any project!")))
 
 ;;;###autoload


thierryvolpiatto:
Quang Trung Ta ***@***.***> writes:

> Hi,
>
> I added a new function to run helm-grep-ag from the current project root.
> Could you help to review it?

Thanks, but I am not sure we need such a function, IME when you are
working on a buffer you want to grep in its defaultdirectory 99% of the times,
if you don't find what you want you may want to grep from the parent
directory and so on, this is provided by C-s while you are in helm.
Would this workflow fit your needs?

-- 
Thierry

taquangtrung:
@thierryvolpiatto: Thanks for your comment.

One scenario that I often use is that when I am editing a source file deep in a folder-structured project (e.g., in Rust), and I know exactly that I want to search for a string in the entire project. For example, search for a string in a test file, or search for the function/variable names in other crates.

Calling this new function will save time, instead of having to C-s to go step by step to the parent folder.

thierryvolpiatto:
Quang Trung Ta ***@***.***> writes:

> ● taquangtrung left a comment (emacs-helm/helm#2731)
>
> @thierryvolpiatto: Thanks for your comment.
>
> One scenario that I often use is that when I am editing a source file
> deep in a folder-structured project (e.g., in Rust), and I know
> exactly that I want to search for a string in the entire project. For
> example, search for a string in a test file, or search for the
> function/variable names in other crates.
>
> Calling this new function will save time, instead of having to C-s to go step by step to the parent folder.

You can C-x C-f to root and hit M-g a in such cases.

> —

-- 
Thierry

taquangtrung:
Oh, `M-g` in my config was rebound by other package, so I wasn't aware of such usage.

Still, calling `helm-grep` from `helm-find-files` is still a two-step action.

Will it make sense to expose these functions of Helm as interactive commands so that users can bind them directly?

thierryvolpiatto:
Quang Trung Ta ***@***.***> writes:

> ● taquangtrung left a comment (emacs-helm/helm#2731)
>
> Oh, M-g in my config was rebound by other package, so I wasn't aware of such usage.
>
> Still, calling helm-grep from helm-find-files is still a two-step action.
>
> Will it make sense to expose these functions of Helm as interactive commands so that users can bind them directly?

No these functions are meant to be used from HFF.

> —

-- 
Thierry

thierryvolpiatto:
Merged thanks.

PR #2734, opened 2025-07-21
helm-grep: Support default input for grep actions

sunlin7:
Currently the `helm-do-grep-ag`, `helm-do-grep-ag-project` and `helm-grep-do-git-grep` will start the action with the symbol under cursor, but some wrapper function maybe want start from the word under cursor or start from empty. 
So provide the extra parameter "default" as the initial input.

### DIFF
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -1766,7 +1766,7 @@ grep ag."
 
 (defvar helm-source-grep-ag nil)
 
-(defun helm-grep-ag-1 (directory &optional type input)
+(defun helm-grep-ag-1 (directory &optional type input default-input)
   "Start helm ag in DIRECTORY maybe searching in files of type TYPE.
 If INPUT is provided, use it as the search string."
   (setq helm-source-grep-ag
@@ -1787,6 +1787,7 @@ If INPUT is provided, use it as the search string."
   (helm-set-local-variable 'helm-input-idle-delay helm-grep-input-idle-delay)
   (helm :sources 'helm-source-grep-ag
         :history 'helm-grep-ag-history
+        :default default-input
         :input input
         :truncate-lines helm-grep-truncate-lines
         :buffer (format "*helm %s*" (helm-grep--ag-command))))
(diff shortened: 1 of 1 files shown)


thierryvolpiatto:
"lin.sun" ***@***.***> writes:

> Currently the helm-do-grep-ag, helm-do-grep-ag-project and
> helm-grep-do-git-grep will start the action with the symbol under
> cursor, but some wrapper function maybe want start from the word under
> cursor or start from empty.

To write such wrappers, please use lower level functions like
helm-grep-ag-1 or helm-grep-git-1, I don't like adding optional
parameters to commands accessible only non interactively.

So all we need is an optional DEFAULT parameter to helm-grep-ag-1:

diff --git a/helm-grep.el b/helm-grep.el
index 60c62fb0..caceb2aa 100644
--- a/helm-grep.el
+++ b/helm-grep.el
@@ -1766,7 +1766,7 @@ grep ag."
 
 (defvar helm-source-grep-ag nil)
 
-(defun helm-grep-ag-1 (directory &optional type input)
+(defun helm-grep-ag-1 (directory &optional type input default)
   "Start helm ag in DIRECTORY maybe searching in files of type TYPE.
 If INPUT is provided, use it as the search string."
   (setq helm-source-grep-ag
@@ -1788,6 +1788,7 @@ If INPUT is provided, use it as the search string."
   (helm :sources 'helm-source-grep-ag
         :history 'helm-grep-ag-history
         :input input
+        :default default
         :truncate-lines helm-grep-truncate-lines
         :buffer (format "*helm %s*" (helm-grep--ag-command))))

Then it is easy to call helm-grep-ag-1 or helm-grep-git-1 with 
(thing-at-point 'word).

Note that you have also C-w that yank what you have at point word by
word.

-- 
Thierry

sunlin7:
Okay, got your point. Please submit your charges, and I will turn to the low level APIs, thank you!

thierryvolpiatto:
"lin.sun" ***@***.***> writes:

> ● sunlin7 left a comment (emacs-helm/helm#2734)
> Okay, got your point. Please submit your charges, and I will turn to the low level APIs, thank you!

Done.  Thanks.

> —

-- 
Thierry

sunlin7:
Thank you !

PR #2745, opened 2026-02-08
Quote EMACS_COMMAND and ASYNC_STRAIGHT_DIR variables

alexg0:
Quoting variables to address directories with whitespace


thierryvolpiatto:
Thanks merged!

PR #2748, opened 2026-04-10
Fix edit marked files binding in helm-generic-files-map

renatofdds:


thierryvolpiatto:
Thanks merged, fixed documentation as well.

thierryvolpiatto:
Also fixed the action itself in `helm-type-file-actions`.

renatofdds:
Thank you!

PR #2759, opened 2026-07-22
Fix CRM insertion with regexp separators

wade-arista:
When completing CRM candidates, Helm may need to infer the literal
separator to append from `crm-separator`.

Commit 52e1a7f9 added this for cases such as `[ \t]*:[ \t]*`, where
stripping optional whitespace from the regexp gives the literal `:`
separator and avoids falling back to `,`.

That heuristic also applied to arbitrary regexp separators.  Callers
such as Magit use `\\.\\.\\.\\?` to read revision ranges, where the
regexp describes accepted input but is not itself a literal separator
to insert.  In that case Helm inserted the escaped regexp text after a
candidate.

This keeps the 52e1a7f9 behavior when stripping optional whitespace
leaves a single literal character, but otherwise uses only an explicit
`separator` text property or `helm-crm-default-separator`.

Verification:
- `emacs -q -batch -L . -L ../build/async -l ./helm-mode.el -l ./test/helm-mode-tests.el -f ert-run-tests-batch-and-exit`
- `make STRAIGHT_DIR=../build all`

### DIFF
--- a/helm-mode.el
+++ b/helm-mode.el
@@ -3135,7 +3135,9 @@ Be sure to know what you are doing when modifying this.")
                          crm-separator)
                         (and (stringp crm-separator)
                              (or (get-text-property 0 'separator crm-separator)
-                                 (string-replace "[ \t]*" "" crm-separator)))
+                                 (let ((sep (string-replace
+                                             "[ \t]*" "" crm-separator)))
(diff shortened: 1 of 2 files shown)


thierryvolpiatto:
Thanks, merged!

PR #2760, opened 2026-07-23
Add fallback for non-GNU remote ls

claritystorm:
Removes `helm-list-remote-directory-fn` user option. Remote directory listing now always tries external `ls` first, falling back to the lisp implementation when `ls` is unavailable or incompatible (e.g., FreeBSD, BusyBox). Users no longer need to configure the listing function.

### DIFF
--- a/helm-files.el
+++ b/helm-files.el
@@ -590,35 +590,6 @@ command on remote (and/or locally if you want to trash as root).
 On Ubuntu-based distributions it is \\='trash-cli'."
   :type 'boolean)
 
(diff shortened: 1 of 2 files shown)


thierryvolpiatto:
Thanks, merged! Sorry for the delay.

thierryvolpiatto:
I have reverted for now this patch as it breaks all HFF sort commands, I see as well wrongly quoted fns etc...
I don't have the time currently to fully review this code and fix it, please have a look.
Thanks.

thierryvolpiatto:
Now fixed and merged again.
While I was at it I fixed as well the sort by extension feature on remote directories which was broken.

PR #2761, opened 2026-09-01
Fix helm-completion-theme-affixation for theme with no docstring

zeph1e:
helm--get-theme-doc-1 may return nil for theme with no docstring while helm-completion-theme-affixation propertizes it.


thierryvolpiatto:
Thanks, merged!

